PR #27, opened 2022-11-16
fix PulseAudio init failed crash

Alto1772:
Based on: https://github.com/HarbourMasters/Shipwright/pull/1837

> If Pulseaudio daemon is not running on Linux, the game crashes upon after starting the game.


dcvz:
We can leave this one open but this would be immediately made obsolete by having audio fallback which is also opened here #28. Audio fallback is IMO the better solution.

Kenix3:
I would say we still want to check if we can play audio when the play function is invoked.

Merge pending conflict resolution.

Kenix3 (approved):

PR #28, opened 2022-11-16
implement audio player fallback

Alto1772:
This part came from https://github.com/HarbourMasters/Shipwright/pull/1857 with some major changes:
- both **mGfxBackend** and **mAudioBackend** are not removed
- moved [this part](https://github.com/HarbourMasters/Shipwright/blob/f23159222010b942da46a22543971759c2c823cf/soh/soh/OTRGlobals.cpp#L1715-L1732) into audioplayerbridge.cpp


Kenix3:
Looks like you need to run clang-format.
`find src include -name "*.cpp" -o -name "*.h" | sed 's| |\\ |g' | xargs clang-format-12 -i`

src/core/Window.cpp
```diff
@@ -685,4 +691,26 @@ int32_t Window::GetLastScancode() {
 void Window::SetLastScancode(int32_t scanCode) {
     mLastScancode = scanCode;
 }
+
+bool Window::StartAudioPlayer() {
```
Kenix3: You'll want to call this function in the Init function in audioplayerbridge.cpp

Kenix3 (changes requested):

Kenix3 (approved):

PR #38, opened 2022-11-21
[CMake] Prefix exports with libultraship

dcvz:
This make it so libs that use this now import via `#include <libultraship/libultraship.h>` instead of `# include <libultraship.h>`


Kenix3:
Added this PR so we could get a release candidate for 1.0.0 while you were sleeping:
https://github.com/Kenix3/libultraship/pull/47

PR #42, opened 2022-11-21
Fixes checking game versions on resource bridge.

leggettc18:
Changes function signature for existing call to return a list of game verions (since there can be more than one) and adds one for checking for a specific game version.


src/core/bridge/resourcebridge.cpp
```diff
@@ -126,8 +126,15 @@ size_t GetResourceTexSizeByCrc(uint64_t crc) {
     return -1;
 }
 
-uint32_t GetGameVersion(void) {
-    return Ship::Window::GetInstance()->GetResourceManager()->GetGameVersion();
```
Kenix3, line 130: Now remove this field from ResourceMgr :)

Kenix3 (changes requested):

src/core/bridge/resourcebridge.cpp
leggettc18: Should be done in 8345d8b

src/core/bridge/resourcebridge.cpp
```diff
@@ -126,8 +126,15 @@ size_t GetResourceTexSizeByCrc(uint64_t crc) {
...
+void GetGameVersions(uint32_t* versions, size_t versionsSize, size_t* versionsCount) {
```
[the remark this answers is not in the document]
leggettc18: Fixed in 38be4b8

PR #44, opened 2022-11-22
[Feature] Add API for speech synthesis

dcvz:


src/speechsynthesizer/DarwinSpeechSynthesizer.hpp
```diff
@@ -0,0 +1,29 @@
...
+public:
+    DarwinSpeechSynthesizer();
+
+    void Speak(const char* text);
+
+protected:
+    bool DoInit(void);
+
+private:
+    void *mSynthesizer;
```
Kenix3:
```suggestion
    void* mSynthesizer;
```

Kenix3 (changes requested):
Only other comment is that we use `h` instead of `hpp`. We definitely don't want the bridge to be `hpp`.

src/speechsynthesizer/DarwinSpeechSynthesizer.hpp
dcvz: This was changed by clang tidy as far as I remember.

Kenix3: It really should have failed the clang-format test actually.

dcvz: might be because of the hpp and h. it might only check .h files? renaming them now.

dcvz: yup, that was it! it now catches things in the h files

Kenix3 (commented):
I'm a little nervous that there is no Linux synthesizer, but I'm not sure if it needs to hold up this PR either.

PR #52, opened 2022-11-23
Binarytools-constructors

leggettc18:
These constructors instantiate a MemoryStream, as opposed to the calling code having to provide one separately. This allows consumers of the library to utilize the BinaryTools without us needing to export MemoryStream.


leggettc18:
Messed up with Git on this one, re-doing

PR #56, opened 2022-11-23
Adds `ToVector` method on BinaryReader/Writer and Stream

leggettc18:
For getting stream contents in consuming code without exposing MemoryStream.h directly.


leggettc18:
My branches are messed up, will fix for next time, a fixed version of this PR is already up.

PR #59, opened 2022-11-23
Exposes Archive class to C++ files that include libultraship.h

leggettc18:
includes Archive.h into libultraship.h but only when `#ifdef __cplusplus` is true. This way C++ code linking this library can create and add files to OTR files.

### DIFF
--- a/include/libultraship/classes.h
+++ b/include/libultraship/classes.h
@@ -1,3 +1,9 @@
+#pragma once
+#ifdef __cplusplus
+
+#ifndef LUSCLASSES_H
+#define LUSCLASSES_H
+
 #include "resource/Archive.h"
 #include "resource/ResourceMgr.h"
 #include "core/Window.h"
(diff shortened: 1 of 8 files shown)


include/libultraship/classes.h
```diff
@@ -0,0 +1,19 @@
+#include "resource/Archive.h"
```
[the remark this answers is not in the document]
leggettc18: Considering no GitHub actions failed, I have no idea why this "error" is showing up, it's happening on my local computer as well but clearly it's not actually preventing any builds from going through.

src/menu/Console.cpp
```diff
@@ -2,7 +2,7 @@
 
 #include "core/bridge/consolevariablebridge.h"
 #include "ImGuiImpl.h"
-#include <ImGui/imgui.h>
```
Kenix3, line 5: Why are we removing the ImGui path from the header include here?

leggettc18: I put the FetchContent statement in my project's CMakeLists.txt to my fork of LUS to test and it was erroring out because it couldn't find ImGui/imgui.h. Removing ImGui from that path fixed the error.

PR #63, opened 2022-12-07
Improves CMake settings to ease use in other projects

leggettc18:
This could probably be taken further, but for now this adds the necessary installs and exports to let libultraship play nicer while being used by other apps conforming to modern cmake best practices.

For example, now instead of needing to `FetchContent` libultraship directly in the toplevel `OTRIzer` library of SequenceOTRizer, I can just do a `find_package` of libultraship there and do the `FetchContent` in the `CMakeLists.txt` of SequenceOTRizer directly. Or, if libultraship was previously installed to the system or the CMake User Registry on my computer, I wouldn't even need to `FetchContent`. Speaking of, this part needs some testing, but this _should_ make libultraship installable/packageable as a deb or vcpkg or something.

### DIFF
--- a/.github/workflows/build-validation.yml
+++ b/.github/workflows/build-validation.yml
@@ -28,7 +28,7 @@ jobs:
       uses: actions/upload-artifact@v3
       with:
         name: soh-mac
-        path: build-cmake/src/*.a
+        path: build-cmake/lib/*.a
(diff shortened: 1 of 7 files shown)


Kenix3:
Closed after conversation with author. We feel its better to redo this later.

PR #69, opened 2022-12-20
 GX2: Improve texture updates 

GaryOderNichts:


Kenix3:
The failure here just means you need to run clang format. There's a command to run in the readme.

GaryOderNichts:
> The failure here just means you need to run clang format. There's a command to run in the readme.

Should be good now. Didn't realize the run had failed.

Kenix3 (approved):

PR #70, opened 2022-12-21
Add NetBSD and OpenBSD to the BSD ports

qurious-pixel:


Kenix3 (approved):
This should probably be PRed upstream to StormLib as well.

PR #72, opened 2022-12-24
Added support for switch controllers using native api

KiritoDv:
This PR Includes a whole new backend for the nintendo switch controllers made for lus that supports gyro, rumble and fixes some ui bugs.

Shipwright PR: [#2270](https://github.com/HarbourMasters/Shipwright/pull/2270)


leggettc18:
Left this comment on the SoH PR, also leaving it here in case it's more visible here.

This works great on my Switch apart from one small issue. If I attempt to use a pro controller while in handheld mode, the inputs go through just fine but rumble and Gyro stay on the joycons. They both work fine on the Pro Controller when the switch is docked, and they work great with the joycons in handheld mode, but I do sometimes find myself setting my switch up on my desk and playing with a pro controller with the Switch in handheld mode, would be nice if that would work. I think always using the pro controller for Rumble and Gyro if one is connected would be the ideal solution.

PR #80, opened 2023-01-07
[Fast3D] Metal Renderer

dcvz:
Adds Metal renderer for macOS - this should improve macOS experience as OpenGL is deprecated and does not receive good driver support. 

In order to use `C++` instead of `Objective-C` we use Apple's [metal-cpp](https://developer.apple.com/metal/cpp/) which is included in the PR - as a single header. These come from Apple and are licensed under Apache 2.0.


dcvz:
@Kenix3 this is ready now! There's 4 tidy warnings, but I followed the naming of what was around it.. so maybe this time we don't want to take it. I'd say your call.

dcvz:
This is still ready! As mentioned there are tidy warnings but I'm following the pattern already in that file. I don't think we should take the suggestions in this PR.

src/graphic/Fast3D/gfx_metal.cpp
```diff
@@ -0,0 +1,1041 @@
...
+
+bool Metal_IsSupported() {
+    return MTLCopyAllDevices()->count() > 0;
+}
+
+bool Metal_Init(SDL_Renderer* renderer) {
+    mctx.renderer = renderer;
+    NS::AutoreleasePool* autorelease_pool = NS::AutoreleasePool::alloc()->init();
+
+    mctx.layer = (CA::MetalLayer *)SDL_RenderGetMetalLayer(renderer);
```
briaguya0:
```suggestion
    mctx.layer = (CA::MetalLayer*)SDL_RenderGetMetalLayer(renderer);
```
this should make format happy

PR #94, opened 2023-01-17
fix: use `STATIC_LIBRARY_PREFIX` to address liblib issue

briaguya0:
this works

image

https://stackoverflow.com/questions/41804282/stop-cmake-from-prepending-lib-to-library-names

@Kenix3 if you would rather rebase the old commit out instead of having the revert feel free to cherry pick https://github.com/Kenix3/libultraship/commit/4436b0a6d37cfe6151974a3ef34c472731aac70d instead of merging this PR


Kenix3:
See 8c23df30edac0acd69afdf181d4fd95b826057f7

PR #96, opened 2023-01-17
Feature/switch-controllers on main

KiritoDv:


leggettc18:
Left this comment on the SoH PR, also leaving it here in case it's more visible here.

This works great on my Switch apart from one small issue. If I attempt to use a pro controller while in handheld mode, the inputs go through just fine but rumble and Gyro stay on the joycons. They both work fine on the Pro Controller when the switch is docked, and they work great with the joycons in handheld mode, but I do sometimes find myself setting my switch up on my desk and playing with a pro controller with the Switch in handheld mode, would be nice if that would work. I think always using the pro controller for Rumble and Gyro if one is connected would be the ideal solution.

extern/ImGui/imgui.h
```diff
@@ -1046,7 +1046,8 @@ enum ImGuiInputTextFlags_
...
     ImGuiInputTextFlags_CharsScientific     = 1 << 17,  // Allow 0123456789.+-*/eE (Scientific notation input)
     ImGuiInputTextFlags_CallbackResize      = 1 << 18,  // Callback on buffer capacity changes request (beyond 'buf_size' parameter value), allowing the string to grow. Notify when the string wants to be resized (for string types which hold a cache of their Size). You will be provided a new BufSize in the callback and NEED to honor it. (see misc/cpp/imgui_stdlib.h for an example of using this)
-    ImGuiInputTextFlags_CallbackEdit        = 1 << 19   // Callback on any edit (note that InputText() already returns true on edit, the callback is useful mainly to manipulate the underlying buffer while focus is active)
+    ImGuiInputTextFlags_CallbackEdit        = 1 << 19,  // Callback on any edit (note that InputText() already returns true on edit, the callback is useful mainly to manipulate the underlying buffer while focus is active)
+    ImGuiInputTextFlags_NoPlusMinusButtons  = 1 << 20   // Disable increment/decrement buttons
```
Kenix3, line 1050: What is this new flag doing?

briaguya0:
I tested this using https://github.com/HarbourMasters/Shipwright/pull/2444

First I merged in latest LUS, this exists on https://github.com/briaguya-ai/libultraship/tree/update-switch-on-main
Then I ran into some issues with https://github.com/Kenix3/libultraship/pull/97, so I made https://github.com/briaguya-ai/libultraship/tree/update-switch-on-main-two to address this

In my testing, I found:
* Rumble on joycons never worked (either connected or dual wireless)
* When a pro controller was connected, both the pro controller and the connected joycons could be used at the same time
  * When playing with the connected joycons, the pro controller would rumble
  
@KiritoDv these issues will need to be addressed before this can be considered ready to merge.

Kenix3:
While we want the changes, this will have to be redone under the new controller system. Feel free to open a new PR to do that.

PR #100, opened 2023-01-18
Makes necessary changes for compiling on Windows with llvm

leggettc18:


leggettc18:
Also compiles with clang on linux with these changes, which I don't think it did before.

Kenix3 (approved):

PR #105, opened 2023-01-25
Fix CrashHandler compile on arm64-linux

dapetcu21:
(Or any non x86 arch for that matter)

I'm trying to get Shipwright to build on Switchroot Ubuntu (arm64-linux) and this seemed to be the only compile error so far.


Kenix3 (approved):
We definitely want to verify we're on x86 before we read x86 registers. Good fix.

PR #109, opened 2023-02-08
[Controller] Virtual Notches

dcvz:
Adds the concept of virtual notches and adds a setting in ImGui to configure it.


Kenix3:
We'll want to get the Switch controller code updated with the new notch management
https://github.com/Kenix3/libultraship/pull/96

dcvz:
Drafting for now until #113 gets in

PR #110, opened 2023-02-08
[Audio] OSX Audio Engine

dcvz:
Adds an OSX audio player - SDLAudio which is the only supported audio backend is very choppy on macOS and suffers poor performance. This native player does better. It uses to OSS libraries:

- OpenEmuHelpers from the OpenEmu project: BSD
- TPCircularBuffer: MIT


extern/OpenEmuHelpers/OEGameAudio.h
```diff
@@ -0,0 +1,58 @@
+/*
+ Copyright (c) 2009, OpenEmu Team
+ 
+ Redistribution and use in source and binary forms, with or without
```
Kenix3, line 4: What license is this?

dcvz: > 3-clause BSD license

PR #112, opened 2023-02-09
Fix: Switch’s Usage of ImGui

dcvz:
Fixes up the Switch platform with some lost changes

### DIFF
--- a/src/port/switch/SwitchImpl.cpp
+++ b/src/port/switch/SwitchImpl.cpp
@@ -5,6 +5,7 @@
 #include "SwitchPerformanceProfiles.h"
 #include "core/bridge/consolevariablebridge.h"
 #include "misc/Hooks.h"
+#include <ImGui/imgui_internal.h>
 
 #define DOCKED_MODE 1
 #define HANDHELD_MODE 0
@@ -12,6 +13,7 @@
 static AppletHookCookie applet_hook_cookie;
 static bool isRunning = true;
 static bool hasFocus = true;
+static bool isShowingVirtualKeyboard = true;
 
 void DetectAppletMode();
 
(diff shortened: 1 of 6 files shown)


src/port/switch/SwitchImpl.cpp
```diff
@@ -74,6 +76,51 @@ void Ship::Switch::SetupFont(ImFontAtlas* fonts) {
     plExit();
 }
 
+void Ship::Switch::ImGuiSwapABXY(int startEvent) {
+    if (CVarGetInteger("gOpenMenuBar", 0)) {
+        return;
+    }
+
+    auto& input_queue = ImGui::GetCurrentContext()->InputEventsQueue;
+    for (int n = start_event; n < input_queue.Size; n++) {
```
briaguya0:
```suggestion
    for (int n = startEvent; n < input_queue.Size; n++) {
```

to fix a build error

src/port/switch/SwitchImpl.cpp
```diff
@@ -74,6 +76,51 @@ void Ship::Switch::SetupFont(ImFontAtlas* fonts) {
     plExit();
 }
 
+void Ship::Switch::ImGuiSwapABXY(int startEvent) {
+    if (CVarGetInteger("gOpenMenuBar", 0)) {
+        return;
+    }
```
briaguya0: this part makes no sense to me. in my testing, it makes it so we aren't doing the button switching at all when using the imgui menus. This means we have the switch "B" button acting as "confirm" and the switch "A" button acting as "cancel"

briaguya0: to add to this, I tried commenting this block out and I'm running into the issue where I can only press a button once when navigating the menu before input stops working and the controller can no longer be used to navigate the menu

dcvz: Yeah, this was a misunderstanding. I thought we don’t actually want this in menus only in game

briaguya0 (approved):
Tested here https://github.com/HarbourMasters/Shipwright/pull/2445/commits/1fd3838d8145b82fc31fd29b2b0258f288249c08

I can confirm this works as expected

Kenix3 (approved):

PR #113, opened 2023-02-13
[Controller] Backend Agnostic

### LINKED ISSUE #106: Input editor preview and gyro calibration don't work as expected
The issue comes from a fix on the ControlDeck, when the InputEditor is open or any other view, the game stops reading data from the controllers, because of that the stick preview and the gyro calibration don't work, as a few other bugs related to this.

This patch should fix it, its tested on both pc and switch

[Input Fix Patch](https://github.com/Kenix3/libultraship/files/10504505/fix.patch)

Demo:

https://user-images.githubusercontent.com/36680385/214713247-4acfe7ac-9c5e-4fdf-a597-a35b410e3a2a.mp4

dcvz:
Moves some core processing of inputs and range mapping out of individual backends and into the shared controller code, making it easier to add backends and reduces the surface in which bugs can be introduced.

Also fixes #106

### DIFF
--- a/src/controller/Controller.cpp
+++ b/src/controller/Controller.cpp
@@ -90,18 +90,6 @@ void Controller::ProcessStick(int8_t& x, int8_t& y, uint16_t deadzone) {
     }
     x *= len;
     y *= len;
-
-    // bound diagonals to an octagonal range {-68 ... +68}
-    if (x != 0.0f && y != 0.0f) {
-        auto slope = y / x;
-        auto edgex = copysign(MAX_AXIS_RANGE / (abs(slope) + 16.0f / 69.0f), x);
-        auto edgey = copysign(std::min(abs(edgex * slope), MAX_AXIS_RANGE / (1.0f / abs(slope) + 16.0f / 69.0f)), y);
-        edgex = edgey / slope;
-
-        auto scale = sqrt(edgex * edgex + edgey * edgey) / MAX_AXIS_RANGE;
-        x *= scale;
-        y *= scale;
-    }
 }
 
 void Controller::Read(OSContPad* pad, int32_t virtualSlot) {
@@ -125,6 +113,10 @@ void Controller::Read(OSContPad* pad, int32_t virtualSlot) {
     auto profile = getProfile(virtualSlot);
     ProcessStick(leftStickX, leftStickY, profile->AxisDeadzones[0]);
     ProcessStick(rightStickX, rightStickY, profile->AxisDeadzones[2]);
+    
+    if (pad == nullptr) {
+        return;
+    }
     
     padToBuffer.stick_x = leftStickX;
     padToBuffer.stick_y = leftStickY;
--- a/src/controller/Controller.cpp
+++ b/src/controller/Controller.cpp
@@ -7,6 +7,7 @@
 #else
 #include <SDL2/SDL_events.h>
 #endif
+#include <spdlog/spdlog.h>
 
 namespace Ship {
 
@@ -78,15 +79,20 @@ int8_t Controller::ReadStick(int32_t virtualSlot, Stick stick, Axis axis) {
     return 0;
 }
 
-void Controller::ProcessStick(int8_t& x, int8_t& y, uint16_t deadzone) {
+void Controller::ProcessStick(int8_t& x, int8_t& y, uint16_t deadzoneX, uint16_t deadzoneY) {
+    if (deadzoneX != deadzoneY) {
+        SPDLOG_TRACE("Invalid Deadzone configured. Up/Down was {} and Left/Right is {}", deadzoneY, deadzoneX);
+    }
+
+    // TODO: handle deadzones separately for X and Y
     // create scaled circular dead-zone in range {-15 ... +15}
     auto len = sqrt(x * x + y * y);
-    if (len < deadzone) {
+    if (len < deadzoneX) {
         len = 0;
     } else if (len > MAX_AXIS_RANGE) {
         len = MAX_AXIS_RANGE / len;
     } else {
-        len = (len - deadzone) * MAX_AXIS_RANGE / (MAX_AXIS_RANGE - deadzone) / len;
+        len = (len - deadzoneX) * MAX_AXIS_RANGE / (MAX_AXIS_RANGE - deadzoneX) / len;
     }
     x *= len;
     y *= len;
@@ -111,8 +117,8 @@ void Controller::Read(OSContPad* pad, int32_t virtualSlot) {
     int8_t rightStickY = ReadStick(virtualSlot, RIGHT, Y);
 
     auto profile = getProfile(virtualSlot);
-    ProcessStick(leftStickX, leftStickY, profile->AxisDeadzones[0]);
-    ProcessStick(rightStickX, rightStickY, profile->AxisDeadzones[2]);
+    ProcessStick(leftStickX, leftStickY, profile->AxisDeadzones[0], profile->AxisDeadzones[1]);
+    ProcessStick(rightStickX, rightStickY, profile->AxisDeadzones[2], profile->AxisDeadzones[3]);
 
     if (pad == nullptr) {
         return;
--- a/src/controller/Controller.cpp
+++ b/src/controller/Controller.cpp
@@ -19,6 +19,92 @@ Controller::Controller() : mIsRumbling(false) {
     }
 }
 
+int8_t Controller::ReadStick(int32_t virtualSlot, Stick stick, Axis axis) {
+    switch (stick) {
+        case Stick::LEFT: {
+            switch (axis) {
+                case Axis::X: {
+                    if (getLeftStickX(virtualSlot) == 0) {
+                        if (getPressedButtons(virtualSlot) & BTN_STICKLEFT) {
+                            return  -MAX_AXIS_RANGE;
+                        } else if (getPressedButtons(virtualSlot) & BTN_STICKRIGHT) {
+                            return MAX_AXIS_RANGE;
+                        }
+                    } else {
+                        return getLeftStickX(virtualSlot);
+                    }
+                }
+                case Axis::Y: {
+                    if (getLeftStickY(virtualSlot) == 0) {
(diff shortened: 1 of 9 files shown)


src/controller/Controller.h
```diff
@@ -44,6 +46,8 @@ class Controller {
     virtual int32_t ReadRawPress() = 0;
     virtual const std::string GetButtonName(int32_t virtualSlot, int32_t n64Button) = 0;
     virtual const std::string GetControllerName() = 0;
+    int8_t ReadStick(int32_t virtualSlot, Stick stick, Axis axis);
```
Kenix3: I think these should be protected methods rather than public. Nothing outside of the controller or it's derived class need to be aware of these functions.

src/controller/Controller.cpp
```diff
@@ -32,46 +105,23 @@ void Controller::Read(OSContPad* pad, int32_t virtualSlot) {
...
-        if (getPressedButtons(virtualSlot) & BTN_STICKDOWN) {
-            padToBuffer.stick_y = -128;
-        } else if (getPressedButtons(virtualSlot) & BTN_STICKUP) {
-            padToBuffer.stick_y = 127;
-        }
-    } else {
-        padToBuffer.stick_y = getLeftStickY(virtualSlot);
-    }
+    auto profile = getProfile(virtualSlot);
+    ProcessStick(leftStickX, leftStickY, profile->AxisDeadzones[0]);
```
Kenix3: I believe the 0 and 2 here are SDL dependent values. The axis used to represent the x direction for the left and right stick aren't guaranteed to be the same for each controller backend (wii u, switch, etc)

src/controller/Controller.cpp
```diff
@@ -19,6 +19,79 @@ Controller::Controller() : mIsRumbling(false) {
...
+    if (len < deadzone) {
+        len = 0;
+    } else if (len > MAX_AXIS_RANGE) {
+        len = MAX_AXIS_RANGE / len;
+    } else {
+        len = (len - deadzone) * MAX_AXIS_RANGE / (MAX_AXIS_RANGE - deadzone) / len;
+    }
+    x *= len;
+    y *= len;
+}
```
Kenix3, line 124: What happened to the octagonal gate processing?

src/controller/SDLController.cpp
```diff
@@ -241,12 +211,6 @@ void SDLController::ReadFromSource(int32_t virtualSlot) {
                     SPDLOG_TRACE("Invalid PosStickX configured. Neg was {} and Pos is {}", leftStickAxisX, axis);
                 }
 
-                if (leftStickDeadzone != 0 && leftStickDeadzone != axisDeadzone) {
```
Kenix3, line 244: Why remove all the warnings about the deadzones being different before adding code to process the x/y deadzones separately?

src/controller/SDLController.cpp
```diff
@@ -57,39 +57,12 @@ bool SDLController::Close() {
...
-        auto slope = ay / ax;
-        auto edgex = copysign(85.0f / (abs(slope) + 16.0f / 69.0f), ax);
-        auto edgey = copysign(std::min(abs(edgex * slope), 85.0f / (1.0f / abs(slope) + 16.0f / 69.0f)), ay);
-        edgex = edgey / slope;
-
-        auto scale = sqrt(edgex * edgex + edgey * edgey) / 85.0f;
-        ax *= scale;
-        ay *= scale;
-    }
+    auto ax = axisValueX * MAX_AXIS_RANGE / 32767.0f;
```
Kenix3: I recommend we keep this comment here
```c
// scale {-32768 ... +32767} to {-84 ... +84}
```

src/controller/Controller.cpp
Kenix3: If we're removing the code that checks that the axis deadzones are separate, then I believe we should also take in both the y deadzone for this function.

src/controller/Controller.h
```diff
@@ -11,10 +11,12 @@
 
 #define EXTENDED_SCANCODE_BIT (1 << 8)
 #define AXIS_SCANCODE_BIT (1 << 9)
+#define MAX_AXIS_RANGE 85.0
```
Kenix3:
```suggestion
#define MAX_AXIS_RANGE 85.0f
```
We use floats in most places and not doubles.

src/controller/SDLController.cpp
briaguya0: i think it'd be nice to just #define `32767.0f` as `MAX_SDL_RANGE` or something
```c
#define MAX_SDL_RANGE (float)INT16_MAX
```

src/controller/Controller.cpp
dcvz: I checked all backends the other one being the WiiU they both use these values as can be seen here: https://github.com/Kenix3/libultraship/pull/113/files#diff-71078cb1485704c3ad0458737361bae29b8b6bb59ceee4926e5053cb0226d6f2L164

Also.. unfortunately deadlines were never treated separately (even if there were those warnings). If you take a look at the previous definitions of Normalise() you'll see only one deadline value was passed. I would not consider fixing that part of the scope of this PR.

src/controller/Controller.cpp
dcvz: That was redundant, it is not needed - mapping of values is already happening in the first part.

src/controller/SDLController.cpp
dcvz: I can keep them - as mentioned in a previous comment though, they're were not being treated separately before.

dcvz: One thing to note however, is that deadlines are no longer being processed at this level - they're being processed in `Controller.h` would you still like me to re-add the logs?

src/controller/Controller.cpp
Kenix3: I'm OK with the answer being that deadzones are not implemented separately yet, but we should keep the warnings for now since the deadzone for x and y should always be the same right now to create a circle.

You checked the backends that _we_ implement. Have you done any research on other controller backends like xinput?

Kenix3: What I'm really asking is this: What if we have a backend like xinput that doesn't have the same range?

src/controller/Controller.cpp
Kenix3: Can you be a little more specific? It seems like the octagonal gating code was different to this code that scales and deadzones.

src/controller/SDLController.cpp
dcvz: regarding your comment @briaguya-ai I thought about doing that.. but then noticed that usually code uses the value literal instead of define.. I wasn't sure if there's some future compatibility reason for that. Shall we do this @Kenix3 ?

src/controller/SDLController.cpp
Kenix3: No, it's ok to keep them removed for now. I'll add a comment for the change we can make for now, even if it's unused.

src/controller/Controller.cpp
Kenix3: Let's pass in both the x and y deadzone here even if we're going to ignore the y deadzone.

src/controller/SDLController.cpp
Kenix3: I think it makes sense to add it in the SDLController header. A similar value as well in the Wii U header.

src/controller/Controller.cpp
dcvz: I've done as requested and I've added the warning and a TODO inside the method as well.

src/controller/Controller.cpp
dcvz: I was looking at the wrong code with the comment, octagonal gate is restored

src/port/wiiu/WiiUGamepad.cpp
```diff
@@ -92,11 +92,13 @@ void WiiUGamepad::ReadFromSource(int32_t virtualSlot) {
...
     if (stickX || stickY) {
-        NormalizeStickAxis(virtualSlot, stickX, stickY, profile->AxisDeadzones[0], false);
+        getLeftStickX(virtualSlot) = stickX;
+        getLeftStickY(virtualSlot) = stickY;
     }
 
     if (camX || camY) {
-        NormalizeStickAxis(virtualSlot, camX, camY, profile->AxisDeadzones[2], true);
+        getRightStickX(virtualSlot) = camX;
+        getRightStickY(virtualSlot) = camYy;
```
GaryOderNichts:
```suggestion
        getRightStickY(virtualSlot) = camY;
```

src/port/wiiu/WiiUController.cpp
```diff
@@ -155,17 +155,19 @@ void WiiUController::ReadFromSource(int32_t virtualSlot) {
...
     if (stickX || stickY) {
-        NormalizeStickAxis(virtualSlot, stickX, stickY, profile->AxisDeadzones[0], false);
+        getLeftStickX(virtualSlot) = stickX;
+        getLeftStickY(virtualSlot) = stickY;
     }
 
     if (camX || camY) {
-        NormalizeStickAxis(virtualSlot, camX, camY, profile->AxisDeadzones[2], true);
+        getRightStickX(virtualSlot) = camX;
+        getRightStickY(virtualSlot) = camYy;
```
GaryOderNichts:
```suggestion
        getRightStickY(virtualSlot) = camY;
```

GaryOderNichts (changes requested):

src/controller/Controller.cpp
```diff
@@ -19,6 +20,103 @@ Controller::Controller() : mIsRumbling(false) {
     }
 }
 
+int8_t Controller::ReadStick(int32_t virtualSlot, Stick stick, Axis axis) {
+    switch (stick) {
```
GaryOderNichts, line 24: Missing `break;`s after each case. This causes unintentional fallthroughs.

dcvz:
@Kenix3 this is ready! verified in SDL and Wii U now

Kenix3 (approved):

PR #115, opened 2023-02-14
ResourceMgr now uses a thread pool and we are no longer caching OtrFile structs.

Kenix3:


src/resource/Resource.h
```diff
@@ -23,20 +23,12 @@ class Resource {
     uint32_t ResourceVersion;
     uint64_t Id;
     bool IsDirty = false;
-    std::shared_ptr<OtrFile> File;
+    std::string Path = "";
+    // std::shared_ptr<OtrFile> File;
```
dcvz: This can go away, right?

src/resource/Resource.h
```diff
@@ -1,7 +1,7 @@
 #pragma once
 
 #include <stdint.h>
-#include "OtrFile.h"
+//#include "OtrFile.h"
```
dcvz: This can go away right?

src/resource/ResourceLoader.cpp
```diff
@@ -67,9 +67,10 @@ std::shared_ptr<Resource> ResourceLoader::LoadResource(std::shared_ptr<OtrFile>
     }
 
     if (result != nullptr) {
-        result->File = fileToLoad;
+        // result->File = fileToLoad;
```
dcvz: this can go away right?

briaguya0:
tested via https://github.com/HarbourMasters/Shipwright/pull/2467 on Linux with a freshly generated `.otr`, saw this
image

edit: could not reproduce the issue using the windows build in wine with either DirectX or OpenGL, tried reopening the linux build and could no longer repro...

edit 2: tested on switch and everything seems to be working as expected

src/resource/Resource.h
Kenix3: I was planning to keep these for now since I'm not sure if I'll allow authentic resources to cache the OtrFile under some configurations.

src/resource/Resource.h
Kenix3: I was planning to keep these for now since I'm not sure if I'll allow authentic resources to cache the OtrFile under some configurations.

src/resource/ResourceLoader.cpp
Kenix3: I was planning to keep these for now since I'm not sure if I'll allow authentic resources to cache the OtrFile under some configurations.

briaguya0 (approved):

PR #118, opened 2023-02-16
Resource Manager now does a OTR signature check in the LoadResourceProcess function because it can be called from the game thread. Also, Archive now correctly stores mMainPath. Also, Archive now correctly passes std::string to SPDLOG functions.

Kenix3:


briaguya0 (approved):
lol, i see what you mean about the long PR title

anyways LGTM :shipit:

PR #120, opened 2023-02-17
[Speech] Allow Setting Language + Allow Uninitialization 

dcvz:
Allows specifying the language to use for the given text in the format: `en-US`.
Also creates a basically `SpeechSynthesizerSpeakEnglish()` method in case you don't want to specify language.

### DIFF
--- a/src/core/bridge/speechsynthesizerbridge.h
+++ b/src/core/bridge/speechsynthesizerbridge.h
@@ -17,7 +17,6 @@ extern "C" {
 bool SpeechSynthesizerInit(void);
 void SpeechSynthesizerUninitialize(void);
 
-void SpeechSynthesizerSpeakEnglish(const char* text);
 void SpeechSynthesizerSpeak(const char* text, const char* language);
 
 #ifdef __cplusplus
--- a/src/core/bridge/speechsynthesizerbridge.h
+++ b/src/core/bridge/speechsynthesizerbridge.h
@@ -15,6 +15,7 @@ extern "C" {
 #endif
 
 bool SpeechSynthesizerInit(void);
+void SpeechSynthesizerUninitialize(void);
 
 void SpeechSynthesizerSpeakEnglish(const char* text);
 void SpeechSynthesizerSpeak(const char* text, const char* language);
--- a/src/core/bridge/speechsynthesizerbridge.h
+++ b/src/core/bridge/speechsynthesizerbridge.h
@@ -15,7 +15,9 @@ extern "C" {
 #endif
 
 bool SpeechSynthesizerInit(void);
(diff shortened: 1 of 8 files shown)


src/core/bridge/speechsynthesizerbridge.h
```diff
@@ -15,7 +15,10 @@ extern "C" {
 #endif
 
 bool SpeechSynthesizerInit(void);
-void SpeechSynthesizerSpeak(const char* text);
+void SpeechSynthesizerUninitialize(void);
+
+void SpeechSynthesizerSpeakEnglish(const char* text);
```
Kenix3: I'm not sure this function is needed. Do we really want to set up a different function for each language? We might as well use the more general function that already takes in a language code.

Do you think you need different code to be run per each language?

dcvz: I actually prefer to not have it.. but since we don’t have default parameters I thought it’d be best to give people an option where they don’t have to specify a language as most use-cases might just be english

Kenix3: Then in that case the preference should be a default parameter defined in the header.

dcvz: But C doesn’t support default parameters, right?

dcvz: If it’s alright with you, I’ll just remove it then?

Kenix3: You're right that C doesn't support it. C++ does and I was confusing it for code in the class.
So let's look at this in perspective. This function is part of a public API in the bridge headers. Is a function for English only something we want to continue to support in 5 years? I think not. I think that also supersedes convenience on not passing in the language.

That doesn't mean we can't still use a "default" value. What about setting a CVar and/or a context variable on the speed synthesizer itself? You can set the context only once and then all calls will use that language.

dcvz: I’d say let’s remove the method. I don’t think we should be using a cvar or context variable on the synthesiser itself. There’s usefulness on the API being flexible like this. The work I’m doing on SoH already benefits from it.

Thinking some more on it.. most ports will be dealing with language as the games support them. So they’ll likely always want to specify it anyway. I would remove the method.

Kenix3: I'm happy with just removing it, but I really do not see the issue with having a context variable on the actual synthesizer class and a new method to change the context.

dcvz: Method has been removed!

Kenix3 (approved):

PR #125, opened 2023-02-19
Fix for vanishing dirt roads on higher resolution

Bagbar:
this should make the behavior of the dirt patches/roads consistent in all resolutions and resolution scalings.
Changing the SSDBfactor will influence the behavior 
120 corresponds to the previous value -2 on 240p


Kenix3:
Looks like clang-format wasn't run on this PR.

```
find src include -name "*.cpp" -o -name "*.h" | sed 's| |\\ |g' | xargs clang-format-12 -i
```
Here's a command for Linux that can be run from within wsl.

Bagbar:
Closing because improved version (with CVar) will be made available

PR #127, opened 2023-02-21
Fix a couple of problems indicated by compiler warnings

nngeorgiy:
This PR dramatically reduces the number of warnings introduced by `libultraship` while building `shipwright` for Ubuntu 22.04/amd64. By 'dramatically' I mean that the error log is down from 13MB to 3MB, with most warnings due to conflicting definitions in `libultra/hardware.h`, a file that wasn't even in the original SDK. While I was at it, I also fixed at least four extremely minor bugs. Cheers!

### DIFF
--- a/include/libultraship/libultra/hardware.h
+++ b/include/libultraship/libultra/hardware.h
@@ -1,90 +1,9 @@
 #ifndef ULTRA64_HARDWARE_H
 #define ULTRA64_HARDWARE_H
 
-#define HW_REG(reg, type) *(volatile type*)((reg) | 0xa0000000)
+/* TODO: This file isn't a part of the original SDK headers, so either remove
+it completely or have it used to include a useful subset of headers */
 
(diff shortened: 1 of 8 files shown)


include/libultraship/libultra/hardware.h
```diff
@@ -1,90 +1,9 @@
 #ifndef ULTRA64_HARDWARE_H
 #define ULTRA64_HARDWARE_H
 
-#define HW_REG(reg, type) *(volatile type*)((reg) | 0xa0000000)
+/* TODO: This file isn't a part of the original SDK headers, so either remove
```
Kenix3, line 4: `hardware.h` should probably just be removed

Kenix3 (changes requested):
`hardware.h` should probably just be removed

dcvz (approved):

Kenix3 (approved):

PR #128, opened 2023-02-21
[Speech] [Windows] Fix: Special character support

dcvz:


src/speechsynthesizer/SAPISpeechSynthesizer.cpp
```diff
@@ -31,19 +32,24 @@ void SAPISpeechSynthesizer::DoUninitialize() {
...
-    wchar_t wtext[w];
-    mbstowcs(wtext, speak, strlen(speak) + 1);
+    const size_t textSize = strlen(text) + 1;
+    wchar_t* wText = new wchar_t[textSize];
+    mbstowcs(wText, text, textSize);
 
-    mVoice->Speak(wtext, SPF_IS_XML | SPF_ASYNC | SPF_PURGEBEFORESPEAK, NULL);
+    auto speakText = fmt::format(
+        L"<speak version='1.0' xmlns='http://www.w3.org/2001/10/synthesis' xml:lang='{}'>{}</speak>", wLanguage, wText);
+    mVoice->Speak(speakText.c_str(), SPF_IS_XML | SPF_ASYNC | SPF_PURGEBEFORESPEAK, NULL);
```
Kenix3, line 47: What happens to the memory allocated into `textSize` and `wLanguage` if this function (`mVoice->Speak`), or any other, throws an exception? I guess for now it doesn't matter because the only catch is going to be in the crash handler. Just wanted to point this out.

src/speechsynthesizer/SAPISpeechSynthesizer.cpp
```diff
@@ -31,19 +32,24 @@ void SAPISpeechSynthesizer::DoUninitialize() {
...
-    int* wp = const_cast<int*>(&w);
-    *wp = strlen(speak);
+    std::string locale = fmt::format("{}.UTF-8", language);
+    locale.replace(locale.find("-"), 1, "_");
+    std::setlocale(LC_ALL, locale.c_str());
 
-    wchar_t wtext[w];
-    mbstowcs(wtext, speak, strlen(speak) + 1);
+    const size_t textSize = strlen(text) + 1;
+    wchar_t* wText = new wchar_t[textSize];
```
Kenix3: Does the `std::wstring` class not work?

dcvz: It was not working (and I tried for a while), I went on different iterations of this -- this one works.

src/speechsynthesizer/SAPISpeechSynthesizer.cpp
dcvz: This method does not throw an exception but rather returns a fail if it did fail

src/speechsynthesizer/SAPISpeechSynthesizer.cpp
dcvz: Thought of another way to do it via `wstring` in the end

src/speechsynthesizer/SAPISpeechSynthesizer.cpp
dcvz: This is simplified now with the use of `wstring`

Kenix3: It's actually less about whether or not it does throw an exception and more that someone could add code which could throw an exception and thus cause a memory leak. :)

std::wstring should alleviate this concern.

Kenix3:
Let me make sure I understand this correctly. UTF8 encoded strings are passed into SpeakThreadTask. You are creating a std::string from that, then using Windows functions to get it's wchar length and then using Windows functions to translate it from UTF8 to wchar?

dcvz:
> Let me make sure I understand this correctly. UTF8 encoded strings are passed into SpeakThreadTask. You are creating a std::string from that, then using Windows functions to get it's wchar length and then using Windows functions to translate it from UTF8 to wchar?

Correct! The given char* is already encoded correctly but we need it in wchar for the speech API - but I do it via this method because of how Windows and UTF8 support work.

dcvz:
@Kenix3 I further tweaked this and I think it make even more sense, so now we actually just convert to string when we receive the char buffers, before passing them to the speech thread.

This way the callers don't have to worry about the lifecycle of the char buffers - if they're doing `c_str()` on a string or so they don't have to try to dup. I think this is the nicest solution in the end. 

We still do the conversion to wstring but that's after the fact and because of how the speech API works

PR #132, opened 2023-02-23
Refactor Config File Handling

dcvz:


src/misc/JsonConfig.cpp
```diff
@@ -0,0 +1,187 @@
...
+                for (size_t j = 0; j < i - 1; j++) {
+                    currentJson = &(*currentJson)[keyParts[j]];
+                }
+                (*currentJson).erase(keyParts[i - 1]);
+            }
+        }
+    } else {
+        // not nested
+        mJson.erase(key);
+    }
```
briaguya0: i'd invert this to avoid having everything deep in the `if`, so just

```cpp
if (keyParts.size() == 1) {
    // not nested
    mJson.erase(key);
    return;
}

[everything from the if block]
```

src/misc/JsonConfig.cpp
```diff
@@ -0,0 +1,187 @@
...
+    return mIsNewConfig;
+}
+
+void JsonConfig::DeleteEntry(std::string key) {
+    auto keyParts = StringHelper::Split(key, ".");
+
+    // clear nested key, if after deletion parent is empty, delete it too
+    if (keyParts.size() > 1) {
+        json* currentJson = &mJson;
+        for (size_t i = 0; i < keyParts.size() - 1; i++) {
```
briaguya0: could we just use
```cpp
for (auto keyPart : keyParts) {
   [...]
}
```
or 
```cpp
for (auto& keyPart : keyParts) {
   [...]
}
```
instead since `StringHelper::Split` gets us a `vector`?

src/misc/JsonConfig.cpp
```diff
@@ -0,0 +1,187 @@
...
+            if (!currentJson->contains(keyParts[i])) {
+                return;
+            }
```
briaguya0: it'd be nice to explain why we might get to a point where `currentJson` does not contain the `keyPart` here, and why returning early is the proper way to handle that situation

src/misc/JsonConfig.cpp
```diff
@@ -0,0 +1,187 @@
...
+bool JsonConfig::IsNewConfig() {
+    return mIsNewConfig;
+}
+
+void JsonConfig::DeleteEntry(std::string key) {
+    auto keyParts = StringHelper::Split(key, ".");
+
+    // clear nested key, if after deletion parent is empty, delete it too
+    if (keyParts.size() > 1) {
+        json* currentJson = &mJson;
```
briaguya0: not sure if/how this applies, but just want to make sure you've looked into [nlohmann json pointers](https://json.nlohmann.me/api/json_pointer/#see-also)

src/misc/JsonConfig.cpp
```diff
@@ -0,0 +1,187 @@
...
+void JsonConfig::SetInteger(std::string key, int32_t value) {
+    return SetArbitraryType(key, value);
+}
+
+int32_t JsonConfig::GetInteger(std::string key, int32_t defaultValue) {
+    json::value_type value = GetArbitraryType(key);
+    if (value.is_number_integer()) {
+        return value;
+    } else {
+        return defaultValue;
```
briaguya0, line 104: i assume this is to handle the case where someone manually edited the json and put a value that doesn't parse to the correct type in there (no need to change anything, just hoping to confirm the reasoning behind these)

briaguya0 (commented):
This is all looking very good! I left a few comments/questions (mostly in `DeleteEntry` but some of the suggestions apply to multiple methods within `JsonConfig`)

I also haven't tested this out yet, some "stress test" use cases I can think of:
* Cosmetics editor (randomizing all, locking, resetting)
* Randomizer menu (changing settings, generating seeds, ensuring the settings properly make it into the spoiler)
* Applying settings presets
* Controller config

Overall this feels like a huge step in the right direction!

src/misc/JsonConfig.cpp
briaguya0: i guess that could be problematic in the places you're using `keyParts[i - 1]`

src/misc/JsonConfig.cpp
dcvz: It was actually a bug 😅 we can get a `nullptr` from GetArbitraryType there should be a null check here too. I've added it now.

src/misc/JsonConfig.cpp
```diff
@@ -0,0 +1,173 @@
...
+    return GetArbitraryType(key);
+}
+
+void JsonConfig::SetInteger(std::string key, int32_t value) {
+    return SetArbitraryType(key, value);
+}
+
+int32_t JsonConfig::GetInteger(std::string key, int32_t defaultValue) {
+    json::value_type value = GetArbitraryType(key);
+    if (value && value.is_number_integer()) {
```
Archez, line 101: For here and all the ones below, seems like you need to explicitly check against nullptr, as just checking itself as truthy when it is a valid json value is actually executing `.get<false>()` behind the scenes, which will throw for any non boolean types.
```suggestion
    if (value != nullptr && value.is_number_integer()) {
```

src/misc/JsonConfig.cpp
```diff
@@ -0,0 +1,173 @@
...
+            if (parent.is_array()) {
+                auto index = std::stoi(keyToDelete);
+                parent.erase(index);
+            } else {
+                parent.erase(keyToDelete);
+            }
+
+            // set variable for whether we should keep recursing parents
+            shouldStop = parent.empty() || parentKey == currentKey;
+            currentKey = parentKey;
```
Archez, line 83: I think the `parentKey == currentKey` check should be moved to the start of the while loop as an early break/return.

As it is now, if you pass in a key like `xxx.yyy`, it will delete `yyy` from `xxx`, but if `xxx` is not empty, it will loop one more time and attempt to do a delete for `xxx` on `xxx`. which would be bad if there was technically a real scenario of `xxx.xxx` containing a value.

src/misc/JsonConfig.cpp
```diff
@@ -0,0 +1,173 @@
...
+            parentKey = currentKey.substr(0, currentKey.find_last_of('.'));
+            auto parentPtr = DotKeyToPointer(parentKey);
+
+            // grab parent object at key and delete the current key
+            auto& parent = mJson.at(parentPtr);
+            auto keyToDelete = currentKey.substr(currentKey.find_last_of('.') + 1);
+
+            // check if parent object is an array and delete index
+            if (parent.is_array()) {
+                auto index = std::stoi(keyToDelete);
```
Archez, line 75: stoi can throw exceptions. Should those be caught and handled with the rest of the try catch here?

Kenix3: We have some helpers in the LUS util as well that will handle the exception.

Kenix3:
Closed as dcvz has communicated he will not be maintaining this PR anytime soon.

PR #134, opened 2023-02-24
Fix opengl api not getting used correctly

Archez:
The early return was removed and prevented the open gl api from being used on windows. https://github.com/HarbourMasters/Shipwright/issues/2534

I also renamed the dx11 string back to lower case to maintain compatibility with existing json files.

### DIFF
--- a/src/core/Window.cpp
+++ b/src/core/Window.cpp
@@ -323,15 +323,14 @@ void Window::InitializeWindowManager(std::string_view gfxBackend, std::string_vi
 #endif
 #if defined(ENABLE_OPENGL) || defined(__APPLE__)
     if (gfxBackend == "sdl") {
-        if (gfxApi == "OpenGL") {
-            mRenderingApi = &gfx_opengl_api;
-        }
+        mRenderingApi = &gfx_opengl_api;
+        mWindowManagerApi = &gfx_sdl;
 #ifdef __APPLE__
-        else if (gfxApi == "Metal") {
+        else if (gfxApi == "Metal" && Metal_IsSupported()) {
             mRenderingApi = &gfx_metal_api;
         }
 #endif
-        mWindowManagerApi = &gfx_sdl;
+        return;
     }
 #if defined(__linux__) && defined(X11_SUPPORTED)
     if (gfxBackend == "glx") {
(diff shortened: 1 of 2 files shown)


src/core/Window.cpp
```diff
@@ -323,15 +323,14 @@ void Window::InitializeWindowManager(std::string_view gfxBackend, std::string_vi
 #endif
 #if defined(ENABLE_OPENGL) || defined(__APPLE__)
     if (gfxBackend == "sdl") {
-        if (gfxApi == "OpenGL") {
```
dcvz, line 326: Why did this if get removed? This is for overriding the setting with what you may have in the config, we don't have to default to GL here.

src/menu/ImGuiImpl.cpp
```diff
@@ -111,7 +111,7 @@ const char* filters[3] = {
 
 std::vector<std::pair<const char*, const char*>> renderingBackends = {
 #ifdef _WIN32
-    { "Dx11", "DirectX" },
+    { "dx11", "DirectX" },
```
dcvz, line 114: I think Kenix updating this one, it may want to be kept. To be honest should probably be comparing lowercased strings.

dcvz (commented):
Looks like I missed that return! :(

src/menu/ImGuiImpl.cpp
Archez: My guess is it was a search and replace with trying to make the instance variable capitalized, but this string is just regarding the dropdown and comparing against the json config. Trying to not break compatibility with old json files (although technically it was falling back to dx11 anyways so it fails "safely")

src/core/Window.cpp
Archez: I removed it because it breaks old json compatibility for people that dont have `GfxAPI` stored in their json file and causes a crash. Basically if you are set to SDL it should always default to use opengl. Only if you are apple and have it set to metal does it change to metal instead.

There are only two sdl apis, OpenGL and now Metal. There needs to be a default, why wouldn't it be OpenGL?

src/menu/ImGuiImpl.cpp
Kenix3: Should be set to whatever it was a month ago. I did indeed do a find and replace.

Kenix3 (approved):

PR #135, opened 2023-02-24
Set notch prox angle onto device profile instead of cvar

Archez:
This moves the notch proximity value to be set on the device profile so that each device can have unique notch angles and matches the rest of controller settings.

This change also removes a broken if statement that was causing CVar save on next tick on every frame, cause performance issues when the controller window was open.

I increased the max window height to account for the notch angle slider without needing to scroll.


dcvz (approved):
LGTM 🚀

Kenix3 (approved):

PR #136, opened 2023-02-26
gfx_pc: Removed clock_gettime-call for non-windows platforms

tdc22:
While checking the results of perf I found some instances of calls to `clock_gettime` within `gfx_flush` even with some high optimization settings (-O3, release). The code is old and propably unused, but in case somebody still uses that to debug/profile rendering code for now I just added the condition, that the debug flag is set. That change removes that call for Linux / Non-Windows platforms in release builds now.

### DIFF
--- a/src/graphic/Fast3D/gfx_pc.cpp
+++ b/src/graphic/Fast3D/gfx_pc.cpp
@@ -831,10 +831,10 @@ static void import_texture_ci8(int tile) {
         result_line_size *= metadata->h_byte_scale;
     }
 
-    uint32_t width = rdp.texture_tile[tile].line_size_bytes;
-    uint32_t height = size_bytes / rdp.texture_tile[tile].line_size_bytes;
+    uint32_t width = result_line_size;
+    uint32_t height = size_bytes / result_line_size;
 
-    gfx_rapi->upload_texture(rgba32_buf, width, height);
+    gfx_rapi->upload_texture(tex_upload_buffer, width, height);
     // DumpTexture(rdp.loaded_texture[rdp.texture_tile[tile].tmem_index].otr_path, rgba32_buf, width, height);
 }
 
@@ -1424,10 +1424,6 @@ static void gfx_sp_tri1(uint8_t vtx1_idx, uint8_t vtx2_idx, uint8_t vtx3_idx, bo
         if (clip_parameters.z_is_from_0_to_1) {
             z = (z + w) / 2.0f;
         }
-
-        // if (markerOn) {
-        //     z = 10;
-        //}
 
         buf_vbo[buf_vbo_len++] = v_arr[i]->x;
         buf_vbo[buf_vbo_len++] = clip_parameters.invert_y ? -v_arr[i]->y : v_arr[i]->y;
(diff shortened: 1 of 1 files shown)


dcvz:
Nice catch @tdc22! The way things stand right now `clock_gettime()`'s value is never used. We fetch the time for no reason. It used to be used for some debug messages but those are commented out as well. 

I think the best thing is to remove it all. Unless we want to enable those logs only `DEBUG`.

tdc22:
Ok, I removed all instances of `get_time` as well as the related commented debug code. I also stumbled upon multiple instances where a variable `int bp = 0` was initialized, but never used. I removed all of them as they seem to have no practical purpose and were maybe used to debug something once, too. Usually it was set depending on the int `markerOn` which I left in since in one case it calls `gfx_s2dex_bg_copy` which seems to do something. The two unused variables `dListBP` and `matrixBP` seemed to be related to this as well and got deleted.

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -1368,9 +1327,9 @@ static void gfx_sp_tri1(uint8_t vtx1_idx, uint8_t vtx2_idx, uint8_t vtx3_idx, bo
             z = (z + w) / 2.0f;
         }
 
-        if (markerOn) {
-            // z = 10;
-        }
+        //if (markerOn) {
+        //     z = 10;
+        //}
```
dcvz: I'd leave this uncommented since this actually does change behaviour and there's a command to set marker to on

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -2405,7 +2347,6 @@ static void gfx_run_dl(Gfx* cmd) {
                 break;
 #ifdef F3DEX_GBI_2
             case G_QUAD: {
-                int bp = 0;
```
dcvz, line 2589: bp = 0 is just a way to get you a line that you can set a breakpoint at in your IDE :)

src/graphic/Fast3D/gfx_pc.cpp
tdc22: Check the change again. z = 10 was commented out before, too. I just did the same to the surrounding if. My change should have no functional impact - unless I'm missing something.

src/graphic/Fast3D/gfx_pc.cpp
tdc22: So should I leave all the bp = 0 in or can they go? I mean this is a fine hack for personal usage but I wouldn't commit them to the repo. Right?

dcvz: was just explaining :)

dcvz (approved):

src/graphic/Fast3D/gfx_pc.cpp
Kenix3: Yes, this is fine to remove. :) Some of these got snuck in unfortunately.

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -1368,9 +1327,9 @@ static void gfx_sp_tri1(uint8_t vtx1_idx, uint8_t vtx2_idx, uint8_t vtx3_idx, bo
             z = (z + w) / 2.0f;
         }
 
-        if (markerOn) {
-            // z = 10;
-        }
+        // if (markerOn) {
```
Kenix3: Can we just remove this?

tdc22: Done

Kenix3 (approved):

PR #138, opened 2023-03-01
Custom model (and XML resources) support

NEstelami:


include/libultraship/libultra/gbi.h
```diff
@@ -1751,11 +1757,16 @@ typedef union {
 #define gsSPViewport(v) gsDma1p(G_MOVEMEM, (v), sizeof(Vp), G_MV_VIEWPORT)
 #endif /* F3DEX_GBI_2 */
 
+#define gsSPPushCD(pkt, dl) gDma1p(pkt, G_PUSHCD, dl, 0, G_DL_PUSH)
 #define __gSPDisplayList(pkt, dl) gDma1p(pkt, G_DL, dl, 0, G_DL_PUSH)
 #define gsSPDisplayList(dl) gsDma1p(G_DL, dl, 0, G_DL_PUSH)
+#define gsSPDisplayListOTR(dl) gsDma1p(G_DL_OTR, dl, 0, G_DL_PUSH)
+#define gsSPDisplayListOTR2(dl) gsDma1p(G_DL_OTR2, dl, 0, G_DL_PUSH)
```
Kenix3: Why are there two versions of this display list command?

src/binarytools/BinaryReader.cpp
```diff
@@ -192,6 +192,18 @@ std::string Ship::BinaryReader::ReadString() {
...
 }
 
+std::string Ship::BinaryReader::ReadCString() {
+    std::string res;
+
+    unsigned char c = 0;
+    do {
+        c = ReadChar();
+        res += c;
+    } while (c != '\0');
```
Kenix3, line 205: What happens if there are no more characters in the buffer?

src/binarytools/FileHelper.h
```diff
@@ -0,0 +1,87 @@
...
+	{
+		std::ifstream file(filePath, std::ios::in | std::ios::binary | std::ios::ate);
+
+		if (!file)
+			return std::vector<uint8_t>();
+
+		int32_t fileSize = (int32_t)file.tellg();
+		file.seekg(0);
+		char* data = new char[fileSize];
+		file.read(data, fileSize);
```
Kenix3: What happens if an exception occurs between the new and delete keywords in this function?

src/binarytools/FileHelper.h
```diff
@@ -0,0 +1,87 @@
...
+
+		return result;
+	};
+
+	static std::string ReadAllText(const fs::path& filePath)
+	{
+		std::ifstream file(filePath, std::ios::in | std::ios::binary | std::ios::ate);
+		int32_t fileSize = (int32_t)file.tellg();
+		file.seekg(0);
+		char* data = new char[fileSize + 1];
```
Kenix3: What happens if an exception occurs between the new and delete keywords in this function?

src/core/bridge/resourcebridge.cpp
```diff
@@ -227,4 +236,27 @@ void WriteTextureDataInt16ByCrc(uint64_t crc, size_t index, int16_t valueToWrite
...
+uint8_t* GetTextureModifier(const char* texPath) {
+    if(mTextureModifiers.contains(texPath)){
+        return mTextureModifiers[texPath];
+    }
+    return nullptr;
+}
+
+void RemoveTextureModifier(const char* texPath) {
+    if(mTextureModifiers.contains(texPath)){
+        free(mTextureModifiers[texPath]);
```
Kenix3: Why are we freeing memory here? We're not allocating it as part of this system. If this system frees the memory, it should allocate the memory.

src/core/bridge/resourcebridge.cpp
```diff
@@ -5,7 +5,7 @@
 #include <algorithm>
 #include <StrHash64.h>
 
-#include "resource/type/Texture.h"
+std::unordered_map<std::string, uint8_t*> mTextureModifiers;
```
Kenix3: What's the purpose of this data?

src/core/bridge/resourcebridge.cpp
```diff
@@ -23,6 +23,15 @@ std::shared_ptr<Ship::Resource> LoadResource(uint64_t crc, bool now) {
     return LoadResource(name, now);
 }
 
+std::shared_ptr<Ship::Texture> GetResourceTexByName(const char* name) {
```
Kenix3: I don't think we should be exposing a texture specific resource get function. We should do the appropriate logic in the ResourceLoader.

In other words, the calling code should be responsible for casting to Ship::Texture

src/graphic/Fast3D/gfx_direct3d11.cpp
```diff
@@ -390,6 +390,10 @@ void CSMain(uint3 DTid : SV_DispatchThreadID) {
     SohImGui::Init(window_impl);
 }
 
+static int gfx_d3d11_get_max_texture_size() {
+    return 16384;
```
Kenix3: Isn't this hardware dependent? Doesn't d3d have a way to query this? We should be doing that rather than hard coding.

src/graphic/Fast3D/gfx_gx2.cpp
```diff
@@ -135,6 +135,10 @@ static const char* gfx_gx2_get_name() {
     return "GX2";
 }
 
+static int gfx_gx2_get_max_texture_size() {
+    return 8192;
```
Kenix3, line 140: This can probably be hardcoded since it's a console, but I'd assume gx2 also has a way to query this.

src/graphic/Fast3D/gfx_opengl.cpp
```diff
@@ -87,6 +87,12 @@ static FilteringMode current_filter_mode = FILTER_THREE_POINT;
 GLuint pixel_depth_rb, pixel_depth_fb;
 size_t pixel_depth_rb_size;
 
+static int gfx_opengl_get_max_texture_size() {
+    GLint max_texture_size;
+    glGetIntegerv(GL_MAX_TEXTURE_SIZE, &max_texture_size);
```
Kenix3, line 93: We should do this for the other renderers.

src/resource/Archive.cpp
```diff
@@ -77,43 +78,60 @@ std::shared_ptr<OtrFile> Archive::LoadFileFromHandle(const std::string& filePath
         mpqHandle = mMainMpq;
     }
 
-    bool attempt = SFileOpenFileEx(mpqHandle, filePath.c_str(), 0, &fileHandle);
+    // TEST - Read from filesystem first, then the MPQ archive...
```
Kenix3: I think the code to read files from the file system should probably be behind `#IFDEF _DEBUG`

src/resource/Archive.cpp
```diff
@@ -77,43 +78,60 @@ std::shared_ptr<OtrFile> Archive::LoadFileFromHandle(const std::string& filePath
...
+            return fileToLoad;
+        }
 
-    if (!SFileReadFile(fileHandle, fileData.get(), fileSize, &countBytes, NULL)) {
-        SPDLOG_ERROR("({}) Failed to read file {} from mpq archive {}", GetLastError(), filePath.c_str(),
-                     mMainPath.c_str());
         if (!SFileCloseFile(fileHandle)) {
-            SPDLOG_ERROR("({}) Failed to close file {} from mpq after read failure in archive {}", GetLastError(),
-                         filePath.c_str(), mMainPath.c_str());
+            SPDLOG_ERROR("({}) Failed to close file {} from mpq archive {}", GetLastError(), filePath.c_str(),
```
Kenix3:
```suggestion
            SPDLOG_ERROR("({}) Failed to close file {} from mpq archive {}", GetLastError(), filePath,
```

src/resource/Archive.cpp
```diff
@@ -77,43 +78,60 @@ std::shared_ptr<OtrFile> Archive::LoadFileFromHandle(const std::string& filePath
...
+                         mMainPath.c_str());
```
Kenix3:
```suggestion
                         mMainPath);
```

src/resource/Archive.cpp
```diff
@@ -198,41 +216,43 @@ bool Archive::RenameFile(const std::string& oldPath, const std::string& newPath)
...
+            bool fileFound;
+            do {
+                fileFound = SFileFindNextFile(hFind, &findContext);
+
+                if (fileFound) {
+                    fileList.push_back(findContext);
+                } else if (!fileFound && GetLastError() != ERROR_NO_MORE_FILES)
+                    // else if (!fileFound)
+                {
+                    SPDLOG_ERROR("({}), Failed to search with mask {} in archive {}", GetLastError(), searchMask.c_str(),
```
Kenix3:
```suggestion
                    SPDLOG_ERROR("({}), Failed to search with mask {} in archive {}", GetLastError(), searchMask,
```

src/resource/Archive.cpp
```diff
@@ -198,41 +216,43 @@ bool Archive::RenameFile(const std::string& oldPath, const std::string& newPath)
...
+                                 mMainPath.c_str());
```
Kenix3:
```suggestion
                                 mMainPath);
```

src/resource/Archive.cpp
```diff
@@ -198,41 +216,43 @@ bool Archive::RenameFile(const std::string& oldPath, const std::string& newPath)
...
+            } while (fileFound);
+        } else if (GetLastError() != ERROR_NO_MORE_FILES) {
+            SPDLOG_ERROR("({}), Failed to search with mask {} in archive {}", GetLastError(), searchMask.c_str(),
                          mMainPath.c_str());
+            return fileList;
+        }
+
+        if (hFind != nullptr) {
+            if (!SFileFindClose(hFind)) {
+                SPDLOG_ERROR("({}) Failed to close file search {} in archive {}", GetLastError(), searchMask.c_str(),
```
Kenix3:
```suggestion
                SPDLOG_ERROR("({}) Failed to close file search {} in archive {}", GetLastError(), searchMask,
```

src/resource/Archive.cpp
```diff
@@ -198,41 +216,43 @@ bool Archive::RenameFile(const std::string& oldPath, const std::string& newPath)
...
+                             mMainPath.c_str());
```
Kenix3:
```suggestion
                             mMainPath);
```

src/resource/ResourceFactory.cpp
```diff
@@ -14,4 +13,7 @@ void ResourceVersionFactory::WriteFileBinary(std::shared_ptr<BinaryWriter> write
 void ResourceVersionFactory::WriteFileXML(std::shared_ptr<tinyxml2::XMLElement> writer,
                                           std::shared_ptr<Resource> resource) {
 }
+std::shared_ptr<Resource> ResourceFactory::ReadResourceXML(uint32_t version, tinyxml2::XMLElement* reader) {
+    return std::shared_ptr<Resource>();
```
Kenix3: Why not return a nullptr here since the resource isn't valid?

src/resource/ResourceLoader.cpp
```diff
@@ -18,20 +18,22 @@ ResourceLoader::ResourceLoader(std::shared_ptr<Window> context) : mContext(conte
 }
 
 void ResourceLoader::RegisterGlobalResourceFactories() {
-    RegisterResourceFactory(ResourceType::Texture, std::make_shared<TextureFactory>());
-    RegisterResourceFactory(ResourceType::Vertex, std::make_shared<VertexFactory>());
-    RegisterResourceFactory(ResourceType::DisplayList, std::make_shared<DisplayListFactory>());
-    RegisterResourceFactory(ResourceType::Matrix, std::make_shared<MatrixFactory>());
-    RegisterResourceFactory(ResourceType::Array, std::make_shared<ArrayFactory>());
-    RegisterResourceFactory(ResourceType::Blob, std::make_shared<BlobFactory>());
+    RegisterResourceFactory(ResourceType::Texture, "Texture", std::make_shared<TextureFactory>());
```
Kenix3, line 21: Why are we including the string name along with the resource type enum? Is it to match what comes out of the XML?

src/resource/ResourceLoader.cpp
```diff
@@ -40,21 +42,53 @@ std::shared_ptr<Window> ResourceLoader::GetContext() {
...
+    auto stream = std::make_shared<MemoryStream>(fileToLoad->Buffer.get(), fileToLoad->BufferSize);
+    auto reader = std::make_shared<BinaryReader>(stream);
     std::shared_ptr<Resource> result = nullptr;
+    ResourceType resourceType;
+    uint64_t id = 0xDEADBEEFDEADBEEF;
 
-    if (fileToLoad != nullptr) {
-        auto stream = std::make_shared<MemoryStream>(fileToLoad->Buffer.data(), fileToLoad->Buffer.size());
-        auto reader = std::make_shared<BinaryReader>(stream);
+    // Determine if file is binary or XML...
```
Kenix3: First byte is endian order. We should be ready this as the second byte to maintain OTR compatibility.

src/resource/ResourceLoader.cpp
```diff
@@ -40,21 +42,53 @@ std::shared_ptr<Window> ResourceLoader::GetContext() {
...
+    uint8_t firstByte = reader->ReadInt8();
 
+    if (firstByte == '<') 
+    {
+        // XML
+        reader->Seek(-1, SeekOffsetType::Current);
```
Kenix3: If it's xml, we should probably still include the OTR header 64 bytes, even if it's unused. Everything that goes in the archive should have an OTR header.

I'm thinking the best design here is to add a new resource type in LUS called XML, and then reading the cstring as part of that code.

src/resource/factory/TextureFactory.cpp
```diff
@@ -12,6 +12,9 @@ std::shared_ptr<Resource> TextureFactory::ReadResource(uint32_t version, std::sh
         case 0:
             factory = std::make_shared<TextureFactoryV0>();
             break;
+        case 4:
```
Kenix3: Should be either 1 for the second version, or 9 for Spock

include/libultraship/libultra/gbi.h
NEstelami: One accesses the display list by hash and the other accesses it by file path.

src/core/bridge/resourcebridge.cpp
NEstelami: I have no idea, I did not write that.

src/resource/ResourceLoader.cpp
NEstelami: We want people to be able to drop in XML files and quickly make edits in a text editor. Wrapping a text format in a OTR header removes that convenience without much benefit.

include/libultraship/libultra/gbi.h
Kenix3: Can we name it based on that please?

src/core/bridge/resourcebridge.cpp
Kenix3: @KiritoDv @dcvz

src/resource/ResourceLoader.cpp
NEstelami: Yup, that's how it checks which resource types are in the XML.

src/core/bridge/resourcebridge.cpp
dcvz: Oh, just saw that this is now based on our stuff. I think we can remove all the `TextureModifier` stuff - it was one exploration we were doing at tackling runtime modified textures. But ultimately we want to go with our proposed approach of doing it at the GPU.

@Jack-Walker could you invite me to your repo? I can do that cleanup and I guess perhaps more things will come up that I can/should help with.

src/core/bridge/resourcebridge.cpp
dcvz: I explained more in detail here: https://github.com/Kenix3/libultraship/pull/138#discussion_r1121357731

But this can go away, it was a system we explored for allowing runtime modified textures. But it doesn't scale with higher resolutions so we'll need the GPU based approach we'd been talking about.

src/graphic/Fast3D/gfx_direct3d11.cpp
dcvz: I could not find an appropriate way to query this - as far as I could see all d3d11 API has this limit (which doesn't really make sense given GPU's are different). If someone knows the best way to do this, that's great.

I'm also not a Windows person.

src/graphic/Fast3D/gfx_gx2.cpp
dcvz: @GaryOderNichts do you know if this is something we can query? I think hardcoding is also not wrong here though since it's known.

src/graphic/Fast3D/gfx_opengl.cpp
dcvz: Sure, I obviously wanted to. I've answered elsewhere already but its not always known whether we can or how.

src/resource/factory/TextureFactory.cpp
dcvz: Yeah, we started building this some time ago already..

include/libultraship/libultra/gbi.h
```diff
@@ -1751,11 +1757,16 @@ typedef union {
...
 
 #define gSPBranchList(pkt, dl) gDma1p(pkt, G_DL, dl, 0, G_DL_NOPUSH)
 #define gsSPBranchList(dl) gsDma1p(G_DL, dl, 0, G_DL_NOPUSH)
+#define gsSPBranchListOTR(dl) gsDma1p(G_DL_OTR, dl, 0, G_DL_NOPUSH)
+#define gsSPBranchListOTR2(dl) gsDma1p(G_DL_OTR2, dl, 0, G_DL_NOPUSH)
```
dcvz: same comment as above here on, let's use clear naming

include/libultraship/libultra/gbi.h
```diff
@@ -2844,6 +2858,12 @@ typedef union {
                                    G_ACMUX_##Ad1))                                                                 \
     }
 
+#define gsDPSetCombineLERP2(a0, b0, c0, d0, Aa0, Ab0, Ac0, Ad0, a1, b1, c1, d1, Aa1, Ab1, Ac1, Ad1) \
```
dcvz: Same comment about naming based on difference

src/resource/Archive.cpp
```diff
@@ -75,33 +76,60 @@ std::shared_ptr<OtrFile> Archive::LoadFileFromHandle(const std::string& filePath
         mpqHandle = mMainMpq;
     }
 
-    bool attempt = SFileOpenFileEx(mpqHandle, filePath.c_str(), 0, &fileHandle);
+    // TEST - Read from filesystem first, then the MPQ archive...
```
dcvz: This method `LoadFileFromHandle` has lost the new file system loading mechanism. @Jack-Walker you'll likely have to take the base from the current main branch and try to fit your XML changes into it.

src/resource/factory/DisplayListFactory.cpp
```diff
@@ -54,4 +77,908 @@ void DisplayListFactoryV0::ParseFileBinary(std::shared_ptr<BinaryReader> reader,
...
+            g.words.w1 |= v01 << 16;
+            g.words.w1 |= v02 << 0;
+        } else if (childName == "Triangles2") {
+            g = gsSP2Triangles(child->IntAttribute("V00"), child->IntAttribute("V01"), child->IntAttribute("V02"),
+                               child->IntAttribute("Flag0"), child->IntAttribute("V10"), child->IntAttribute("V11"),
+                               child->IntAttribute("V12"), child->IntAttribute("Flag1"));
+        } else if (childName == "LoadVertices") {
+            std::string fName = child->Attribute("Path");
+            // fName = ">" + fName;
+            g = { gsSPVertex(0, child->IntAttribute("Count"), child->IntAttribute("VertexBufferIndex"),
```
dcvz: On my machine, it says `gsSPVertex` is taking too many parameters. Is there perhaps a definition missing from `gbi.h`?

src/resource/ResourceLoader.cpp
```diff
@@ -40,21 +42,53 @@ std::shared_ptr<Window> ResourceLoader::GetContext() {
 }
 
 std::shared_ptr<Resource> ResourceLoader::LoadResource(std::shared_ptr<OtrFile> fileToLoad) {
+    auto stream = std::make_shared<MemoryStream>(fileToLoad->Buffer.get(), fileToLoad->BufferSize);
```
dcvz: Looks like this method has had regressions
```suggestion
    auto stream = std::make_shared<MemoryStream>(fileToLoad->Buffer.data(), fileToLoad->Buffer.size());
```

src/graphic/Fast3D/gfx_gx2.cpp
GaryOderNichts: Not aware of any ways to query this. Could be turned into a define in the Wii U toolchain.
I think it's fine to hardcode it for now.

src/resource/ResourceLoader.cpp
dcvz: Sorted out by me earlier

src/resource/Archive.cpp
dcvz: Sorted out by me earlier

src/core/bridge/resourcebridge.cpp
dcvz: Removed!

src/core/bridge/resourcebridge.cpp
dcvz: Removed

src/resource/Archive.cpp
dcvz: I added this while fixing compilation

src/resource/Archive.cpp
dcvz: Fixed

src/resource/Archive.cpp
dcvz: Fixed

src/resource/Archive.cpp
dcvz: Fixed

src/resource/Archive.cpp
dcvz: Fixed

src/resource/Archive.cpp
dcvz: Fixed

src/resource/Archive.cpp
dcvz: Fixed

src/core/bridge/resourcebridge.cpp
dcvz: Yup, ok this has been removed now

src/graphic/Fast3D/gfx_direct3d11.cpp
Kenix3: `D3D11_REQ_TEXTURE2D_U_OR_V_DIMENSION ` would be better than a hardcoded value.
https://learn.microsoft.com/en-us/windows/win32/direct3d11/overviews-direct3d-11-resources-limits

It looks like this might be the minimum that hardware needs to support to be D3D11 compatibility.

src/graphic/Fast3D/gfx_gx2.cpp
Kenix3: If there's no define in the WII U toolchain this is fine for now.

Kenix3, line 140:
```suggestion
    // TODO: This should be a define from the Wii U toolchain, but it doesn't appear to exist right now.
    return 8192;
```

src/resource/factory/TextureFactory.cpp
```diff
@@ -8,10 +8,13 @@ std::shared_ptr<Resource> TextureFactory::ReadResource(uint32_t version, std::sh
     auto resource = std::make_shared<Texture>();
     std::shared_ptr<ResourceVersionFactory> factory = nullptr;
 
-    switch (version) {
-        case 0:
+    switch ((Version)version) {
+        case Version::Deckard:
             factory = std::make_shared<TextureFactoryV0>();
             break;
+        case Version::Spock:
```
Kenix3: We should talk about whether this should be `Spock` or `Roy`.
In other words, are we going by this being the second version of _this_ resource, or it being introduced as part of SoH `Spock`. I'd probably prefer we go with `Roy` here.

src/resource/type/Texture.h
```diff
@@ -24,6 +26,9 @@ class Texture : public Resource {
 
     TextureType Type;
     uint16_t Width, Height;
+    uint32_t Flags = 0;
+    float HByteScale = 1;
```
Kenix3:
```suggestion
    float HByteScale = 1.0f;
```

src/resource/type/Texture.h
```diff
@@ -24,6 +26,9 @@ class Texture : public Resource {
...
+    float VPixelScale = 1;
```
Kenix3:
```suggestion
    float VPixelScale = 1.0f;
```

src/resource/factory/TextureFactory.cpp
dcvz: Roy is fine. Basically the second version of the texture factory

PR #139, opened 2023-03-01
[Renderer] Fix graphics switching on macOS

dcvz:
Fixes a bug caused by the union of graphics option, now correctly splits the backends

### DIFF
--- a/src/graphic/Fast3D/gfx_sdl2.cpp
+++ b/src/graphic/Fast3D/gfx_sdl2.cpp
@@ -332,6 +332,7 @@ static void gfx_sdl_init(const char* game_name, const char* gfx_api_name, bool s
         SDL_GL_SetSwapInterval(1);
 
         window_impl.Opengl = { wnd, ctx };
+        window_impl.backend = SohImGui::Backend::SDL_OPENGL;
     } else {
         renderer = SDL_CreateRenderer(wnd, -1, SDL_RENDERER_ACCELERATED | SDL_RENDERER_PRESENTVSYNC);
         if (renderer == NULL) {
@@ -341,9 +342,9 @@ static void gfx_sdl_init(const char* game_name, const char* gfx_api_name, bool s
 
         SDL_GetRendererOutputSize(renderer, &window_width, &window_height);
         window_impl.Metal = { wnd, renderer };
+        window_impl.backend = SohImGui::Backend::SDL_METAL;
     }
 
-    window_impl.backend = SohImGui::Backend::SDL;
     SohImGui::Init(window_impl);
 
     for (size_t i = 0; i < sizeof(windows_scancode_table) / sizeof(SDL_Scancode); i++) {
--- a/src/menu/ImGuiImpl.cpp
+++ b/src/menu/ImGuiImpl.cpp
@@ -35,7 +35,7 @@
 #include "graphic/Fast3D/gfx_gx2.h"
 #endif
 
-#if __APPLE__
+#ifdef __APPLE__
 #include <SDL_hints.h>
 #include <SDL_video.h>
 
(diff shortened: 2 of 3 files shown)


src/graphic/Fast3D/gfx_sdl2.cpp
```diff
@@ -341,9 +342,9 @@ static void gfx_sdl_init(const char* game_name, const char* gfx_api_name, bool s
 
         SDL_GetRendererOutputSize(renderer, &window_width, &window_height);
         window_impl.Metal = { wnd, renderer };
+        window_impl.backend = SohImGui::Backend::SDL_METAL;
     }
 
-    window_impl.backend = SohImGui::Backend::SDL;
```
briaguya0, line 346: this was concerning when i first saw it, but after looking into it, unless we're on apple, we will always hit
```cpp
window_impl.backend = SohImGui::Backend::SDL_OPENGL;
```
on L335 on non-mac systems using sdl because L314 has
```cpp
if (use_opengl) {
```
and that is set up on L265
```cpp
#if defined(__APPLE__)
    bool use_opengl = strcmp(gfx_api_name, "OpenGL") == 0;
#else
    bool use_opengl = true;
#endif
```

src/menu/ImGuiImpl.cpp
```diff
@@ -202,17 +202,19 @@ void ImGuiWMInit() {
...
-            }
-
             SDL_SetHint(SDL_HINT_JOYSTICK_ALLOW_BACKGROUND_EVENTS, "1");
             ImGui_ImplSDL2_InitForOpenGL(static_cast<SDL_Window*>(impl.Opengl.Window), impl.Opengl.Context);
             break;
 #endif
+#if __APPLE__
+        case Backend::SDL_METAL:
+            SDL_SetHint(SDL_HINT_TOUCH_MOUSE_EVENTS, "1");
+            SDL_SetHint(SDL_HINT_JOYSTICK_ALLOW_BACKGROUND_EVENTS, "1");
```
briaguya0, line 214: it'd be nice to not have these in 2 of the cases, but trying to not copy/paste would probably make this switch/case even messier

src/menu/ImGuiImpl.cpp
```diff
@@ -287,7 +291,8 @@ void ImGuiWMNewFrame() {
         case Backend::GX2:
             break;
 #else
-        case Backend::SDL:
+        case Backend::SDL_OPENGL:
+        case Backend::SDL_METAL:
```
briaguya0, line 295: we're `#ifdef`ing GX2, shoudl we `#ifdef` this one too?

src/menu/ImGuiImpl.cpp
```diff
@@ -383,7 +387,8 @@ bool SupportsViewports() {
     switch (impl.backend) {
         case Backend::DX11:
             return true;
-        case Backend::SDL:
+        case Backend::SDL_OPENGL:
+        case Backend::SDL_METAL:
```
briaguya0, line 391: it seems we have some inconsistencies around when we ifdef and when we don't, it seems like we should ifdef away DX11 when not on windows but that's not happening here, so maybe not ifdefing away metal when not on mac is fine too?

briaguya0 (commented):
Overall this seems to clean up logic quite a bit, left a few comments with questions around what should/shouldn't be `#ifdef`'d out. I think it'd be good to get a clear answer to those before merging this

src/menu/ImGuiImpl.cpp
dcvz: I'm ifdeffing when we use something specific that would not be available on non Apple machines and therefore wouldn't compile or would not run. Here we're just returning values so there's no risk. That's what the other backends are doing as well.

src/menu/ImGuiImpl.cpp
dcvz: Here I don't ifdef, because `ImGui_ImplSDL2_NewFrame` will be available no matter the backend and doesn't really do anything apple specific.

src/menu/ImGuiImpl.cpp
dcvz: Yeah, couldn't think of a better way here given they both call different methods :/

src/graphic/Fast3D/gfx_sdl2.cpp
dcvz: Correct!

briaguya0 (approved):
The logic behind what is and isn't `#ifdef`'d seems sound, this LGTM

Kenix3 (approved):

PR #140, opened 2023-03-01
Force vsync interval to 1

aMannus:
I'd like to preface this that I'm not very familiar with how VSync works exactly, so most of this information is based from what I gathered from others.

Currently, variable refresh rate (gsync, freesync) monitors drop a large amount of frames. What I gathered is that this has to do with the interval rapidly changing constantly. The "fix" until now was to tell those people to force VSync on for SoH. I've given some people with VRR displays a test build where this PR's change was applied, and they all mentioned they saw a large increase in FPS.

As far as I understood it, this does force on VSync. I don't know why this line was commented out and it was added months ago.

I'm mostly making this PR just so it's out there and people that know more about this might be able to take a look. If it doesn't hurt anything though, I'd say this is much more preferable to the situation where we need to tell people to dive into GPU drivers, even if it forces on VSync.


briaguya0 (approved):
I think it makes sense to get this in. In the longer term we probably want to look into better VRR support ([something about a tearing flag](https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/variable-refresh-rate-displays)?), but I agree that for now it's better to have this fix in than to tell everyone with a VRR display to manually force vsync.

briaguya0:
context from digging into fast3d https://github.com/Emill/n64-fast3d-engine/: this has been commented out forever https://github.com/Emill/n64-fast3d-engine/blob/a99492ddb3dd7e068a6f788558b63a17cd561374/gfx_direct3d12.cpp#L1295

it looks like it's getting set in

```cpp
static bool gfx_dxgi_start_frame(void) {
    // [...]
    if (dxgi.frame_stats.size() >= 2) {
        // [...] lots of logic that sets vsyncs_to_wait
        dxgi.length_in_vsync_frames = vsyncs_to_wait;
    } else {
        dxgi.length_in_vsync_frames = 1;
        dxgi.use_timer = true;
    }
}
```

https://github.com/Kenix3/libultraship/blob/b7d11ccfbd0af608ee029d33684d19cb35ee227e/src/graphic/Fast3D/gfx_dxgi.cpp#L461-L560

and from the quick look i took at fast3d in sm64ex it seems to be the same there

briaguya0:
more context from `Spodi` on discord: https://discord.com/channels/808039310850130000/1019824570938699798/1064986464976371922

> I don't think it goes more proper than that. A present interval of 0 (V-sync off) or 1 (V-sync on) is fine with g-sync (and seemingly all VRR techs), but V-Sync on is recommended (there might be people that complain about V-sync on, so be prepared for that...). The only thing that does not work is a present interval that is >1 (also called fractional V-Sync. 2 = 1/2 V-Sync for example.), cause that is completely disabling VRR. Rapidly switching between them is even worse, as that will enable and disable VRR in a rapid manner and refresh rate and FPS will jump around all over the place.
> TLDR: Present interval >1 = no VRR, rapidly switching present interval = stutter

briaguya0:
had chatGPT explain `gfx_dxgi_start_frame` (response in this gist https://gist.github.com/briaguya-ai/a17675e57e5301ac1c04949acdc2af78)

After looking through that I'm thinking there's probably something wonky going on with `SyncRefreshCount` when using VRR

briaguya0:
looks like sm64-ex-coop just commented out all the start frame fractional vsync logic https://github.com/djoslin0/sm64ex-coop/blob/8746a503086793c87860daadfaeaaf0a31b2d6cf/src/pc/gfx/gfx_dxgi.cpp#L409-L522

```cpp
    // HACK: all of this is too confusing to bother with right now
    // [...]
    dxgi.length_in_vsync_frames = configWindow.vsync;
    dxgi.dropped_frame = false;

    return true;
```

where `configWindow.vsync` is just a `bool` https://github.com/djoslin0/sm64ex-coop/blob/8746a503086793c87860daadfaeaaf0a31b2d6cf/src/pc/configfile.h#L20

briaguya0 (approved):
LGTM! It's doing exactly what https://github.com/djoslin0/sm64ex-coop/blob/8746a503086793c87860daadfaeaaf0a31b2d6cf/src/pc/gfx/gfx_dxgi.cpp#L409-L522 does for `sm64ex-coop`.

I'm also a fan of just removing the code instead of commenting it out. We will have commit history and this PR to reference in the future if we need it.

aMannus:
Gave a build to Spodi to test it out and apparently there's issues with the game wrongfully speeding up. Investigating now.

aMannus:
I've reverted the code to the same way I was doing it originally. Confirmed to properly force VSync without any weird speedups. I'm not sure what's different between our implementation and `sm64ex-coop` but clearly there's something different on our side. 

That said, on the latest develop of SoH there's now problems with forcing on VSync when using GSync. I don't know why latest develop now has issues with this (6.1.2 works fine), so it might be a good idea to not merge this before we have an answer to that.

briaguya0:
what happens if you don't revert and instead use
```cpp
static bool gfx_dxgi_start_frame(void) {
    // Originally this contained code which was responsible for handling fractional VSync.
    // It was causing the interval to change rapidly when using VRR displays (GSync or FreeSync),
    // causing a big hit to performance.
    // Instead of using fractional VSync, set the interval to a specific number.
    // Can be potentially set with a config option in the future to disable/enable VSync entirely.
    // 0 = VSync off.
    // 1 = VSync on.
    dxgi.length_in_vsync_frames = 1;
    dxgi.dropped_frame = false;
    dxgi.use_timer = false;
    return true;
}
```

https://github.com/djoslin0/sm64ex-coop/blob/8746a503086793c87860daadfaeaaf0a31b2d6cf/src/pc/gfx/gfx_dxgi.cpp#L530-L537 has
```cpp
static void gfx_dxgi_swap_buffers_begin(void) {
    ThrowIfFailed(dxgi.swap_chain->Present(dxgi.length_in_vsync_frames, 0));
    UINT this_present_id;
    if (dxgi.swap_chain->GetLastPresentCount(&this_present_id) == S_OK) {
        dxgi.pending_frame_stats.insert(std::make_pair(this_present_id, dxgi.length_in_vsync_frames));
    }
    dxgi.dropped_frame = false;
}
```
which lines up with just skipping over the `if (dxgi.use_timer)` logic

briaguya0:
so it seems that the change i suggested in https://github.com/Kenix3/libultraship/pull/140#issuecomment-1509450029 leads to the game running way too fast (builds used for testing here: https://github.com/HarbourMasters/Shipwright/pull/2710)

kicked off some builds with `use_timer` set to false, will comment again with test results from that

Kenix3:
Superseded by https://github.com/Kenix3/libultraship/pull/171

PR #141, opened 2023-03-03
Custom model (and XML resources) support

NEstelami:


src/binarytools/FileHelper.h
```diff
@@ -0,0 +1,83 @@
...
+        int32_t fileSize = (int32_t)file.tellg();
+        file.seekg(0);
+        char* data = nullptr;
+        std::vector<uint8_t> result;
+
+        try {
+            data = new char[fileSize];
+            file.read(data, fileSize);
+            result = std::vector<uint8_t>(data, data + fileSize);
+        } catch (const std::exception& e) { delete[] data; }
```
Kenix3:
```suggestion
        } catch (const std::exception& e) { delete[] data; throw e; }
```

src/binarytools/FileHelper.h
```diff
@@ -0,0 +1,83 @@
...
+        delete[] data;
+
+        return result;
+    };
+
+    static std::string ReadAllText(const fs::path& filePath) {
+        std::ifstream file(filePath, std::ios::in | std::ios::binary | std::ios::ate);
+        int32_t fileSize = (int32_t)file.tellg();
+        file.seekg(0);
+        char* data = new char[fileSize + 1];
```
Kenix3, line 45: Same as above here.

include/libultraship/libultra/gbi.h
```diff
@@ -164,7 +164,13 @@
 #define G_SETFB 0x21
 #define G_RESETFB 0x22
 #define G_SETTIMG_FB 0x23
-#define G_DL_OTR 0x31
+#define G_VTX_OTR2 0x24
```
Kenix3: Better name than just 2?

include/libultraship/libultra/gbi.h
```diff
@@ -164,7 +164,13 @@
...
+#define G_SETTIMG_OTR2 0x25
```
Kenix3, line 168: Better name than just 2?

include/libultraship/libultra/gbi.h
```diff
@@ -164,7 +164,13 @@
...
+#define G_TRI1_OTR 0x26
+#define G_DL_OTR_FILEPATH 0x27
+#define G_PUSHCD 0x28
+#define G_MTX_OTR2 0x29
```
Kenix3, line 172: Better name than just 2?

include/libultraship/libultra/gbi.h
```diff
@@ -1726,6 +1732,13 @@ typedef union {
     })
 #define gsSPVertex(v, n, v0) \
     { (_SHIFTL(G_VTX, 24, 8) | _SHIFTL((n), 12, 8) | _SHIFTL((v0) + (n), 1, 7)), (uintptr_t)(v) }
+
+#define gsSPVertexOTR2_P1(filePathPtr) \
```
Kenix3, line 1736: Better name than just 2?

include/libultraship/libultra/gbi.h
```diff
@@ -1726,6 +1732,13 @@ typedef union {
...
+    { (_SHIFTL(G_VTX_OTR2, 24, 8), (uintptr_t)(filePathPtr)) }
+
+#define gsSPVertexOTR2_P2(vtxCnt, vtxBufOffset, vtxDataOffset) \
```
Kenix3, line 1739: Better name than just 2?

src/binarytools/BinaryReader.cpp
```diff
@@ -192,6 +192,21 @@ std::string Ship::BinaryReader::ReadString() {
     return res;
 }
 
+std::string Ship::BinaryReader::ReadCString() {
+    std::string res;
+
+    unsigned char c = 0;
+    do {
+        if (mStream->GetBaseAddress() >= mStream->GetLength())
```
Kenix3: Why not put this in the while? I'm OK with either, just curious on the preference.

NEstelami: I usually don't use do-while loops but in this case I felt it made sense. It needs to read at least one character before it knows whether or not to stop (checking for the terminator), a do-while guarantees that. A pure-while would have to be written a bit differently.

PR #144, opened 2023-03-08
Removed small redundant validation on gfx_s2dex_bg_copy and added ifs to avoid null textures

KiritoDv:


src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -2243,18 +2243,16 @@ static void gfx_s2dex_bg_copy(uObjBg* bg) {
...
-            texFlags = tex->Flags;
-            rawTexMetadata.width = tex->Width;
-            rawTexMetadata.height = tex->Height;
-            rawTexMetadata.h_byte_scale = tex->HByteScale;
-            rawTexMetadata.v_pixel_scale = tex->VPixelScale;
-            rawTexMetadata.type = tex->Type;
-            rawTexMetadata.name = std::string((char*)data);
-            data = (uintptr_t) reinterpret_cast<char*>(tex->ImageData);
-        }
+    if ((bool) gfx_check_image_signature((char*)data)) {
```
briaguya0:
```suggestion
    if ((bool)gfx_check_image_signature((char*)data)) {
```
make clang format happy

Kenix3 (approved):

Kenix3:
Looks like we need to run clang-format on this again. After that it appears to be ready.

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -2243,18 +2243,16 @@ static void gfx_s2dex_bg_copy(uObjBg* bg) {
     uint32_t texFlags = 0;
     RawTexMetadata rawTexMetadata = {};
 
-    if ((data & 1) != 1) {
```
Kenix3, line 2246: Why are you getting rid of the lowest order bit check?

KiritoDv: Because the gfx_check_image_signature already does this

PR #152, opened 2023-03-31
gfx_opengl.cpp: OpenGL / GLSL optimizations

tdc22:
With this PR I propose the following changes:

1. **reduction of `glUniform`-calls**: previously, once per frame every `ShaderProgram` received two `glUniform` calls. One to update the `frame_count` and one for `noise_scale`. For the OpenGL-renderer, both variables were only used for the GLSL function `random(vec3)` which isn't actually used in most shaders within SOH. To avoid this I introduced a variable `used_noise` into `ShaderProgram` which determines whether the `glUniform` calls are made or not. (Note: I apparently wasn't the first one with this idea, as the renderer `gfx_gx2.cpp` uses a similar approach)
2. **removal of unused GLSL functions**: both `random(vec3)` and `hookTexture2D(sampler2D, vec2, vec2)` were only used in specific situations and don't need to be included in every GLSL-shader. I simply introduced appropriate conditions to include them in the shader source.
3. **limitation of frame_count**: the renderers for dx11, dx12 and metal do this, too. I'm not actually sure if this is needed within SOH as this change seems to originate from the original SM64 renderer. Since I'm uncertain where this shader noise is used in SOH I couldn't test and compare this change but I thought it might make sense for the renderers to behave as similarily to each other as possible.

### DIFF
--- a/src/graphic/Fast3D/gfx_opengl.cpp
+++ b/src/graphic/Fast3D/gfx_opengl.cpp
@@ -58,6 +58,7 @@ struct ShaderProgram {
     GLint attrib_locations[16];
     uint8_t attrib_sizes[16];
     uint8_t num_attribs;
+    bool used_noise;
     GLint frame_count_location;
     GLint noise_scale_location;
 };
@@ -115,8 +116,10 @@ static void gfx_opengl_vertex_array_set_attribs(struct ShaderProgram* prg) {
 }
 
 static void gfx_opengl_set_uniforms(struct ShaderProgram* prg) {
-    glUniform1i(prg->frame_count_location, frame_count);
-    glUniform1f(prg->noise_scale_location, current_noise_scale);
+    if (prg->used_noise) {
+        glUniform1i(prg->frame_count_location, frame_count);
+        glUniform1f(prg->noise_scale_location, current_noise_scale);
+    }
 }
 
 static void gfx_opengl_unload_shader(struct ShaderProgram* old_prg) {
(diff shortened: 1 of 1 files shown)


tdc22:
There's still a possible bug if the GLSL `random` function is inserted at the end: it has to be declared before it is used. I'll fix this.

tdc22:
The case that GLSL shaders are generated that use the `random` function before declaring it now should not be possible anymore. The fix I chose is a bit inelegant as the 1-2 iteration for-loop that iterates over the `cc_features` now exists twice but otherwise I would have to do more complex manipulations on the `fs_buf` than just `append_str` and `append_line`. Instead I basically recreated the logic of `append_formula` to determine if noise is used within this shader. While this might add a bit of complexity for shader initialization I think that overall this optimization is worth it since it _saves up to two glUniform-calls per shader per frame_.
I would appreciate a review / feedback on this. :)

Kenix3 (commented):
For my curiosity, did you measure performance before and after? If so, what kind of numbers did you see?

Kenix3:
I decided to do some benchmarking myself now that a lot of the needed rapid development has finished. The game crashes when using OpenGL on Windows.

image

Kenix3:
Hi, is there any interest still in maintaining this PR? Thankyou!

tdc22:
Sorry, right now I'm unavailable. I didn't get to finish this before my thesis started. Thanks for your replies and your patience.

PR #153, opened 2023-04-02
Custom model support fixes

NEstelami:


briaguya0:
tidy/format fixes here https://github.com/NEstelami/libultraship/pull/1

Kenix3 (approved):
Not a fan of the names on the importer functions, but it looks good.

PR #161, opened 2023-04-12
add methods to get macos app bundle path

Archez:
This adds methods to get the app bundle path on mac which will return the corresponding locations in the following scenarios:
* local development: `build-cmake/soh` (where the soh-macos and otr files are placed)
* release app: `xxx/soh.app/Contents/Resources`

Then files can be bundled into the resources folder within the app to be read later.


Kenix3:
Superseded by #162

PR #163, opened 2023-04-12
DirectX: Adjust Windows Scancodes

stratomaster64:
Currently, when a key is pressed on the keyboard while using DirectX as a GFX backend, we receive a keycode (index relative to physical location) and scancode (index relative to actual value, after some bitmagic), then send the scancode to ImGui to process input events. However, when keys other than function keys are pressed, there is a mismatch between the Windows scancode and the SDL scancode (which applications such as SoH's `OTRGlobals.cpp` use).
To resolve this, a lookup table similar to the one in `gfx_sdl2.cpp` and [`scancodes_windows.h`](https://github.com/spurious/SDL-mirror/blob/master/src/events/scancodes_windows.h) was added for DirectX so we can use that to get the correct SDL scancode.
This does not interfere with key combinations such as `Alt+Enter`.
As this is my first PR to LUS, please let me know how I can improve!


src/graphic/Fast3D/gfx_dxgi.h
```diff
@@ -20,4 +20,143 @@ void ThrowIfFailed(HRESULT res, HWND h_wnd, const char* message);
...
+#include "SDL_scancode.h"
+const SDL_Scancode windows_scancode_table[] = {
+    SDL_SCANCODE_UNKNOWN,
+    SDL_SCANCODE_ESCAPE,
+    SDL_SCANCODE_1,
+    SDL_SCANCODE_2,
+    SDL_SCANCODE_3,
+    SDL_SCANCODE_4,
+    SDL_SCANCODE_5,
+    SDL_SCANCODE_6, /* 0 */
```
Kenix3, line 32: What are the commented numbers?

Kenix3 (approved):

src/graphic/Fast3D/gfx_dxgi.h
stratomaster64: The commented numbers are from copying over `gfx_sdl2`'s table, and were probably there to assist with enumerating. a similar system is in `scancodes_windows.h`.

PR #164, opened 2023-04-13
Add shader support for some CPU-modified textures in OpenGL and D3D11

Rozelette:
This introduces shader features in order to support some CPU-modified textures in OOT.

In OOT, several effects are implemented by gradually clearing textures. Additionally, there is an effect that gradually replaces one texture with another. I have implemented this by adding a mechanism to indicate that a texture is masked by another, and when it is, it is replaced by another texture. When another texture is not specified, it is replaced by transparent black, which is what OOT uses in all cases.

I considered two options for this. Firstly, adding a custom DL command. Secondly, by specifying what texture would be replaced by file path so that the information could be referenced when loaded. I opted for the latter because it was the more general-purpose solution. I was not sure of the scope patching all the DLs in OOT would be if I added a custom command for this, or what it would be in other games. Having a hook for the game to say "mask this texture," with no worry about any other data, seemed ideal. This approach does not exclude adding a command, if desired.

To facilitate this, I needed additional bits for the color combiner settings. Accordingly, I have done some work to expand the bits for the keys for the them. For speed, instead of splitting them all into bools, they are 2 `uint64_t`s, but when more space is needed, a third, fourth, etc. can easily be added. For now, there is plenty of free space for additional flags.

This PR is limited to DX11 and OpenGL, as I do not have a way of testing other APIs. There is a risk of crashing with other APIs until they are changed to support them, as with these changes Fast3D assumes that they support 6 textures, up from 2.

### DIFF
--- a/src/core/bridge/resourcebridge.cpp
+++ b/src/core/bridge/resourcebridge.cpp
@@ -193,24 +193,4 @@ void UnloadResourceByCrc(uint64_t crc) {
 void UnloadResourceDirectory(const char* name) {
     Ship::Window::GetInstance()->GetResourceManager()->UnloadDirectory(name);
 }
-
-void WriteTextureDataInt16ByName(const char* name, size_t index, int16_t valueToWrite, bool now) {
-    const auto res = static_pointer_cast<Ship::Texture>(LoadResource(name, now));
-
-    if (res != nullptr) {
-        if ((index * sizeof(int16_t)) < res->ImageDataSize) {
-            ((int16_t*)res->ImageData)[index] = valueToWrite;
-        }
-    }
-}
-
-void WriteTextureDataInt16ByCrc(uint64_t crc, size_t index, int16_t valueToWrite, bool now) {
-    const auto res = static_pointer_cast<Ship::Texture>(LoadResource(crc, now));
-
-    if (res != nullptr) {
-        if ((index * sizeof(int16_t)) < res->ImageDataSize) {
-            ((int16_t*)res->ImageData)[index] = valueToWrite;
-        }
-    }
-}
 }
--- a/src/graphic/Fast3D/gfx_cc.h
+++ b/src/graphic/Fast3D/gfx_cc.h
@@ -3,6 +3,7 @@
 
 #include <stdint.h>
 #include <stdbool.h>
+#include <compare>
 
 /*enum {
     CC_0,
@@ -45,7 +46,22 @@ enum {
 #define SHADER_OPT_TEXEL0_CLAMP_T (1 << 9)
 #define SHADER_OPT_TEXEL1_CLAMP_S (1 << 10)
 #define SHADER_OPT_TEXEL1_CLAMP_T (1 << 11)
-#define CC_SHADER_OPT_POS 56
+#define SHADER_OPT_TEXEL0_MASK (1 << 12)
+#define SHADER_OPT_TEXEL1_MASK (1 << 13)
+#define SHADER_OPT_TEXEL0_BLEND (1 << 14)
+#define SHADER_OPT_TEXEL1_BLEND (1 << 15)
+
+struct ColorCombinerKey {
+    uint64_t combine_mode;
+    uint64_t options;
+
+    auto operator<=>(const ColorCombinerKey&) const = default;
+};
+
+#define SHADER_MAX_TEXURES 6
+#define SHADER_FIRST_TEXTURE 0
+#define SHADER_FIRST_MASK_TEXTURE 2
+#define SHADER_FIRST_REPLACEMENT_TEXTURE 4
 
 struct CCFeatures {
     uint8_t c[2][2][4];
@@ -58,6 +74,8 @@ struct CCFeatures {
     bool opt_invisible;
     bool opt_grayscale;
     bool used_textures[2];
+    bool used_masks[2];
+    bool used_blend[2];
     bool clamp[2][2];
     int num_inputs;
     bool do_single[2][2];
--- a/src/graphic/Fast3D/gfx_direct3d11.cpp
+++ b/src/graphic/Fast3D/gfx_direct3d11.cpp
@@ -51,7 +51,7 @@ struct PerDrawCB {
         uint32_t height;
         uint32_t linear_filtering;
         uint32_t padding;
-    } textures[2];
+    } textures[SHADER_MAX_TEXURES];
 };
 
 struct Coord {
@@ -86,7 +86,7 @@ struct ShaderProgramD3D11 {
     uint32_t shader_id1;
     uint8_t num_inputs;
     uint8_t num_floats;
-    bool used_textures[2];
+    bool used_textures[SHADER_MAX_TEXURES];
 };
 
 static struct {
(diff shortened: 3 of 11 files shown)


src/core/bridge/resourcebridge.cpp
```diff
@@ -193,24 +193,4 @@ void UnloadResourceByCrc(uint64_t crc) {
 void UnloadResourceDirectory(const char* name) {
     Ship::Window::GetInstance()->GetResourceManager()->UnloadDirectory(name);
 }
-
-void WriteTextureDataInt16ByName(const char* name, size_t index, int16_t valueToWrite, bool now) {
```
Rozelette, line 197: Removed as writing directly textures should explicitly not be supported.

src/graphic/Fast3D/gfx_cc.h
```diff
@@ -45,7 +46,22 @@ enum {
...
 #define SHADER_OPT_TEXEL1_CLAMP_T (1 << 11)
-#define CC_SHADER_OPT_POS 56
+#define SHADER_OPT_TEXEL0_MASK (1 << 12)
+#define SHADER_OPT_TEXEL1_MASK (1 << 13)
+#define SHADER_OPT_TEXEL0_BLEND (1 << 14)
+#define SHADER_OPT_TEXEL1_BLEND (1 << 15)
+
+struct ColorCombinerKey {
+    uint64_t combine_mode;
+    uint64_t options;
```
Rozelette, line 56: If more options are desired, `options_2` or similar can be added.

src/graphic/Fast3D/gfx_opengl.cpp
```diff
@@ -447,26 +459,44 @@ static struct ShaderProgram* gfx_opengl_create_and_load_new_shader(uint64_t shad
...
+                if (cc_features.used_blend[i]) {
+                    fs_len += sprintf(fs_buf + fs_len,
+                                      "vec4 blendVal%d = hookTexture2D(uTexBlend%d, vTexCoordAdj%d, texSize%d);\n", i,
+                                      i, i, i);
+                } else {
+                    fs_len += sprintf(fs_buf + fs_len, "vec4 blendVal%d = vec4(0, 0, 0, 0);\n", i);
+                }
+
+                fs_len += sprintf(fs_buf + fs_len, "texVal%d = mix(texVal%d, blendVal%d, maskVal%d.a);\n", i, i, i, i);
+            }
```
Rozelette, line 499: This block of code had been reworked to more closely match DirectX, in that it calculates texture coordinates based of wrap/repeat settings, then has a common texture sampling point.

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -130,6 +130,7 @@ struct RawTexMetadata {
     uint16_t width, height;
     float h_byte_scale = 1, v_pixel_scale = 1;
     std::shared_ptr<Ship::Texture> resource;
+    std::string name;
```
Rozelette: This reintroduces the removed texture name field. I reintroduced this because the replacement, `std::shared_ptr<Ship::Texture> resource`, does not have good way of testing equivalence. Games will be specifying textures to be replaced via file paths, and there is no good way of going from paths -> resources is the first place, so back to paths it is.

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -944,6 +968,63 @@ static void import_texture(int i, int tile) {
...
+            tex_upload_buffer[4 * texIndex + 1] = 0;
+            tex_upload_buffer[4 * texIndex + 2] = 0;
+            tex_upload_buffer[4 * texIndex + 3] = 0xFF;
+        } else {
+            tex_upload_buffer[4 * texIndex + 0] = 0;
+            tex_upload_buffer[4 * texIndex + 1] = 0;
+            tex_upload_buffer[4 * texIndex + 2] = 0;
+            tex_upload_buffer[4 * texIndex + 3] = 0;
+        }
+    }
```
Rozelette, line 1037: This is something I want to fix, but currently all textures are treated as RGBA32. It was easiest to just work with this for now.

src/graphic/Fast3D/gfx_pc.cpp
Kenix3: Is name the same as path? If it's the same then the resource has the path in it's InitData.

src/core/bridge/resourcebridge.cpp
Kenix3: Sounds good to me.

src/graphic/Fast3D/gfx_direct3d11.cpp
```diff
@@ -51,7 +51,7 @@ struct PerDrawCB {
         uint32_t height;
         uint32_t linear_filtering;
         uint32_t padding;
-    } textures[2];
+    } textures[SHADER_MAX_TEXURES];
```
Kenix3, line 54: Glad that this is now defined elsewhere for the texture count.

src/graphic/Fast3D/gfx_pc.cpp
Kenix3: All cpu modified textures, or all textures period?

src/graphic/Fast3D/gfx_pc.cpp
Rozelette: Using that did work. Strangely there were some textures that did not have an associated resource, but a null check fixed that.

src/graphic/Fast3D/gfx_pc.cpp
Rozelette: All textures period. Every texture is converted to RGBA32 before being given to the APIs, and the APIs can only handle RGBA32. Even HD textures are being converted. I think this is done because not all textures map 1-1 between N64 and modern APIs, but it still could be optimized. Especially in this case, where the mask texture has literally 1 bit on information per texel but becomes 32 bits wide.

This would be an overhaul well outside the scope of the PR.

src/graphic/Fast3D/gfx_pc.cpp
Kenix3: Sounds good to me.

src/graphic/Fast3D/gfx_pc.cpp
Kenix3: Yeah that sounds good to me, especially given that it was existing behaviour. I do know that the easiest textures to upload to the backends are RGB. Definitely out of scope.

src/graphic/Fast3D/gfx_pc.cpp
Kenix3: Any insight into the cases where the Texture did not have a Resource? I assume this isn't the Texture subclass of Resource.

Another way to ask the question: Any cases that a texture may not have a Resource that is not set currently, but would have been set when we were using the name?

Kenix3: Roze on Discord:
```
I'm not sure, it also did not have a name. It occurred during boot before anything was drawn. It may have not even been a texture, just a load_block before anything was set up? I didn't look into it too much.
```

PR #165, opened 2023-04-13
LUS Scancodes

stratomaster64:
Cleans up #163 by adding LUS scancodes, which mimic Windows scancodes. When checking for key presses in application code, we will refer to LUS's `KbScancode` rather than SDL's `SDL_Scancode`.


briaguya0 (approved):
:shipit:

Kenix3 (approved):

PR #166, opened 2023-04-14
Declutter logs

stratomaster64:
Changes the default for release builds to `debug` rather than `trace` through `spdlog`. This should drastically reduce the size of files and allow for easier client-side debugging.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -418,7 +418,8 @@ if (NOT CMAKE_SYSTEM_NAME STREQUAL "CafeOS")
         ENABLE_OPENGL
         $<$<CONFIG:Debug>:_DEBUG>
         $<$<NOT:$<CONFIG:Debug>>:NDEBUG>
-        SPDLOG_ACTIVE_LEVEL=1
+        $<$<CONFIG:Debug>:SPDLOG_ACTIVE_LEVEL=0>
+        $<$<NOT:$<CONFIG:Debug>>:SPDLOG_ACTIVE_LEVEL=1>
     )
 else ()
     target_compile_definitions(libultraship PRIVATE
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -418,7 +418,7 @@ if (NOT CMAKE_SYSTEM_NAME STREQUAL "CafeOS")
         ENABLE_OPENGL
         $<$<CONFIG:Debug>:_DEBUG>
         $<$<NOT:$<CONFIG:Debug>>:NDEBUG>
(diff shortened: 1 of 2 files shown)


briaguya0 (changes requested):
this will make it so we don't have trace logs when building in debug either, we need conditional logic

src/CMakeLists.txt
```diff
@@ -418,7 +418,7 @@ if (NOT CMAKE_SYSTEM_NAME STREQUAL "CafeOS")
         ENABLE_OPENGL
         $<$<CONFIG:Debug>:_DEBUG>
         $<$<NOT:$<CONFIG:Debug>>:NDEBUG>
-        SPDLOG_ACTIVE_LEVEL=0
+        SPDLOG_ACTIVE_LEVEL=1
```
briaguya0:
```suggestion
        $<$<CONFIG:Debug>:SPDLOG_ACTIVE_LEVEL=0>
        $<$<NOT:$<CONFIG:Debug>>:SPDLOG_ACTIVE_LEVEL=1>
```

src/core/Window.cpp
```diff
@@ -454,13 +454,13 @@ void Window::InitializeLogging() {
 #ifndef __WIIU__
         auto logPath = GetPathRelativeToAppDirectory(("logs/" + GetName() + ".log").c_str());
         auto fileSink = std::make_shared<spdlog::sinks::rotating_file_sink_mt>(logPath, 1024 * 1024 * 10, 10);
-        fileSink->set_level(spdlog::level::trace);
+        fileSink->set_level(spdlog::level::debug);
```
Kenix3, line 460: Should ifdef here for `_DEBUG`. If it's `_DEBUG`, allow `trace` level, otherwise `debug` level.

src/core/Window.cpp
```diff
@@ -454,13 +454,13 @@ void Window::InitializeLogging() {
...
         sinks.push_back(fileSink);
 #endif
 
         mLogger = std::make_shared<spdlog::async_logger>(GetName(), sinks.begin(), sinks.end(), spdlog::thread_pool(),
                                                          spdlog::async_overflow_policy::block);
-        GetLogger()->set_level(spdlog::level::trace);
+        GetLogger()->set_level(spdlog::level::debug);
```
Kenix3, line 468: Should ifdef here for `_DEBUG`. If it's `_DEBUG`, allow `trace` level, otherwise `debug` level.

Kenix3 (changes requested):

Kenix3 (approved):

PR #167, opened 2023-04-14
Change log size to ~7MB

aMannus:
To accomodate for users uploading logs to Discord without Nitro.

image

If we want to reduce the number of logs to keep around, we can change that here too, but I wasn't sure if that's still something we want to do.


stratomaster64 (approved):

Kenix3:
Since Nitro sizes are now larger, and we log less, I think we don't need any changes here. Thankyou for the effort!

PR #175, opened 2023-04-22
fix out of bounds raw texture loading cases from authentic bugs

Archez:
There are some rare instances where the game will incorrectly load textures as the wrong size. In the event that a texture is loaded as a size greater than its actual size, the underlying `memcpy` would go out of bounds. On some platforms this would crash.

To account for this, I've adjusted the `memcpy` byte logic to safely bound to the raw resource image size if the expect copy bytes is larger. Then afterwards `memset` the remaining expected bytes to 0. The extra data being set as 0 results in the texture being black or transparent.

This crash/fix was noticed for Freezards and Iron Knuckles.

The downsides to this approach, is that the garbage data cannot be replicated from authentic behavior (presumably the overflow on hardware is just reading rom data).

### DIFF
--- a/src/graphic/Fast3D/gfx_pc.cpp
+++ b/src/graphic/Fast3D/gfx_pc.cpp
@@ -845,6 +845,7 @@ static void import_texture_raw(int tile) {
     uint16_t width = metadata->width;
     uint16_t height = metadata->height;
     Ship::TextureType type = metadata->type;
+    std::shared_ptr<Ship::Texture> resource = metadata->resource;
 
     // if texture type is CI4 or CI8 we need to apply tlut to it
     switch (type) {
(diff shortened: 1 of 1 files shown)


src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -881,10 +882,31 @@ static void import_texture_raw(int tile) {
         rdp.loaded_texture[rdp.texture_tile[tile].tmem_index].full_image_line_size_bytes;
     uint32_t line_size_bytes = rdp.loaded_texture[rdp.texture_tile[tile].tmem_index].line_size_bytes;
 
-    for (uint32_t i = 0, j = 0; i < num_loaded_bytes; i += line_size_bytes, j += full_image_line_size_bytes) {
-        memcpy(tex_upload_buffer + i, addr + j, line_size_bytes);
+    // Get the resource's true image size
+    uint32_t resource_image_size_bytes = resource->ImageDataSize;
```
Archez, line 886: I'm not happy using the resource pointer directly here. If we can compute the image size from the width/height info instead from the meta data, that would probably be better.

(width * height * 4) does that in the case of Freezards, but I don't want to assume.

Kenix3: GetPointerSize on the Resource is the physical size of data that the resource has allocated, from the perspective of what the game needs from the resource.

Kenix3 (commented):
One small comment, not sure anything needs to change.

PR #183, opened 2023-04-29
Add IsOnlyHexDigit

louist103:
This is for the #harbourmasters/shipwright#2292 PR.

### DIFF
--- a/extern/ZAPDUtils/Utils/StringHelper.cpp
+++ b/extern/ZAPDUtils/Utils/StringHelper.cpp
@@ -157,6 +157,12 @@ bool StringHelper::IsValidHex(std::string_view str)
 	return false;
 }
 
+
+bool StringHelper::IsValidHex(const std::string& str)
+{
+	return IsValidHex(std::string_view(str.c_str()));
+}
+
 bool StringHelper::IsValidOffset(std::string_view str)
 {
 	if (str.length() == 1)
@@ -167,10 +173,16 @@ bool StringHelper::IsValidOffset(std::string_view str)
 	return IsValidHex(str);
 }
 
-bool StringHelper::IsValidHex(const std::string& str)
+bool StringHelper::IsValidOffset(const std::string& str)
 {
-	return IsValidHex(std::string_view(str.c_str()));
+	if (str.length() == 1)
+	{
+		// 0 is a valid offset
+		return isdigit(str[0]);
+	}
+	return IsValidHex(str);
 }
+
 
 bool StringHelper::IEquals(const std::string& a, const std::string& b)
 {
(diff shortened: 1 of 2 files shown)


extern/ZAPDUtils/Utils/StringHelper.cpp
```diff
@@ -1,7 +1,9 @@
 #include "StringHelper.h"
 
+#if (_MSC_VER)
 #pragma optimize("2", on)
 #define _CRT_SECURE_NO_WARNINGS
+#endif
```
briaguya0, line 6: could you explain why this is being `#if`'d now (and why it wasn't before)?

extern/ZAPDUtils/Utils/StringHelper.cpp
```diff
@@ -140,6 +142,81 @@ bool StringHelper::HasOnlyDigits(const std::string& str)
...
+        {
+            return false;
+        }
+    }
+    else
+    {
+        return false; // ??
+    }
+}
+
```
briaguya0: feels like a lot of code when there are simpler solutions
https://stackoverflow.com/questions/8899069/how-to-find-if-a-given-string-conforms-to-hex-notation-eg-0x34ff-without-regex

i'm a fan of the "call [strtoul](http://linux.die.net/man/3/strtoul) and check for an error" method

briaguya0: if it's not possible to simplify this, then adding comments explaining why it's complex would be good. also, it seems like both methods are identical except one is taking in a view and the other is taking in a string, it'd be nice to not have copypasta

extern/ZAPDUtils/Utils/StringHelper.cpp
louist103: I can't explain why it wasn't before but I added it because that define is only needed for Windows. That pragma style from I have seen is only recognized on MSVC.

extern/ZAPDUtils/Utils/StringHelper.h
```diff
@@ -23,5 +24,8 @@ class StringHelper
 	static int64_t StrToL(const std::string& str, int32_t base = 10);
 	static std::string BoolStr(bool b);
 	static bool HasOnlyDigits(const std::string& str);
+	static bool IsValidHex(std::string_view str);
+	static bool IsValidHex(const std::string& str); 
+	static bool IsValidOffset(std::string_view str);
```
briaguya0, line 29: is this providing everything needed for https://github.com/HarbourMasters/Shipwright/pull/2292? it's not clear why we'd need both string and string view versions of `IsValidHex` but not of `IsValidOffset`, if anything i'd think it'd be the opposite

Kenix3 (approved):

PR #190, opened 2023-05-03
Removes all lingering references to SoH from LUS

Kenix3:


src/port/wiiu/WiiUImpl.cpp
```diff
@@ -59,9 +59,9 @@ void Init() {
     // make sure the required folders exist
     mkdir("/vol/external01/wiiu/", 0755);
     mkdir("/vol/external01/wiiu/apps/", 0755);
-    mkdir("/vol/external01/wiiu/apps/soh/", 0755);
+    mkdir(("/vol/external01/wiiu/apps/" + Window::GetInstance()->GetShortName()).c_str(), 0755);
```
GaryOderNichts, line 62: This might crash since the window instance might not be created yet at the time `Ship::WiiU::Init` is called (for example in SoH).
`Ship::WiiU::Init` provides the logging devoptab so this should probably not rely on the Window instance.

PR #202, opened 2023-05-07
Adds a new "Context" class that splits the functionality of the "Window" class.

Kenix3:


briaguya0 (approved):
approving from a code stand point, assuming everything builds/works in soh LGTM!

PR #247, opened 2023-05-16
Better input blocking

briaguya0:
this includes https://github.com/Kenix3/libultraship/pull/246 because i forgot to go back to main before making this branch

### DIFF
--- a/src/core/libultra/os.cpp
+++ b/src/core/libultra/os.cpp
@@ -37,7 +37,8 @@ int32_t osContStartReadData(OSMesgQueue* mesg) {
 void osContGetReadData(OSContPad* pad) {
     memset(pad, 0, sizeof(OSContPad) * __osMaxControllers);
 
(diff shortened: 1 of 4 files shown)


src/core/libultra/os.cpp
```diff
@@ -37,8 +37,8 @@ int32_t osContStartReadData(OSMesgQueue* mesg) {
 void osContGetReadData(OSContPad* pad) {
     memset(pad, 0, sizeof(OSContPad) * __osMaxControllers);
 
-    auto inputEditor = LUS::Context::GetInstance()->GetWindow()->GetGui()->GetGuiWindow("Input Editor");
-    if (!inputEditor || !inputEditor->IsVisible()) {
+    auto inputEditor = std::reinterpret_pointer_cast<LUS::InputEditorWindow>(LUS::Context::GetInstance()->GetWindow()->GetGui()->GetGuiWindow("Input Editor"));
+    if (!inputEditor || inputEditor->ShouldBlockGameInput()) {
```
briaguya0, line 42: if we just want to get old functionality back all we need to do is change this one line to be
```cpp
if (!inputEditor || inputEditor->IsVisible()) {
```
so we don't block input when the controller config menu is hidden but instead block it when it's visible (which is how it works in spock)

PR #248, opened 2023-05-16
Expose MemoryStream in classes.h

Archez:
We already expose BinaryReader for consumers, but MemoryStream wasn't exposed.

### DIFF
--- a/include/libultraship/classes.h
+++ b/include/libultraship/classes.h
@@ -26,6 +26,7 @@
 #include "controller/DummyController.h"
 #include "binarytools/BinaryReader.h"
 #include "binarytools/MemoryStream.h"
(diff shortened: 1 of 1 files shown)


include/libultraship/classes.h
```diff
@@ -25,6 +25,7 @@
 #include "controller/KeyboardScancodes.h"
 #include "controller/DummyController.h"
 #include "binarytools/BinaryReader.h"
+#include "binarytools/MemoryStream.h"
```
Kenix3, line 28:
```suggestion
#include "binarytools/MemoryStream.h"
#include "binarytools/BinaryWriter.h"
```
Might as well also include BinaryWriter if both of these will be there.

PR #287, opened 2023-05-23
Changed calculation for `threadCount` to prevent unsigned underflow.

Malkierian:
If the `reservedThreadCount` in `ResourceManager` was bigger than `hardware_concurrency`, the comparison for `std::max` would underflow to max positive value when calculating `hardware_concurrency() - reservedThreadCount - 1`, thus trying to create a thread pool with a max uint32_t value, causing an instant crash. This will prevent that underflow in max calculation. Machines with fewer threads available will still have issues, but are generally still mostly playable even in that state with this change.

### DIFF
--- a/src/resource/ResourceManager.cpp
+++ b/src/resource/ResourceManager.cpp
@@ -15,14 +15,14 @@ extern bool SFileCheckWildCard(const char* szString, const char* szWildCard);
 namespace LUS {
 
 ResourceManager::ResourceManager(const std::string& mainPath, const std::string& patchesPath,
-                                 const std::unordered_set<uint32_t>& validHashes, uint32_t reservedThreadCount) {
+                                 const std::unordered_set<uint32_t>& validHashes, int32_t reservedThreadCount) {
     mResourceLoader = std::make_shared<ResourceLoader>();
     mArchive = std::make_shared<Archive>(mainPath, patchesPath, validHashes, false);
 #if defined(__SWITCH__) || defined(__WIIU__)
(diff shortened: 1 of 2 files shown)


src/resource/ResourceManager.cpp
```diff
@@ -22,7 +22,7 @@ ResourceManager::ResourceManager(const std::string& mainPath, const std::string&
     size_t threadCount = 1;
 #else
     // the extra `- 1` is because we reserve an extra thread for spdlog
-    size_t threadCount = std::max(1U, (std::thread::hardware_concurrency() - reservedThreadCount - 1));
+    size_t threadCount = std::max(1, (int)(std::thread::hardware_concurrency() - reservedThreadCount - 1));
```
Kenix3: Change reservedThreadCount to an int32_t. Also should be using int32_t rather than int.

Kenix3 (changes requested):

src/resource/ResourceManager.cpp
Malkierian: What would changing reservedThreadCount to int32 accomplish?

Kenix3: You want to work in signed because the underflow in unsigned doesn't allow you to test correctly. It accomplishes working in the correct data range.

Also, as mentioned, we use int32_t rather than int.

Malkierian: I didn't mean that to seem like I was asking why I should bother, it just didn't seem like it would change anything functionally.

PR #296, opened 2023-05-30
Remove skip over ports 2 and 4

inspectredc:
Not quite sure if there was any reason why the slots were increased again here but it resulted in the input configs for the 2nd and 4th ports to not be saved


Kenix3 (approved):
LGTM

PR #298, opened 2023-05-31
Fixed various V-Sync issues

Spodi:
Made V-Sync actual V-Sync off, as flip model requires an additional flag for that. Fixed V-Sync toggle doing nothing, by setting present interval according to the toggle. off = 0, on =1. As a result this doesn't use the calculated present interval anymore and never goes above 1. Values above 1 (aka fractional V-Sync) turned off VRR. Dynamically and rapid switching the interval between 0 or 1 and >1 where the cause for massive stutters with VRR.

### DIFF
--- a/src/graphic/Fast3D/gfx_dxgi.cpp
+++ b/src/graphic/Fast3D/gfx_dxgi.cpp
@@ -599,7 +599,6 @@ static void gfx_dxgi_swap_buffers_begin(void) {
         ThrowIfFailed(dxgi.swap_chain->Present(dxgi.length_in_vsync_frames, 0));
     }
 
-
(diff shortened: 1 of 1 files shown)


Kenix3:
Looks like you need to run clang-format. See CONTRIBUTING.md for troubleshooting this 

Windows is also failing to build.

Kenix3:
Windows failure looks to be an intermittent vcpkg error. Rerunning your action.

Spodi:
Still feels like a bandaid-fix. It works, but somebody should

Spodi:
I feel like this is more like a band-aid fix than anything, but hey it works (for me at least).

PR #301, opened 2023-06-04
Do not use SwapChain interface version as a check for Waitable SwapChain support

Kaldaien:
It's entirely possible for the D3D runtime to implement IDXGISwapChain2 while the SwapChain was never created with the appropriate flag to use a frame latency waitable object.

ALWAYS check the SwapChain flags.


Kenix3:
Windows is failing to build because an intermediate issue with GitHub actions.

Clang format is failing, though, which does need to be addressed. You can see contributing.md with information on how to troubleshoot this.

Kenix3:
This time was not intermittent and vcpkg had an update that broke compatibility. That has since be fixed. Please rebased against current master to fix windows build issues.

PR #303, opened 2023-06-04
Prevent deadlock during gfx_dxgi_set_cursor_visibility

Kaldaien:
If you continue calling `ShowCursor (...)` until you "get the value you want" and there is no mouse attached, it will lock the software up because Windows always returns -1 if there is no mouse.


Kenix3:
Any reason for this being closed?

PR #307, opened 2023-06-07
utilize the user configuration directory for configs and savefiles

Alto1772:
This change is based from the [patch in this unofficial AUR package](https://aur.archlinux.org/cgit/aur.git/tree/lus-install-paths.patch?h=soh&id=1ec450651c23837ed1a158e2d61440c732b72710) and has been implemented for quite some time now, so the goal of it is to address or introduce some issues or ideas related to installation paths on Linux systems or anything path related.

Some of the few key changes are:

- The **`Context::GetAppDirectoryPath`** function now returns the location of the user's application data directory using [`SDL_GetPrefPath`](https://wiki.libsdl.org/SDL2/SDL_GetPrefPath). This means any game configurations or save files will be stored and read from that directory. This can be overridden by the `SHIP_HOME` environment variable. By default, this function returns `~/.local/share/soh` on Linux or `%APPDATA%\soh` on Windows.
- The **`Context::GetAppBundlePath`** (replaced with *`Context::GetAppInstallationPath`*) function now hardcodes the installation path utilizing `CMAKE_INSTALL_PREFIX` instead of the `SHIP_BIN_DIR` environment variable, by activating `NON_PORTABLE` in the configure flags. (In the aforementioned AUR package the installation prefix is set to `/opt/soh`). ~~Also, I've removed the SHIP_BIN_DIR env var as said and done it's only exclusive to Ship.~~ **Update: When on Linux and is portable, it now gives the program's path instead, thus SHIP_BIN_DIR is not needed.**
- Added **`Context::LocateFileAcrossAppDirs`** to search for a given file in the above paths. If the file is not found, it prepends `./` to the file path without checking. 

~~Note that this modifies where application data (typically configuration and save files) is stored.~~ **Update: I've added the configure option `-DNON_PORTABLE=ON` to seperate portability and system-installed** Now these files are located in the OS-specific directories mentioned above instead of the current directory where the program is running. When the game launches unless you did move them, it will reset to the original state. Let me know what are other problems arise from this. Additionally you can check for any other potential issues that I haven't looked upon.

### DIFF
--- a/src/Context.cpp
+++ b/src/Context.cpp
@@ -272,7 +272,7 @@ std::string Context::GetAppInstallationPath() {
         // Find the last '/' and remove everything after it
         int lastSlash = progpath.find_last_of("/");
         if (lastSlash != std::string::npos) {
-            progpath.erase(last_slash);
+            progpath.erase(lastSlash);
         }
 
         return progpath;
--- a/src/Context.cpp
+++ b/src/Context.cpp
@@ -254,9 +254,9 @@ std::string Context::GetShortName() {
     return mShortName;
 }
 
-std::string Context::GetAppBundlePath() {
-#ifdef SHIP_BIN_DIR
-    return SHIP_BIN_DIR;
+std::string Context::GetAppInstallationPath() {
+#ifdef NON_PORTABLE
+    return CMAKE_INSTALL_PREFIX;
 #else
 #ifdef __APPLE__
     FolderManager folderManager;
@@ -275,6 +275,7 @@ std::string Context::GetAppDirectoryPath(std::string appname) {
     }
 #endif
 
+#ifdef NON_PORTABLE
     if (appname.empty()) {
         appname = GetInstance()->mShortName;
     }
@@ -284,12 +285,13 @@ std::string Context::GetAppDirectoryPath(std::string appname) {
         SDL_free(prefpath);
         return ret;
     }
+#endif
 
     return ".";
 }
 
-std::string Context::GetPathRelativeToAppBundle(const std::string path) {
-    return GetAppBundlePath() + "/" + path;
+std::string Context::GetPathRelativeToAppInstallation(const std::string path) {
+    return GetAppInstallationPath() + "/" + path;
 }
 
 std::string Context::GetPathRelativeToAppDirectory(const std::string path, std::string appname) {
@@ -305,7 +307,7 @@ std::string Context::LocateFileAcrossAppDirs(const std::string path, std::string
         return fpath;
     }
     // app install dir
-    fpath = GetPathRelativeToAppBundle(path);
+    fpath = GetPathRelativeToAppInstallation(path);
     if (std::filesystem::exists(fpath)) {
         return fpath;
     }
--- a/src/Context.cpp
+++ b/src/Context.cpp
@@ -267,7 +267,7 @@ std::string Context::GetAppBundlePath() {
 #endif
 }
 
-std::string Context::GetAppDirectoryPath() {
+std::string Context::GetAppDirectoryPath(const std::string appname) {
 #if defined(__linux__) || defined(__APPLE__)
     char* fpath = std::getenv("SHIP_HOME");
     if (fpath != NULL) {
@@ -275,7 +275,7 @@ std::string Context::GetAppDirectoryPath() {
     }
 #endif
 
-    char* prefpath = SDL_GetPrefPath(NULL, GetInstance()->mShortName.c_str());
+    char* prefpath = SDL_GetPrefPath(NULL, appname.c_str());
     if (prefpath != NULL) {
         std::string ret(prefpath);
         SDL_free(prefpath);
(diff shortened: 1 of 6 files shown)


leggettc18 (commented):
LUS probably shouldn't be the one doing this searching of directories to begin with, instead of this, LUS should probably just be handed file paths and not be doing any logic, and the implementing application should probably be responsible for searching these directories.

Also, the SHIP_BIN_DIR env variable is specific to Ship of Harkinian. Other games would probably want to use a different env variable. I realize that one was already in the codebase, it probably shouldn't have been.

Part of me feels like it _might_ be a decent idea for LUS to still expose a function like `LocateFileAcrossAppDirs` for games to use if they want, but another part of me feels like other libraries for searching the common config/data/xdg folders for these files already exists and should be used by the end-user applications if they want that functionality.

briaguya0:
I would want to make sure we test the use case of having multiple versions of ship with different settings. It seems this logic would always load from the user config dir instead of `.` if one exists, which removes some of the benefits of portability that we currently get by keeping everything in one place. 

For example, if I were to have a release build set up and ready to play, then I download a PR build to test, I wouldn't want that PR build to touch my release build's settings

Alto1772:
Here's my own suggestions:
- If you want to make the program look for data files in current directory, at least I would add a configure option to disable using the preferred user config folder to store configs/saves. SM64EX for example uses the user config directory for any reason.
- For `SHIP_BIN_DIR` I could move it to SoH and change the AppBundlePath functionality and its function to name something akin to giving the installation path. I have tried applying `CMAKE_INSTALL_PATH` to that function (as in the AUR package case) and my concern is that using it by default, by means without adding it to configure options, is still `#define`'d on the  `install_config.h` after configure so that's why I checked to make it sure.

I've always wanted to put this patch in the codebase as this idea was implemented and engraved to the Arch AUR package i mentioned since the package's creation, using patches that adds slight features that can make program's functionality of what an installed package should do. For portability concerns this is a kind of a problem to users who wanted the program to make it portable. But for some users who are willing to install the program like installing a package this is the way I would do. I don't wanna always update the patch in the unofficial package on failed builds because of new code changes.

src/Context.cpp
```diff
@@ -304,6 +302,13 @@ std::string Context::GetAppDirectoryPath() {
     }
 #endif
 
+    char* prefpath = SDL_GetPrefPath(NULL, GetInstance()->mShortName.c_str());
```
Alto1772: `GetInstance()` doesnt work if it's uninitialized yet. I'll change this...

briaguya0:
maybe a CLI argument for `XDG_MODE` or something makes sense? then for installed packages the executable could just be a shell script or something and run the binary with that argument, and for every other instance it would still look in `.`

src/Context.cpp
```diff
@@ -253,38 +254,81 @@ std::string Context::GetShortName() {
...
-    char* fpath = std::getenv("SHIP_BIN_DIR");
-    if (fpath != NULL) {
-        return std::string(fpath);
+    std::string progpath(PATH_MAX, '\0');
+    int len = readlink("/proc/self/exe", &progpath[0], progpath.size() - 1);
+    if (len != -1) {
+        progpath.resize(len);
+
+        // Find the last '/' and remove everything after it
+        int last_slash = progpath.find_last_of("/");
```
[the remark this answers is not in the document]
Kenix3: Please address

src/Context.cpp
```diff
@@ -253,38 +254,81 @@ std::string Context::GetShortName() {
...
+        return progpath;
     }
 #endif
 
     return ".";
+#endif
 }
 
-std::string Context::GetAppDirectoryPath() {
+std::string Context::GetAppDirectoryPath(std::string appname) {
```
Kenix3:
```suggestion
std::string Context::GetAppDirectoryPath(std::string appName) {
```

src/Context.cpp
```diff
@@ -253,38 +254,81 @@ std::string Context::GetShortName() {
...
     return ".";
 }
 
-std::string Context::GetPathRelativeToAppBundle(const std::string path) {
-    return GetAppBundlePath() + "/" + path;
+std::string Context::GetPathRelativeToAppInstallation(const std::string path) {
+    return GetAppInstallationPath() + "/" + path;
+}
+
+std::string Context::GetPathRelativeToAppDirectory(const std::string path, std::string appname) {
```
Kenix3:
```suggestion
std::string Context::GetPathRelativeToAppDirectory(const std::string path, std::string appName) {
```

src/Context.cpp
```diff
@@ -253,38 +254,81 @@ std::string Context::GetShortName() {
...
+    return GetAppDirectoryPath(appname) + "/" + path;
 }
 
-std::string Context::GetPathRelativeToAppDirectory(const std::string path) {
-    return GetAppDirectoryPath() + "/" + path;
+std::string Context::LocateFileAcrossAppDirs(const std::string path, std::string appname) {
```
Kenix3:
```suggestion
std::string Context::LocateFileAcrossAppDirs(const std::string path, std::string appName) {
```

src/Context.h
```diff
@@ -25,10 +25,11 @@ class Context {
...
 
-    static std::string GetAppBundlePath();
-    static std::string GetAppDirectoryPath();
-    static std::string GetPathRelativeToAppDirectory(const std::string path);
-    static std::string GetPathRelativeToAppBundle(const std::string path);
+    static std::string GetAppInstallationPath();
+    static std::string GetAppDirectoryPath(std::string appname = "");
+    static std::string GetPathRelativeToAppDirectory(const std::string path, std::string appname = "");
+    static std::string GetPathRelativeToAppInstallation(const std::string path);
+    static std::string LocateFileAcrossAppDirs(const std::string path, std::string appname = "");
```
Kenix3:
```suggestion
    static std::string GetAppDirectoryPath(std::string appName = "");
    static std::string GetPathRelativeToAppDirectory(const std::string path, std::string appName = "");
    static std::string GetPathRelativeToAppInstallation(const std::string path);
    static std::string LocateFileAcrossAppDirs(const std::string path, std::string appName = "");
```

Kenix3 (approved):

Spodi:
Seems like the Shipwrigth repo needs an update to build with those changes.

PR #308, opened 2023-06-07
Implements Config Versioning

leggettc18:
Implements config versioning in such a way that it does not have to be used. Applications who wish to use config versioning have to opt in by implementing a subclass of `CustomVersionUpdater`, pushing at least one instance of it via `RegisterConfigVersion` and then calling `RunVersionUpdates`. Otherwise configs continue to work exactly as they always have. I've also removed `CreateDefaultSettings`, as their default values are already handled correctly if they are missing from the config file. This pushes us toward the goal of "if a value is in the config file, it was set by the user, if not, it is default and we can freely change it if needed". To avoid the F10->F11 debacle we are running into now. The config versioning then allows SoH (or any other apps that got the `CreateDefaultSetting` stuff before it was deleted) to override the config files to be able to fix the bad shortcut by removing the default values from the config in the update function.

### DIFF
--- a/src/config/Config.cpp
+++ b/src/config/Config.cpp
@@ -273,15 +273,15 @@ bool Config::RegisterConfigVersionUpdater(std::shared_ptr<ConfigVersionUpdater> 
 void Config::RunVersionUpdates() {
     for (auto [_, versionUpdater] : mVersionUpdaters) {
         uint32_t version = GetUInt("ConfigVersion", 0);
-        if (version == versionUpdater->GetVersion()) {
+        if (version < versionUpdater->GetVersion()) {
             versionUpdater->Update(this);
-            SetUInt("ConfigVersion", version + 1);
+            SetUInt("ConfigVersion", versionUpdater->GetVersion());
         }
     }
     Save();
 }
 
-ConfigVersionUpdater::ConfigVersionUpdater(uint32_t version) : mVersion(version) {
+ConfigVersionUpdater::ConfigVersionUpdater(uint32_t toVersion) : mVersion(toVersion) {
 }
 
 uint32_t ConfigVersionUpdater::GetVersion() {
--- a/src/config/Config.cpp
+++ b/src/config/Config.cpp
@@ -272,9 +272,10 @@ void Config::RegisterConfigVersion(std::shared_ptr<ConfigVersionUpdater> version
 void Config::RunVersionUpdates() {
     uint32_t version = GetUInt("ConfigVersion", 0);
     for (uint32_t i = version; i < mVersions.size(); i++) {
-        mVersions[i]->Update(std::shared_ptr<Config>(this));
+        mVersions[i]->Update(this);
         SetUInt("ConfigVersion", i + 1);
     }
+    Save();
 }
 
 } // namespace LUS
--- a/src/config/Config.cpp
+++ b/src/config/Config.cpp
@@ -265,4 +265,16 @@ void Config::SetWindowBackend(WindowBackend backend) {
     }
 }
 
+void Config::RegisterConfigVersion(std::shared_ptr<ConfigVersionUpdater> version) {
+    mVersions.push_back(version);
+}
+
+void Config::RunVersionUpdates() {
+    uint32_t version = GetUInt("ConfigVersion", 0);
+    for (uint32_t i = version; i < mVersions.size(); i++) {
+        mVersions[i]->Update(std::shared_ptr<Config>(this));
+        SetUInt("ConfigVersion", i + 1);
+    }
+}
+
 } // namespace LUS
--- a/src/config/Config.cpp
+++ b/src/config/Config.cpp
@@ -281,7 +281,7 @@ void Config::RunVersionUpdates() {
     Save();
 }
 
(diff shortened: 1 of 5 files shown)


leggettc18:
SoH PR implementing this config versioning: https://github.com/HarbourMasters/Shipwright/pull/2966

Kenix3 (approved):

src/config/Config.h
```diff
@@ -8,6 +8,10 @@
 #include "window/Window.h"
 
 namespace LUS {
+class ConfigVersionUpdater {
+  public:
+    virtual void Update(Config* conf) = 0;
```
Kenix3, line 34: I think it's a more flexible system if this class stores a version. That's how it works in the other places we store versioning.

src/config/Config.cpp
```diff
@@ -265,4 +265,17 @@ void Config::SetWindowBackend(WindowBackend backend) {
     }
 }
 
+void Config::RegisterConfigVersion(std::shared_ptr<ConfigVersionUpdater> version) {
+    mVersions.push_back(version);
+}
+
+void Config::RunVersionUpdates() {
+    uint32_t version = GetUInt("ConfigVersion", 0);
+    for (uint32_t i = version; i < mVersions.size(); i++) {
```
Kenix3: This loop should compare against the version for the class.

src/config/Config.h
```diff
@@ -47,5 +54,6 @@ class Config {
     nlohmann::json mNestedJson;
     std::string mPath;
     bool mIsNewInstance;
+    std::vector<std::shared_ptr<ConfigVersionUpdater>> mVersions;
```
Kenix3: Maybe a map instead of a vector so it can be sorted in a foreach?

Kenix3 (changes requested):

src/config/Config.cpp
```diff
@@ -265,4 +265,17 @@ void Config::SetWindowBackend(WindowBackend backend) {
     }
 }
 
+void Config::RegisterConfigVersion(std::shared_ptr<ConfigVersionUpdater> version) {
+    mVersions.push_back(version);
+}
```
briaguya0: the naming here feels a little confusing, maybe `mVersionUpdaters` and `versionUpdater` instead?

src/config/Config.h
briaguya0: same comment as before about naming, i like `mVersionUpdaters` but i'm open to other ideas

src/config/Config.h
briaguya0: are you thinking it would store the version that the config is being updated *from* or the version it's being updated *to*?

edit: this is a reply to @Kenix3's comment

> I think it's a more flexible system if this class stores a version. That's how it works in the other places we store versioning.

briaguya0 (commented):
tiny comment about naming but i like it!

src/config/Config.h
leggettc18: The version we're updating from. That makes for the easiest way to tell that we are on the latest version. If there's no version updater class matching the version we are on, then we are on the latest version. Or there's a missing version updater which should definitely be considered an error.

src/config/Config.cpp
leggettc18: I agree, will be renaming them to mVersionUpdaters and RegisterConfigVersionUpdater.

src/config/Config.h
leggettc18: Changed to store a class and foreach loop over a map.

src/config/Config.cpp
leggettc18: Fixed to the new names.

src/config/Config.cpp
```diff
@@ -265,4 +265,27 @@ void Config::SetWindowBackend(WindowBackend backend) {
...
+
+void Config::RunVersionUpdates() {
+    for (auto [_, versionUpdater] : mVersionUpdaters) {
+        uint32_t version = GetUInt("ConfigVersion", 0);
+        if (version == versionUpdater->GetVersion()) {
+            versionUpdater->Update(this);
+            SetUInt("ConfigVersion", version + 1);
+        }
+    }
```
briaguya0:
```suggestion
    for (auto [_, versionUpdater] : mVersionUpdaters) {
        uint32_t version = GetUInt("ConfigVersion", 0);
        if (version < versionUpdater->GetVersion()) {
            versionUpdater->Update(this);
            SetUInt("ConfigVersion", versionUpdater->GetVersion());
        }
    }
```

briaguya0 (commented):
i think with this change we can use `mVersion` to mean `toVersion` instead of `fromVersion`

src/config/Config.cpp
Kenix3: This looks like the correct logic to me. This can allow a port to do something like using version codes in the config version.

leggettc18: Hmmm, I wasn't originally convinced that storing the `toVersion` was the correct play. In my head, a `ConfigVersionUpdater` describes the version of the config it's meant to update. But I see now that this is the only way to allow for more freedom in what version numbers implementing apps use. I'll update this today.

leggettc18: Updated

PR #312, opened 2023-06-09
New frame timing approach, fix "Match Refresh Rate" button

### LINKED ISSUE #299: Timer precision issues
https://github.com/Kenix3/libultraship/blob/70c3201a99a5c56b9b56a53595a6cb3cf8de3fc4/src/graphic/Fast3D/gfx_dxgi.cpp#L585-L588

I would like to point out that the Windows scheduler cannot actually do 100 ns on a wait, even at the highest possible timer resolution :)

For proper framepacing, you should adopt something more conservative:

```
 Timer = Deadline - 1 ms
 Busy-wait the remaining time, checking QPC and calling YieldProcessor (...).
```

I know this sounds inefficient, but in fact it is not.

The amount of time that you'll spend busy-waiting because the next frame deadline is less than 1 ms is probably 10-20% of your total wait time at 60 FPS. And you'll never -miss- a frame, unlike the current design that's constantly going to return after the frame deadline (because, again, 100 ns scheduling is ridiculous).

Spodi:
- Minor fixes for edge cases
   - fix lockup if no mouse connected
   - better check for waitable SwapChain
   - no need to recreate the SwapChain when jitter fix is turned off
- "Match Refresh Rate" button uses the refresh rate of the monitor the window is on
   - also fixes it on VRR displays
- New frame timing approach
   - always use timer instead of relying on V-Sync
   - improved timer precision to minimize missed frames (also fixes #299)
   - Sadly one oddly specific downside I coudn't fix:
   If the window is on a secondary monitor that runs on a lower refresh rate than the primary, the rendering is also composed via DWM (Composed: Flip) and you're setting the FPS higer than your primary monitor: then it starts to stutter horribly. But one shoudn't push the limiter higher up than their refresh rate with V-Sync on anyway. 😛

Thanks to Aemony and Kaldaien from Special K, who helped a lot with this.

### DIFF
--- a/src/graphic/Fast3D/gfx_dxgi.cpp
+++ b/src/graphic/Fast3D/gfx_dxgi.cpp
@@ -147,8 +147,11 @@ template <typename Fun> static void run_as_dpi_aware(Fun f) {
 }
 
 static void apply_maximum_frame_latency(bool first) {
+    DXGI_SWAP_CHAIN_DESC       swap_desc = {};
+    dxgi.swap_chain->GetDesc (&swap_desc);
+    
     ComPtr<IDXGISwapChain2> swap_chain2;
-    if (dxgi.swap_chain->QueryInterface(__uuidof(IDXGISwapChain2), &swap_chain2) == S_OK) {
+    if ((swap_desc.Flags & DXGI_SWAP_CHAIN_FLAG_FRAME_LATENCY_WAITABLE_OBJECT) && dxgi.swap_chain->QueryInterface(__uuidof(IDXGISwapChain2), &swap_chain2) == S_OK) {
         ThrowIfFailed(swap_chain2->SetMaximumFrameLatency(dxgi.maximum_frame_latency));
         if (first) {
             dxgi.waitable_object = swap_chain2->GetFrameLatencyWaitableObject();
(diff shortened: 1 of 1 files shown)


briaguya0:
> If the window is on a secondary monitor that runs on a lower refresh rate than the primary, the rendering is also composed via DWM (Composed: Flip) and you're setting the FPS higer than your primary monitor:

just want to make sure i'm understanding this correctly:
primary monitor 144
secondary monitor 60
window on secondary, FPS set to 60, things work?
window on secondary, FPS set to 144, things work?
window on secondary, FPS set to 145, things break?

Spodi:
Yes it is exactly like you described, well almost: it only happens if V-Sync is on too. But in borderless fullscreen everything works like expected.
This is due to windows DWM. It always V-syncs to the primary monitor.

Spodi:
There is a some other code in this file, that is getting the current display. As my code stores all displays, we can probably use that for this too. Might look at that and try later.
Does this need some fallback if the window isn't on any screen at all for some reason? Can that happen?

Spodi:
Closed in favor of those:

https://github.com/Kenix3/libultraship/pull/325
https://github.com/Kenix3/libultraship/pull/326
https://github.com/Kenix3/libultraship/pull/327

PR #315, opened 2023-06-15
[WIP] Add CMake option to disable building against GLEW.

vanfanel:
This adds `NO_GLEW` as a CMake option (`-DNO_GLEW=ON`) to build libultraship against SDL2+OpenGL only, thus eliminating the huge pack of dependencies GLEW brings.

SDL2 is the way to abstract windowing systems: It can initialize an OpenGL/GLES context on any operating system.


briaguya0:
SoH PR here https://github.com/HarbourMasters/Shipwright/pull/3010

briaguya0:
My thought here is if the SoH testing reveals we don't need GLEW, we can get this PR merged and then stop using it in SoH. Then we can make an issue to deprecate it in LUS and discuss the pros/cons of keeping/removing it there.

briaguya0:
i'm going to close this. it doesn't fully remove glew (for example, on windows it is still included but init is never called, which is not how it should be used)

it's also not something that actually benefits any of our currently supported platforms. if you're trying to build out support for another platform i recommend working from a fork until you have a good proof of concept. once that is in place we'll have a better idea of what `#defines` it makes sense to have

PR #316, opened 2023-06-15
yeet glew

briaguya0:
based on this https://discourse.libsdl.org/t/do-i-need-glew/33523/3 it seems we're better off using GLAD if we run into any issues with GL extensions not loading properly


briaguya0:
if we want to do this we should probably just move to [glad](https://glad.dav1d.de/)

PR #322, opened 2023-06-23
Fix bogus type conversion for window position (dxgi)

Spodi:
Related to https://github.com/HarbourMasters/Shipwright/issues/3028. The window should now save and use coordinates correctly when on a monitor above or below the primary (negative coordinates).
Does NOT fix the window going off screen when your screen layout changed in between two starts.

EDIT: removed "fix" keyword, so it doesn't close the issue (no full fix).

### DIFF
--- a/src/graphic/Fast3D/gfx_dxgi.cpp
+++ b/src/graphic/Fast3D/gfx_dxgi.cpp
@@ -8,6 +8,7 @@
 #include <string>
 
 #include <windows.h>
+#include <windowsx.h> // GET_X_LPARAM(), GET_Y_LPARAM()
 #include <wrl/client.h>
 #include <dxgi1_3.h>
 #include <dxgi1_4.h>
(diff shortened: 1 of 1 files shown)


briaguya0 (approved):
Based on the documentation for [GET_X_LPARAM](https://learn.microsoft.com/en-us/windows/win32/api/windowsx/nf-windowsx-get_x_lparam#remarks)
> Use GET_X_LPARAM instead of [LOWORD](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/legacy/ms632659(v=vs.85)) to extract signed coordinate data. Negative screen coordinates may be returned on multiple monitor systems.

and [GET_Y_LPARAM](https://learn.microsoft.com/en-us/windows/win32/api/windowsx/nf-windowsx-get_y_lparam#remarks)

> Use GET_Y_LPARAM instead of [HIWORD](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/legacy/ms632657(v=vs.85)) to extract signed coordinate data. Negative screen coordinates may be returned on multiple monitor systems.

This looks like the correct way to address this issue!

:shipit:

Kenix3 (approved):

PR #323, opened 2023-06-24
Use current display for fullscreen size (SDL)

Spodi:
Used to always use the dimensions of the primary monitor, which of course lead to weird behavior, when not on the primary monitor. Related to https://github.com/HarbourMasters/Shipwright/issues/2991?
Does NOT fix the window going off screen when your screen layout changed in between two starts.

EDIT: removed "fix" keyword, as I can't test Linux and I actually don't know if this fixes that. I know it fixes some issues in Windows.


Spodi:
Maybe related to https://github.com/HarbourMasters/Shipwright/issues/2991

Spodi:
Integrated into https://github.com/Kenix3/libultraship/pull/326.
sry, but this fix was needed to get things going with the other PR...

PR #326, opened 2023-06-26
Display handling

### LINKED ISSUE #297: Fullscreen doesn't save monitor
If I exited SoH while it was in fullscreen and at my second monitor, the next start is always in fullscreen at my primary monitor. after that toggling back to windowed makes the window jump back to my secondary monitor.

TLDR: Monitor saves fine for windowed mode. It doesn't in fullscreen.

SoH build with PR https://github.com/HarbourMasters/Shipwright/pull/2881

Spodi:
Small rework of how monitors and monitor changes are handled
- [DirectX] "Match Refresh Rate" button uses the refresh rate of the monitor the window is on (also fixes it on VRR displays)
- [DirectX, OpenGL] Fallback to default position (100,100 on primary), if top left corner of the window is not on any monitor
- [OpenGL] Fix fullscreen size for secondary displays
- [DirectX, OpenGL] Fullscreen re-engages on the correct monitor if you exited in fullscreen 

Split off from https://github.com/Kenix3/libultraship/pull/312, so testing things is hopefully easier.

Should fix https://github.com/Kenix3/libultraship/issues/297 and should also fix https://github.com/HarbourMasters/Shipwright/issues/3028.

### DIFF
--- a/src/graphic/Fast3D/gfx_dxgi.cpp
+++ b/src/graphic/Fast3D/gfx_dxgi.cpp
@@ -69,8 +69,11 @@ static struct {
     std::map<UINT, DXGI_FRAME_STATISTICS> frame_stats;
     std::set<std::pair<UINT, UINT>> pending_frame_stats;
     bool dropped_frame;
+    std::tuple<HMONITOR, RECT, BOOL> h_Monitor; // 0: Handle, 1: Display Monitor Rect, 2: Is_Primary
+    std::vector<std::tuple<HMONITOR, RECT, BOOL>> monitor_list;
     bool zero_latency;
-    float detected_hz;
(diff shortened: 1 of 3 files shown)


Spodi:
Related to https://github.com/HarbourMasters/Shipwright/issues/2991?

Spodi:
This should also make `shipofharkinian.json` sharable as is, without the the window disappearing into nothingness for the other person.

The window can still disappear under the taskbar in some scenarios, but so can Firefox, Discord and GitHub desktop... 🤷 Press `WIN + Up` while SoH has focus and then move the window to a proper place if that ever happens.

Kenix3 (approved):

PR #328, opened 2023-06-26
add RomToBigEndian method to rewrite rom data as z64 format

Archez:
Adds a method to ZAPDUtils to convert rom data into z64 format. This can be leveraged by ZAPDTR, OTRExporter, and the SoH built-in extractor to support z64, n64, and v64 rom formats.

If it is preferable to return a new array instead of modifying in-place the existing array, then I can make that change.

### DIFF
--- a/extern/ZAPDUtils/Utils/BitConverter.h
+++ b/extern/ZAPDUtils/Utils/BitConverter.h
@@ -181,4 +181,25 @@ public:
 		std::memcpy(&value, &floatData, sizeof(value));
 		return value;
 	}
+
+	// Rewrites the rom data in-place to be in BigEndian/z64 format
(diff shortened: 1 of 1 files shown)


extern/ZAPDUtils/Utils/BitConverter.h
```diff
@@ -181,4 +181,25 @@ class BitConverter
 		std::memcpy(&value, &floatData, sizeof(value));
 		return value;
 	}
+
+	// Rewrites the rom data in-place to be in BigEndian/z64 format
+	static inline void RomToBigEndian(uint8_t* rom, size_t romSize) {
+		// Use the first byte to determine byte order
+		uint8_t firstByte = rom[0];
```
NEstelami, line 192: It would probably be a good idea to check `romSize > 0` before reading from `rom`.

PR #330, opened 2023-07-10
Better framepacing - Windows SDL Edition

Spodi:
New frame timing approach for OpenGL renderer:
- improved timer precision to minimize missed frames

Hopefully helps non-windows a bit too. -> nope need help for other architectures and compilers


Spodi:
This won't work. Maybe I can do it for Windows only. I don't want to mess with compiler and instruction sets 😅
Maybe someone else can. Trying to do a busy-wait here and windows YieldProcessor() is pretty much what I want here. But ofc this isn't available on Linux and Mac

Kenix3 (approved):

PR #331, opened 2023-07-18
Allow stick deadzones to be applied for thumbstick -> button mapping

nick-michael:
When mapping thumbsticks to buttons the deadzones are ignored. There is an option `AxisMinimumPress` to control the behaviour, but there is no UI option to actually modify it, so it's hardcoded as `7680.0f` and can only be changed by modifying the config JSON directly.

Rather than making a new thumbstick UI with deadzones I figured it made sense to add a boolean checkbox that would allow you to apply your thumbstick deadzones for thumbstick -> button mapping.

I'm not really familiar with drawing UI in C++ so most of the options menu stuff is copied/pasted & modified from existing code. Please shout if it's wrong and let me know what to change 🙏


Kenix3:
clang-format is failing, please see https://github.com/Kenix3/libultraship/blob/main/CONTRIBUTING.md#running-clang-format

Kenix3:
Merged via https://github.com/Kenix3/libultraship/pull/332

PR #333, opened 2023-07-25
Window positon tweaks and fullscreen in OpenGL

Spodi:
* Fixed window positon not restored (OpenGL). Broke with my display handling changes - oops
* Fixes going full screen on the next monitor when the game was on the second half of the screen. (OpenGL)
* Consistent behavior for getting monitor between DirectX and OpenGL.
* Window position now only gets reset, if there is no pixel of the window on any screen. (DirectX)

### DIFF
--- a/src/graphic/Fast3D/gfx_sdl2.cpp
+++ b/src/graphic/Fast3D/gfx_sdl2.cpp
@@ -223,19 +223,21 @@ static void set_fullscreen(bool on, bool call_callback) {
     if (fullscreen_state == on) {
         return;
     }
-    fullscreen_state = on;
     int display_in_use = SDL_GetWindowDisplayIndex(wnd);
+    if (display_in_use < 0) {
+        SPDLOG_WARN("Can't detect on wich monitor we are. Probably out of display area?");
+        SPDLOG_WARN(SDL_GetError());
+    }
 
     if (on) {
         // OTRTODO: Get mode from config.
         SDL_DisplayMode mode;
-        if (display_in_use < 0) {
-            SDL_GetDesktopDisplayMode(0, &mode);
+        if (SDL_GetDesktopDisplayMode(display_in_use, &mode) >= 0) {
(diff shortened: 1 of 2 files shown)


src/graphic/Fast3D/gfx_sdl2.cpp
```diff
@@ -223,20 +223,21 @@ static void set_fullscreen(bool on, bool call_callback) {
...
         return;
     }
-    fullscreen_state = on;
     int display_in_use = SDL_GetWindowDisplayIndex(wnd);
-    SDL_DisplayMode mode;
     if (display_in_use < 0) {
-        SDL_GetDesktopDisplayMode(0, &mode);
-    } else {
-        SDL_GetDesktopDisplayMode(display_in_use, &mode);
+        SPDLOG_WARN("Can't detect on wich monitor we are. Probably out of display area?");
```
Kenix3:
```suggestion
        SPDLOG_WARN("Can't detect on which monitor we are. Probably out of display area?");
```

src/graphic/Fast3D/gfx_sdl2.cpp
```diff
@@ -223,20 +223,21 @@ static void set_fullscreen(bool on, bool call_callback) {
...
+        SPDLOG_WARN(SDL_GetError());
```
Kenix3, line 229: Does this return a string or an error code? If it's an error code we might want to put some text here.

Kenix3 (commented):
Just one minor nitpick

src/graphic/Fast3D/gfx_sdl2.cpp
Spodi: SDL dogs says it is a string:
Returns a message with information about the specific error that occurred, [...]
https://wiki.libsdl.org/SDL2/SDL_GetError

Spodi:
I probably overcomplicated position checks in DirectX. but hey, at least it should be performant.

PR #349, opened 2023-09-23
[ci] silently continue when trying to delete ccache.exe if it doesn't exist

briaguya0:
same fix as https://github.com/HarbourMasters/Shipwright/pull/3222


Kenix3:
Thankyou for this!

PR #350, opened 2023-09-25
controller rework

briaguya0:
~note: don't let the lines changed number scare you, a lot of stuff is commented out instead of removed at the moment~ no longer true

soh PR to get builds https://github.com/HarbourMasters/Shipwright/pull/3230

### todo:
- [x] axis -> axis raw press mapping
- [x] sdl button to axis
- [x] keyboard
- [x] rumble
- [ ] migrate from old configs
- [ ] wii u
- [x] analog deadzone and notch angles (save/load config and use those values)
- [ ] set defaults button (with controller selection popup/dropdown)
- [x] clear all does more than just buttons
- [ ] reset to default buttons for all non-mapping options (rumble strength, analog deadzone, etc.)
- [x] ~do uuids in a good way~ don't use uuids
- [x] use colors to show which device index mappings are from
  - [x] show details about mappings on hover
- [x] stuff from https://github.com/Kenix3/libultraship/pull/350/commits/8a69f248fe62a878f4c1cec53c10d73918a71f08
- [x] move mapping creation into factories
- [x] clean up game input blocking
- [ ] deadzone for buttons (`AxisDirectionToButtonMapping`) - the current implementation has "use axis deadzone for buttons" but that feels kinda messy and it won't work with the new architecture. a global threshold for presses is probably a good thing to have, but that doesn't take into account different thresholds for triggers vs sticks, so just allowing the user to configure one value isn't quite good enough.
- [x] led
- [x] gyro
- [ ] unknown unknowns/things i'm forgetting about rn

### v2 todo (decided it's out of scope for this PR but it'd still be cool):
- attachments
  - i see the reasoning here, but i don't see a clear path forward for this given the current limitations in how we handle "controller output." i plan to build things in a modular way that should make moving to an attachment system straightforward, but i think it makes sense to make those architectural decisions in a PR separate from this one
- some sort of "switch device" functionality. the set defaults functionality should be decent for v1, but being able to say "take all of the mappings on this port currently applied to sdl device index 2 and switch them to apply to sdl device index 0" could be nice. not exactly sure what the UX for that would be, but it seems doable
  - this could also be done via saving/loading profiles
  - another option would be to allow users to configure what "reset to defaults" does. basically saving defaults so anytime the user resets it uses those. this could get a little meta because we'd need to have a "reset defaults to default" option
  - maybe allow switching a device's color (as seen in mapping button colors and on the devices tab). this would not just be cosmetic, it will function as a way to redirect mappings to a different controller. if the mappings are all red, then you have 3 controllers connected showing red/blue/green, you can switch the green one to red to use the saved mappings on the new controller.
- "chord" style mapping - make it possible to require multiple physical buttons be pressed to activate a virtual button/axis
- https://github.com/Kenix3/libultraship/issues/351
- (super nitpicky, might ignore this) get it so the controller config menu remembers the ordering of mappings when closing and opening the game
- mapping stuff when imgui controller nav is enabled. it's currently really messy and i don't feel like diving into it
  - pressing "A" on something to start mapping maps A to it
  - clicking on something to map and then pressing "B" just closes the popup and doesn't map it

current behavior:
image
then quit and restart game and sometimes see
image
i know this has to do with `UpdateBitmaskToMappingUuids` in `InputEditorWindow.cpp` not having any sense of the order they were added in, not sure how much complexity this would add/if it's worth it

### maybe v2, maybe not (might be cool, might not need it)
- add a tab to display a list of all devices
  - colors associated
  - buttons pressed
  - button to force rumble?


briaguya0:
closing this and opening a new one because i can't tell what tidy is talking about in fixes.yml

PR #353, opened 2023-10-01
Fix color24

### LINKED ISSUE #352: `ConsoleVariable::SetColor24` doesn't properly save
when we set using `SetColor24` the `Color24` variable is set

https://github.com/Kenix3/libultraship/blob/6671d7e4984fe609928239937d5525b12a4b170d/src/config/ConsoleVariable.cpp#L127-L135

but then when we go to save the `Color` variable is read

https://github.com/Kenix3/libultraship/blob/6671d7e4984fe609928239937d5525b12a4b170d/src/config/ConsoleVariable.cpp#L186-L199

leading to `{0, 0, 0}` always being saved for `Color24`

briaguya0:
fixes https://github.com/Kenix3/libultraship/issues/352

not sure if you had a different way to handle this in mind @Kenix3  but imo this is way better than having it be straight up broken

### DIFF
--- a/src/config/ConsoleVariable.cpp
+++ b/src/config/ConsoleVariable.cpp
@@ -186,9 +186,15 @@ void ConsoleVariable::Save() {
         } else if (variable.second->Type == ConsoleVariableType::Color ||
                    variable.second->Type == ConsoleVariableType::Color24) {
             auto keyStr = key.c_str();
-            conf->SetUInt(StringHelper::Sprintf("%s.R", keyStr), variable.second->Type == ConsoleVariableType::Color ? variable.second->Color.r : variable.second->Color24.r);
-            conf->SetUInt(StringHelper::Sprintf("%s.G", keyStr), variable.second->Type == ConsoleVariableType::Color ? variable.second->Color.g : variable.second->Color24.g);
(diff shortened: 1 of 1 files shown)


Malkierian:
Could you do an `auto` on the grabbing of the variable from `variable->second` and keep the `clr.r/g/b` intact for the rest of it and have it work?

briaguya0:
@Malkierian no, `Color` and `Color24` are separate members of the cvar struct

https://github.com/Kenix3/libultraship/blob/6671d7e4984fe609928239937d5525b12a4b170d/src/config/ConsoleVariable.h#L13-L21

if we wanted to do a bigger rework we could change that struct to have
```cpp
typedef struct CVar {
    const char* Name;
    ConsoleVariableType Type;
    int32_t Integer;
    float Float;
    std::string String;
    uint8_t ColorR;
    uint8_t ColorG;
    uint8_t ColorB;
    uint8_t ColorA;
} CVar;
``` 
and then just return a Color or Color24 using those on Get or something

Kenix3 (approved):
Not super thrilled with the complicated ternaries, but I don't like the fact that we have a struct on the cvar at all, so fixing that feels out of scope.

Issue #351, opened 2023-09-29
deadzone math


briaguya0:
i would like to allow users to set a deadzone with a unit that means something. the current deadzone is just a number with no context. i think it would be ideal to have this number be a percentage, with 100% meaning no inputs from the stick are counted because 100% of the stick range is included in the deadzone.

if SDL always reported a value that fit within a circle this wouldn't be a problem. `sqrt(x^2 + x^2)` would be at max `MAX_AXIS_RANGE` and we could just have the user set a percentage of `MAX_AXIS_RANGE`, but we can't count on that being the case.

in my testing on an xbox controller (with a circular gate on the analog stick), i've been able to get normalized values (`MAX_AXIS_RANGE/MAX_SDL_RANGE`), that give a radius of closer to `90` (as opposed to `85`) (~64 and ~63 for x and y respectively)

for https://github.com/Kenix3/libultraship/pull/350 i'm going to go ahead and just do a percentage of `MAX_AXIS_RANGE`, as that still feels better than what we have now

![image](https://github.com/Kenix3/libultraship/assets/70942617/5f275495-eeae-4128-befd-47771003e1b6)

# tl;dr

try setting the deadzone in the current controller config window somewhere between 85 and 95 and look at the input preview, you'll see the oddities

PR #354, opened 2023-10-02
controller rework

briaguya0:
soh PRs to get builds:
Single Player Mode: https://github.com/HarbourMasters/Shipwright/pull/3331
Multiplayer Mode: https://github.com/HarbourMasters/Shipwright/pull/3354

### todo:
- [x] set defaults on reorder after clear even if we already have a device index mapping ~i think clear isn't clearing the device index mapping ids? pressing clear all and then reorder isn't setting defaults~
- [x] handle using controller to nav imgui and map controllers
- [x] axis -> axis raw press mapping
- [x] sdl button to axis
- [x] keyboard
- [x] rumble
- [x] wii u
  - [x] figure out why release mode is being wonky (issue with newer devkit pro, it's a problem on latest dev too, resolved by not upgrading devkitpro for now)
- [x] analog deadzone and notch angles (save/load config and use those values)
- [x] set defaults button (with controller selection popup) https://github.com/Kenix3/libultraship/pull/354#issuecomment-1783775439
- [x] save device name to sdl mapping along with guid (so we can see "xbox controller (disconnected)")
- [x] clear all does more than just buttons
- [x] reset to default buttons for all non-mapping options (rumble strength, analog deadzone, etc.)
- [x] ~do uuids in a good way~ don't use uuids
- [x] use colors to show which device index mappings are from
  - [x] show details about mappings on hover
- [x] stuff from https://github.com/Kenix3/libultraship/pull/350/commits/8a69f248fe62a878f4c1cec53c10d73918a71f08
- [x] move mapping creation into factories
- [x] clean up game input blocking
- [x] deadzone for buttons (`AxisDirectionToButtonMapping`) - the current implementation has "use axis deadzone for buttons" but that feels kinda messy and it won't work with the new architecture. a global threshold for presses is probably a good thing to have, but that doesn't take into account different thresholds for triggers vs sticks, so just allowing the user to configure one value isn't quite good enough.
  - [x] cog on mappings
    - [x] slider
- [x] led
- [x] gyro
- [x] remove use of + to mean dpad https://github.com/HarbourMasters/Shipwright/pull/3230#issuecomment-1742413428
- [x] put color around controller name text in rumble options
- [x] add colored font awesome icon for each controller with a mapping in a section
- [x] add modal warning users that everything will be cleared (be explicit) for clear all
- [x] analog x and y under preview for analog sticks (just ints)
- [x] devices tab
  - [x] list of all devices and associated color
- [x] handle controller connect/disconnect events on directx
- [x] decide on UX for showing reorder button on disconnected screen when only 1 physical device connected
- [x] fix keyboard mapping to multiple buttons (1 key to multiple n64 inputs) https://github.com/Kenix3/libultraship/pull/354/commits/0f2b706e8006601a36ff136b6fd57f10fe713c47
- [x] some wonkyness with connecting/disconnecting - need to do more investigation
- [x] handle mutliple controllers disconnected when window up
- [x] get singleplayer mode working well

### v2 todo (decided it's out of scope for this PR but it'd still be cool):
- figure out wonkyness with wiimote+nunchuk mappings (it seems nunchuck buttons aren't getting set but the wiimote equivalent are - this is happening in KPadStatus - `hold` vs `nunchuck.hold`)
- pressing the menubar toggle button on a controller toggles it twice
  - imgui logic for reading from controllers feels kind of wonky in general, i tried a few things here and nothing felt consistent, i think a true solution to this issue will come with remappable hotkeys outside of imgui (i'm thinking another tab like ports that has hotkey config, then we can use the controldeck and controller and read from the devices with a `___ToHotkeyMapping` (this would get us remappable hotkeys)
- game pausing for disconnect/remap screens
- disconnected controllers as vector to show which ports are disconnected when multiple disconnected
- add colored font awesome icon for each controller with a mapping in a port tab
  - imgui tabs just have a label, trying to get multiple colors in a tab label is far from easy in imgui at the moment https://github.com/ocornut/imgui/issues/902
- persist controller button names when controller is disconnected
- handle disconnected controller mappings more elegantly, right now they just fill up the config (this might become an issue with manual clearing, in which case it'll move up to v1)
- move some of the shared logic out of `ControllerStick`/`ControllerButton`/`ControllerGyro`/`ControllerRumble`/`ControllerLED` (maybe just wait and do this as part of attachments)
- attachments
  - i see the reasoning here, but i don't see a clear path forward for this given the current limitations in how we handle "controller output." i plan to build things in a modular way that should make moving to an attachment system straightforward, but i think it makes sense to make those architectural decisions in a PR separate from this one
- "chord" style mapping - make it possible to require multiple physical buttons be pressed to activate a virtual button/axis
- https://github.com/Kenix3/libultraship/issues/351
- (super nitpicky, might ignore this) get it so the controller config menu remembers the ordering of mappings when closing and opening the game (see nitpick appendix)
- mapping stuff when imgui controller nav is enabled. it's currently really messy and i don't feel like diving into it
  - pressing "A" on something to start mapping maps A to it
  - clicking on something to map and then pressing "B" just closes the popup and doesn't map it
- wrap mappings to next line instead of overflowing to the right, put analog stick preview to the right
- advanced devices tab functionality
  - show buttons pressed
  - button to force rumble
  - button to force LED
  - sticks
  - gyro
- TAS
  - flag on controller that short circuits ReadToPad and reads from tas instead
  - flag on controller that writes to tas file
  - need to byteswap https://github.com/KiritoDv/libultraship/commit/bd01dd70f543fea33faa0a6a41e72a07afbaabd4
  - ui needs to make it clear that restarting is needed (maybe force now?)

~migrate from old configs~

#### nitpick appendix:
current behavior:
image
then quit and restart game and sometimes see
image
i know this has to do with `UpdateBitmaskToMappingUuids` in `InputEditorWindow.cpp` not having any sense of the order they were added in, not sure how much complexity this would add/if it's worth it


PurpleHato:
I know it's gonna sounds very picky, but when we're listing directions, the norm is to list in this manner: Up, Down, Left, Right
So seeing Left, Right, Up, Down is feeling wrong
I know it's just visual and everything :P

briaguya0:
set defaults details from discord conversation

* if we do "set default mappings" and we don't have any mappings for that device type on that port then we make a new LUSDeviceIndex that we associate with that sdl device and make defaults

* for the case of "we already have mappings for this controller type on this port" where "set defaults" is picked, just show a modal
```
this will clear all existing mappings for {controller type} on port {port #}.

cancel 
set defaults
```

PR #355, opened 2023-10-07
Make app fully DPI aware

Spodi:
Needs SoH PR: https://github.com/HarbourMasters/Shipwright/pull/3270

Adds `Window::GetDPI()` and `Window::GetDPIScale()`

OpenGL support and other OS than Windows is experimental.
Sadly SDL on windows seems to sometimes ignore a Dpi change and doesn't scale the window when it should.


Spodi:
To clarify things:
It's still a draft because of SDL. I can't test anything other than Windows. And even there it behaves weird. It randomly ignores DPI changes and then does not scale. Although the function still outputs the right value. Also I think for non-Windows SDL outputs the physical DPI, while on Windows you get some virtual one (the one that is used for scaling). And I always assume 96 DPI are 100%, no matter the OS for now. (It's what windows uses as 100%). 

Do we even want the window to scale automatically with DPI changes? I also could not stop SDL from doing the window scaling either btw.

PR #341, opened 2023-09-17
Keyboard resize fix

Spodi:
Use the function that was already there for the message queue and just run dxgi.run_one_game_iter() in a loop. This should be more in line with how this works in SDL/OpenGL.

Fixes bugged (jittery) resize/move with keyboard through the window menu. Although the game pauses when moved/resized or the window menu is open (also like in SDL/OpenGL).

Before:
https://github.com/Kenix3/libultraship/assets/962946/c2e9f37c-31f8-4d32-ad3e-0f595faf40c8

After the changes:
https://github.com/Kenix3/libultraship/assets/962946/d6b3afdd-a53a-4771-a1e5-ea2c098ae5f8

PR #372, opened 2023-11-12
Partially revert #341

Spodi:
This seems to have caused a discrepancy in DPI awareness and therefore ImGUI does not use the whole window.

In my opinion DPI awareness via manifest would be the right thing.
But i also don't want a broken SoH until that is done 😅

Also somebody other then me should test this, so we are actually sure it works again😆


Archez (approved):
Tested on my windows machine and this does appear to resolve the known DPI issues in Soh MacReady

Kenix3 (approved):

PR #375, opened 2023-11-15
Disable MSAA on feature level 10_0

Spodi:
Using MSAA in this constellation would otherwise crash on 10_0 hardware.

Would probably need a corresponding change in the GUI, so the user is informed about that.

Not really useful, as long the app doesn't even really work on 10_1 hardware (I can't find the reason why)

### DIFF
--- a/src/graphic/Fast3D/gfx_direct3d11.cpp
+++ b/src/graphic/Fast3D/gfx_direct3d11.cpp
@@ -170,7 +170,7 @@ static void create_depth_stencil_objects(uint32_t width, uint32_t height, uint32
     texture_desc.ArraySize = 1;
     texture_desc.Format =
         d3d.feature_level >= D3D_FEATURE_LEVEL_10_0 ? DXGI_FORMAT_R32_TYPELESS : DXGI_FORMAT_R24G8_TYPELESS;
-    texture_desc.SampleDesc.Count = msaa_count;
+    texture_desc.SampleDesc.Count = d3d.feature_level >= D3D_FEATURE_LEVEL_10_1 ? msaa_count : 1;
     texture_desc.SampleDesc.Quality = 0;
     texture_desc.Usage = D3D11_USAGE_DEFAULT;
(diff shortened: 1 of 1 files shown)


Kenix3:
Is this PR still meant to be in draft?

Spodi:
> Is this PR still meant to be in draft?

Yesnt. I dunno how useful this really is. 
This changes makes at least DirectX' debug layer shut up, when testing against feature level 10_0.
However, according to @malkierian 's testings with old Hardware, it seems we also need Compute Shader / Direct Compute support on the hardware we are running at or else SoH is just crashing on start (even without MSAA). Compute shaders / Direct compute is optional on feature level 10_0 and 10_1. (There is an example how to test for that in here somewhere: https://learn.microsoft.com/en-us/windows/win32/direct3d11/direct3d-11-advanced-stages-compute-create). I think Malkierian wanted to do *some* check against Direct Compute support and automatically switch over to OpenGL.

I don't have any old hardware to test and poke around at stuff. 

Tldr:
* SoH still crashes at start if there is no Direct Compute support on your Hardware. 
* This PR probably prevents MSAA crashes on hardware with only feature level 10_0 support, by disabling MSAA - if your hardware supports direct compute in the first place.
* there is no change in the UI telling you the MSAA slider does nothing

Spodi:
Welp, I'm marking the Pr as ready. You decide what to do with it. Maybe someone else is patient enough to find if the port is even getting max MSAA from LUS for the MSAA slider or if this need some additional cvar or function. I'm open to suggestions.

Kenix3:
> Welp, I'm marking the Pr as ready. You decide what to do with it. Maybe someone else is patient enough to find if the port is even getting max MSAA from LUS for the MSAA slider or if this need some additional cvar or function. I'm open to suggestions.

It's hardcoded in Ship. 
https://github.com/HarbourMasters/Shipwright/blob/develop/soh/soh/SohMenuBar.cpp#L316
I don't think LUS exposes this, but the slider is always 1-8.

1 is in the slider, and it's always a valid value. This basically just means on older hardware the slider will be ignored. Maybe make a PR to SoH with a tooltip warning.

PR #390, opened 2023-12-03
Fix holes in color clamped effects

Qlonever:
Limits color clamping to RGB instead of RGBA; fixes a few bugged graphics in SoH.
Before:
image
After:
image


garrettjoecox (approved):

garrettjoecox:
I looked into this for a few hours a few days ago, thanks for this! Verified on OpenGL that saria's song is now fixed, and Ganondorf's magic effects also remain fixed.

MelonSpeedruns:
I approve of this change!

Kenix3 (approved):

PR #391, opened 2023-12-05
Revert "Fix potential loss of timer precision (#376)"

garrettjoecox:
Reproduction steps:
- Check out latest develop branch (on SoH)
- To prevent the performance issue I caused on File Select screen, wipe your shipofharkinian.json (if there is an active spoiler file it's parsing JSON every frame)
- Set up cmake project in release mode
- Build in release mode
- Navigate to file select screen
- Wait like 10 seconds into the file select sequence
- Observe static sound
- Revert #376 locally
- Repeat the above
- Observe lack of static sound

I won't pretend to understand what's going on here, but logging the results of the function before/after the change show huge differences in their output, maybe there is some code somewhere that relies on the precision loss? No idea.

[Ship_of_Harkinian.log](https://github.com/Kenix3/libultraship/files/13555333/Ship_of_Harkinian.log)

### DIFF
--- a/src/graphic/Fast3D/gfx_dxgi.cpp
+++ b/src/graphic/Fast3D/gfx_dxgi.cpp
@@ -593,13 +593,11 @@ static void gfx_dxgi_handle_events(void) {
 }
 
 static uint64_t qpc_to_ns(uint64_t qpc) {
-    qpc *= 1000000000;
-    return qpc / dxgi.qpc_freq;
+    return qpc / dxgi.qpc_freq * 1000000000 + qpc % dxgi.qpc_freq * 1000000000 / dxgi.qpc_freq;
 }
 
 static uint64_t qpc_to_100ns(uint64_t qpc) {
-    qpc *= 10000000;
-    return qpc / dxgi.qpc_freq;
+    return qpc / dxgi.qpc_freq * 10000000 + qpc % dxgi.qpc_freq * 10000000 / dxgi.qpc_freq;
 }
 
 static bool gfx_dxgi_start_frame(void) {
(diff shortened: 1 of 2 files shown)


Spodi:
At first glance I thought miss counted the digits, but that doesn't seem to be the case. There must be something off with my math? I didn't expect it to be an order of magnitude sort of difference! Everything seemed okay in my testings (but as we all know by now that seems to mean nothing). I will investigate, hold on

Spodi:
I can't even reproduce the issue in the first place. No static or crackling sound, when building the latest dev locally or downloading it from github (this was latest at this point: https://github.com/HarbourMasters/Shipwright/commit/2d22a3ebcc9d735eeccc4df6eb6e78a4a31cfb91).

Also did this
```cpp
static uint64_t qpc_to_100ns(uint64_t qpc) {
    uint64_t oldresult = qpc / dxgi.qpc_freq * 10000000 + qpc % dxgi.qpc_freq * 10000000 / dxgi.qpc_freq;
    SPDLOG_INFO("100ns old: {}", oldresult);
    qpc *= 10000000;
    uint64_t newresult = qpc / dxgi.qpc_freq;
    SPDLOG_INFO("100ns new: {}", newresult);
    return newresult;
}
```
with that result:
[Ship_of_Harkinian.log](https://github.com/Kenix3/libultraship/files/13558736/Ship_of_Harkinian.log)

Spodi:
can you add some logging for what th qpc und dxgi.gpc_freq variables are for you in that function too? cause no matter what I try, I can't reproduce it

garrettjoecox:
Fixed by #392

PR #392, opened 2023-12-05
Fix busy loop for dxgi framepacing

Spodi:
Fix for busy loop for dxgi framepacing. Negates #391, and also fixes frame timing and skipping with FPS settings higher than monitor's refresh rate for some people.


garrettjoecox (approved):
Fixed the issue for me

PR #400, opened 2023-12-19
fix saving mappings for additional (port provided) bitmasks

briaguya0:
soh PR to get builds/test https://github.com/HarbourMasters/Shipwright/pull/3714


Kenix3 (approved):
Looks like an improvement over what we have now. This code will need to be revisted as we work towards 2.0, though.

PR #403, opened 2024-01-02
Replace mVariables map by unordered map to optimize get functions

jeremy010203:
mVariables is used to store config values as a simple dict.
There is no need to use a map as we don't need to keep track of insertion order of keys.
I propose to replace the map by unordered map.

On soh the getinteger function seems to be 2x faster (1,8% -> 0,9% of cpu usage):

Screenshot 2024-01-02 at 17 04 30

Screenshot 2024-01-02 at 17 04 04


Kenix3 (approved):
Good catch. I think this one got missed when we did the map -> unordered_map conversion.

PR #409, opened 2024-01-05
Expose tex clear funcs and fix more Metal issues with blended textures (cherry-pick)

Archez:
This exposes the texture cache clearing func so that it can be used outside of draw commands. I also added and exposed an unregister blended texture effect so that they can be cleaned up when not needed.

Additionally I resolved another issue with Metal shaders not working properly with blended effects, where the "last bound" textures on the frame buffer specifically for the mask and blended texture IDs were not cleared compared to DirectX.

Used with https://github.com/HarbourMasters/Shipwright/pull/3773

(cherry-picked from https://github.com/Kenix3/libultraship/commit/96c8a8929c18c1bffd7d92a35a589f74cf16fc59)


Kenix3:
Thankyou for the cherry pick!

PR #410, opened 2024-01-06
discovery: remove libultra headers

briaguya0:


Kenix3:
Headers should be able to be easily removed once the libultra implementations are moved elsewhere. Closing for now.

PR #413, opened 2024-01-14
Restore osGetTime/osGetCount to match n64 specs

Archez:
`osGetTime`/`osGetCount` where returning the values based on modern machine cycle rates at the nanosecond level.
The specifications for N64 report these methods as running at 48.675MHz for the CPU clock cycle.

I've added a set of ratio and custom chrono duration values to allow us to return accurate values from these methods for ports that depend on them for things like in-game timers.

The only detail I didn't implement yet is that the main difference between `osGetTime` and `osGetCount`, is that `osGetTime` specifically returns the time since last "boot". I'm not sure how we would want to handle that detail, but for now its not immediately needed.

I am also unsure if LUS should still provide a "modern" osGetTime that returns normalized seconds/miliseconds values.

### DIFF
--- a/src/public/libultra/os.cpp
+++ b/src/public/libultra/os.cpp
@@ -1,5 +1,11 @@
 #include "libultraship/libultraship.h"
 #include <SDL2/SDL.h>
+#include <ratio>
+
+// Establish a chrono duration for the N64 46.875MHz clock rate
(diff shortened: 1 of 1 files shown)


Kenix3:
I think the answer is that osGetTime should work authentically, but LUS should have a modern timer class with C bridges for real time timers. This PR works for getting stuff going in MM.

PR #415, opened 2024-01-29
Fix OpenGL_ES support on GNU/Linux.

vanfanel:
This fixes GLES support on GNU/Linux.
What this does is propagate `-DUSE_OPENGLES=1` when it's passed to SOH build system, to libultraship and imGui.
Also, it deals with the fact that `-lGLESv3` does not exist on GNU/Linux, where GLES3 programs are built with `-lGLESv2` because GLES3 is considered simply a GLES2 extension.

Corresponding SOH PR is: https://github.com/HarbourMasters/Shipwright/pull/3885


src/CMakeLists.txt
```diff
@@ -473,8 +473,13 @@ if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
...
 
 if (USE_OPENGLES)
-    find_library(OPENGLES_LIBRARY GLESv3)
-    target_link_libraries(libultraship PRIVATE GLESv3)
+    if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
+        find_library(OPENGLES_LIBRARY GLESv2)
+        target_link_libraries(libultraship PRIVATE GLESv2)
+    else()
+        find_library(OPENGLES_LIBRARY GLESv3)
+        target_link_libraries(libultraship PRIVATE GLESv3)
```
briaguya0: in the PR description you mention

> Also, it deals with the fact that -lGLESv3 does not exist on GNU/Linux, where GLES3 programs are built with -lGLESv2 because GLES3 is considered simply a GLES2 extension.

if that is the case, then why is GLESv3 still being used outside of linux here?

briaguya0: ah, GLESv3 is an android (only?) thing, got it

Kenix3 (approved):

PR #419, opened 2024-01-31
Remove linking against unused legacy OpenGL GLX implementation.

vanfanel:
Originally, f3d used GLX-specific code for context initialization.

LUS has moved away from that, and expects SDL to be used for abstraction (we aren't doing any X specific stuff that would require us to include GLX).

Based on this (and local tests on GNU/Linux systems with Xorg/Wayland and KMS/DRM only) it's safe to simply remove `${OPENGL_glx_LIBRARY}` from that line.


briaguya0:
it looks like this dependency was inherited from f3d

https://github.com/Kenix3/libultraship/blob/0833afad66e96d2ec4bbc410186d7247dc243ee2/src/graphic/Fast3D/README.md?plain=1#L7

i did a full search for GLX in the SoH codebase and i did not see it being used anywhere

soh testing PR https://github.com/HarbourMasters/Shipwright/pull/3892

briaguya0 (approved):
the appimage from https://github.com/HarbourMasters/Shipwright/pull/3892 is working for me

:shipit:

PR #420, opened 2024-02-01
Remove linking against unused GLEW on GNU/Linux.

vanfanel:
GLEW does not seem to be used on GNU/Linux at all, so linking against it seems unnecessary: not linking against it has zero consequences, the engine builds and works without problems.

Not depending on GLEW makes SOH/LUS more easily portable to light GNU/Linux based desktop-less systems (building Xorg-less GLEW is possible but problematic), so this is all benefits.


briaguya0:
what about
https://github.com/Kenix3/libultraship/blob/d42849d9cddbf610440d6488dc6546b6697578ff/src/graphic/Fast3D/gfx_opengl.cpp#L42

vanfanel:
@briaguya-ai Yes, GLEW headers are apparently used instead of using SDL2 OpenGL headers.

We don't need to depend on GLEW just for headers that SDL2 already provides.

That's corrected in https://github.com/Kenix3/libultraship/pull/420/commits/fdf95bdb5324cfbfb0d5326fac54be29b2986de7

vanfanel:
@briaguya-ai Now everything builds without depending on GLEW headers.

extern/CMakeLists.txt
```diff
@@ -109,21 +109,18 @@ else()
 endif()
 
 if (CMAKE_SYSTEM_NAME STREQUAL "Windows")
-    find_package(GLEW REQUIRED)
-    target_link_libraries(ImGui PUBLIC opengl32 GLEW::GLEW)
```
briaguya0: why are you removing this on windows?

extern/CMakeLists.txt
```diff
@@ -109,21 +109,18 @@ else()
...
+    target_link_libraries(ImGui PUBLIC opengl32)
 elseif (CMAKE_SYSTEM_NAME STREQUAL "Darwin")
-    find_package(GLEW REQUIRED)
-    target_link_libraries(ImGui PUBLIC ${OPENGL_opengl_LIBRARY} GLEW::GLEW)
```
briaguya0: why are you removing this for mac?

src/graphic/Fast3D/gfx_opengl.cpp
```diff
@@ -847,10 +846,6 @@ static void gfx_opengl_draw_triangles(float buf_vbo[], size_t buf_vbo_len, size_
 }
 
 static void gfx_opengl_init(void) {
-#if !defined(__SWITCH__) && !defined(USE_OPENGLES)
-    glewInit();
-#endif
```
briaguya0: why are you fully removing this? just change it to `!defined(__SWITCH__) && !defined(__linux__)`

briaguya0 (commented):
you're changing too much here. keep it linux only for now. if you want to make a follow up PR removing GLEW for all platforms go for it, but this looks like a lot of copy-pasting with zero testing

extern/CMakeLists.txt
vanfanel: Because it's not required anymore.
`SDL2_opengl.h` has the functions previously provided by the GLEW headers.

Maybe I should rename the PR to remove the GNU/Linux part?

vanfanel: @briaguya-ai Ok, I understand what you mean. Will remove the GNU/Linux GLEW dependency only for now, and do a follow up PR for the rest of the platforms.

PR #421, opened 2024-02-02
Remove detection of unused PulseAudio audio libary.

vanfanel:
On GNU/Linux, SDL2 audio is now used instead of PulseAudio, which abstracts the underlying audio system as intended (on GNU/Linux it can ALSA, PipeWire, PulseAudio, etc. and there's no need to access them directly since SDL2 makes a perfect job abstracting that).

PulseAudio is not used anywhere in the code, so trying to detect it is unnecessary.
This makes the PulseAudio dependency finally go away, since this detection at build time is all that remains.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -465,11 +465,6 @@ if (CMAKE_SYSTEM_NAME STREQUAL "Darwin")
     target_link_libraries(libultraship PRIVATE ${OSX_FOUNDATION} ${OSX_AVFOUNDATION} ${METAL} ${QUARTZCORE} ${CMAKE_DL_LIBS})
 endif()
 
-if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
-    find_package(PulseAudio)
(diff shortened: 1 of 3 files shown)


briaguya0:
i think there is more cleanup to be done here. https://github.com/search?q=repo%3AKenix3%2Flibultraship+pulse&type=code

we're already migrating anyone with pulse in their config over to sdl

https://github.com/Kenix3/libultraship/blob/d42849d9cddbf610440d6488dc6546b6697578ff/src/config/Config.cpp#L181-L185

but we still have pulse as an available backend

https://github.com/Kenix3/libultraship/blob/d42849d9cddbf610440d6488dc6546b6697578ff/src/audio/Audio.h#L9

https://github.com/Kenix3/libultraship/blob/d42849d9cddbf610440d6488dc6546b6697578ff/src/audio/Audio.cpp#L40-L42

vanfanel:
@briaguya-ai Removed the remaining PulseAudio backend references from the code in https://github.com/Kenix3/libultraship/pull/421/commits/ff4f9be70f751b50e16eb744212dc2a616b6c7c2

PR #424, opened 2024-02-06
get in game

briaguya0:
https://github.com/HarbourMasters/Shipwright/pull/3912
https://github.com/HarbourMasters/OTRExporter/pull/14

### DIFF
--- a/src/resource/archive/OtrArchive.cpp
+++ b/src/resource/archive/OtrArchive.cpp
@@ -28,8 +28,7 @@ std::shared_ptr<File> OtrArchive::LoadFileRaw(const std::string& filePath) {
     fileToLoad->InitData->Path = filePath;
     DWORD fileSize = SFileGetFileSize(fileHandle, 0);
     DWORD readBytes;
-    fileToLoad->Buffer = std::make_shared<std::vector<char>>();
-    fileToLoad->Buffer->resize(fileSize);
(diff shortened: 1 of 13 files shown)


src/resource/archive/OtrArchive.cpp
```diff
@@ -24,9 +24,11 @@ std::shared_ptr<File> OtrArchive::LoadFileRaw(const std::string& filePath) {
     }
 
     auto fileToLoad = std::make_shared<File>();
+    fileToLoad->InitData = std::make_shared<ResourceInitData>();
```
Kenix3: Instead of trying to prevent crashes on a File that should not have a resource init data, we should probably move path from ResourceInitData to File.

src/resource/archive/OtrArchive.cpp
```diff
@@ -24,9 +24,11 @@ std::shared_ptr<File> OtrArchive::LoadFileRaw(const std::string& filePath) {
...
     fileToLoad->InitData->Path = filePath;
     DWORD fileSize = SFileGetFileSize(fileHandle, 0);
     DWORD readBytes;
+    fileToLoad->Buffer = std::make_shared<std::vector<char>>();
     fileToLoad->Buffer->resize(fileSize);
```
Kenix3:
```suggestion
    fileToLoad->Buffer = std::make_shared<std::vector<char>>(fileSize);
```

Kenix3 (changes requested):

PR #430, opened 2024-02-08
Remove the remains of GLEW usage on the rest of operating systems.

vanfanel:
This is a follow-up to removing the remains of the unused GLEW dependency on GNU/Linux:
https://github.com/Kenix3/libultraship/commit/bd20dd4470462420a55d9a3d999868c2110add5d

The situation on these remaining systems is the same it was on GNU/Linix: GLEW headers were used for providing GL functions that should be provided by the SDL2 headers instead, since SDL2 is already a dependency.

### DIFF
--- a/src/graphic/Fast3D/gfx_opengl.cpp
+++ b/src/graphic/Fast3D/gfx_opengl.cpp
@@ -19,18 +19,10 @@
 #define FOR_WINDOWS 0
 #endif
 
-#ifdef _MSC_VER
-#include <SDL2/SDL.h>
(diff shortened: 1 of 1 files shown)


briaguya0:
This needs to be tested. This cannot be merged until an SoH testing PR is made and people can verify that this doesn't break *anything* on Mac, Windows, and Switch.

vanfanel:
@briaguya-ai The Shipwright PR corresponding to this one is:
https://github.com/HarbourMasters/Shipwright/pull/3922

PR #446, opened 2024-02-14
Remove GLEW remains on Windows.

vanfanel:
This is a follow-up to removing the remains of the unused GLEW dependency on GNU/Linux:

https://github.com/Kenix3/libultraship/commit/bd20dd4470462420a55d9a3d999868c2110add5d

The situation on Windows is the same it was on GNU/Linux: GLEW headers were used for providing GL functions that should be provided by the SDL2 headers instead, since SDL2 is already a dependency.


vanfanel:
The corresponding SOH PR is:

https://github.com/HarbourMasters/Shipwright/pull/3936

Kenix3:
Going to close this since there's been no movement since Feb 14th. The SoH PR is failing to build still. I'm pretty sure GLEW is currently required on Windows. Feel free to make a new PR when you have more time to work on the issue.

PR #453, opened 2024-02-25
Restore window before closing

Spodi:
This looks a bit funny, but prevents saving the size of a maximized or minimized window. The window is always restored to it's normal state before closing down (maximized and minimized state is NOT saved.)

The only thing that doesn't really work is when going fullscreen while the window is maximized and the game is closed while still in fullscreen. It still starts in fullscreen, but the maximized size is restored (without it being actually maximized) when going windowed.

Also quitting should always call gfx_dxgi_close() or gfx_dxgi_close() now, no matter how you quit (except outright killing the process I guess).

Test-PR: https://github.com/HarbourMasters/Shipwright/pull/4395


Spodi:
There is one exception to gfx_dxgi_close(): I don't call it on WM_QUIT, cause any window should be already destroyed when this is called.

PR #457, opened 2024-03-04
Bring over work from 2ship

louist103:
Still a few things to work out before this is good to go.

- [ ] Testing with SoH
- [x] Investigate the ABI difference
- [x] Figure out why `ResourceManager::LoadResource` can't use `LoadResourceAsync`

### DIFF
--- a/include/libultraship/libultra/abi.h
+++ b/include/libultraship/libultra/abi.h
@@ -37,6 +37,7 @@ typedef unsigned int u32;
 #define A_INIT 0x01
 #define A_CONTINUE 0x00
 #define A_LOOP 0x02
+#define A_ADPCM_SHORT 0x04
 #define A_OUT 0x02
 #define A_LEFT 0x02
 #define A_RIGHT 0x00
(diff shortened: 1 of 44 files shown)


include/libultraship/libultra/abi.h
```diff
@@ -446,13 +447,20 @@ typedef short ENVMIX_STATE[40];
...
+        _a->words.w0 = (_SHIFTL(A_DUPLICATE, 24, 8) | _SHIFTL(numCopies, 16, 8) | _SHIFTL(dmemSrc, 0, 16)); \
+        _a->words.w1 = _SHIFTL(dmemDest, 16, 16) | _SHIFTL(0x80, 0, 16);                                    \
     }
+//#define aDuplicate(pkt, count, dmemi, dmemo, a4)                                                      \
+//    {                                                                                                 \
+//        Acmd* _a = (Acmd*)pkt;                                                                        \
+//                                                                                                      \
+//        _a->words.w0 = (_SHIFTL(A_DUPLICATE, 24, 8) | _SHIFTL(count, 16, 8) | _SHIFTL(dmemi, 0, 16)); \
+//        _a->words.w1 = _SHIFTL(dmemo, 16, 16) | _SHIFTL(a4, 0, 16);                                   \
+//    }
```
briaguya0: it's not clear why this is commented out instead of removed. if the idea is to keep it around as a reference it'd be nice to add a comment explaining that, otherwise just rip it out

include/libultraship/libultra/gbi.h
```diff
@@ -2271,9 +2284,9 @@ typedef union {
 #define gsSPLoadUcode(uc_start, uc_dstart) gsSPLoadUcodeEx((uc_start), (uc_dstart), SP_UCODE_DATA_SIZE)
 
 #define gSPLoadUcodeL(pkt, ucode) \
-    gSPLoadUcode((pkt), OS_K0_TO_PHYSICAL(&##ucode##TextStart), OS_K0_TO_PHYSICAL(&##ucode##DataStart))
+//    gSPLoadUcode((pkt), OS_K0_TO_PHYSICAL(&##ucode##TextStart), OS_K0_TO_PHYSICAL(&##ucode##DataStart))
```
briaguya0: same as the comment about `aDuplicate` in abi. if there's a reason to keep it around and commented let's add a comment providing context. otherwise just rip it out

include/libultraship/libultra/gbi.h
```diff
@@ -2271,9 +2284,9 @@ typedef union {
...
 #define gsSPLoadUcodeL(ucode) \
-    gsSPLoadUcode(OS_K0_TO_PHYSICAL(&##ucode##TextStart), OS_K0_TO_PHYSICAL(&##ucode##DataStart))
+//    gsSPLoadUcode(OS_K0_TO_PHYSICAL(&##ucode##TextStart), OS_K0_TO_PHYSICAL(&##ucode##DataStart))
```
briaguya0: another "provide context or rip it out" example

src/CMakeLists.txt
```diff
@@ -1,8 +1,12 @@
 add_library(libultraship STATIC)
 set_target_properties(libultraship PROPERTIES PREFIX "")
 set_property(TARGET libultraship PROPERTY CXX_STANDARD 20)
+# Need to set C11 for libgfxd (_Alignas)
+set_property(TARGET libultraship PROPERTY C_STANDARD 11)
```
briaguya0, line 5: *newer* versions don't work?

louist103: I don't think it won't work, but the default didn't.

Archez: I removed this line and didn't have issues building, so it could probably be dropped or maybe match our `CXX` version.
`alignas` has been in c standard since C11. 

Edit: NVM, it built fine for me cause OTRExporter/ZAPDTR also set the same property, so it still ends up in the cmake cache

include/libultraship/libultra/gbi.h
louist103: Old classic "It didn't work before"

PR #458, opened 2024-03-05
Fix: opengles cmake check

Archez:
In the cmake list `option(USE_OPENGLES "Enable GLES3" OFF)` is used to define and default the value of `USE_OPENGLES` in the cmake lists.

Then later a check for `if(DEFINED USE_OPENGLES)` was being used. This always evaluates to true because the value is defined even if its "OFF", which was leading to opengles being used for all linux platforms.

This PR fixes that by actually checking the value of the variable.


Kenix3 (approved):

briaguya0 (approved):
:tada:

PR #473, opened 2024-04-07
Safely check ucode handlers for out of range or missing opcode

Archez:
The execution of the ucode handlers would crash if a bad ucode index was loaded, or the opcode has no handler.

This first adds a bounds check for the index being within the array size of ucode_handlers, and if it isn't, trip a debug `assert` as the call to `G_LOAD_UCODE` was incorrect. Alternatively, we could move the `assert` to be inside `gfx_set_ucode_handler`.

The second is to add a `contains` check for the opcode on the specific ucode handler before accessing the map. If the opcode is not recognized, it silently skips over it. I'm not sure if we would also want a debug `assert` here, or maybe logging, thoughts?

### DIFF
--- a/src/graphic/Fast3D/gfx_pc.cpp
+++ b/src/graphic/Fast3D/gfx_pc.cpp
@@ -3687,8 +3687,15 @@ static void gfx_step() {
         if (rdpHandlers.at(opcode)(&cmd)) {
             return;
         }
-    } else if (ucode_handlers[ucode_handler_index]->at(opcode)(&cmd)) {
-        return;
+    } else if (ucode_handler_index < ucode_handlers.size()) {
+        if (ucode_handlers[ucode_handler_index]->contains(opcode)) {
+            if (ucode_handlers[ucode_handler_index]->at(opcode)(&cmd)) {
+                return;
+            }
+        }
+    } else {
+        // Loaded ucode is out of range of the supported ucode_handlers
+        assert(true);
     }
 
     ++cmd;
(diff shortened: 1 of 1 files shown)


Kenix3:
Definitely a situation we want to log it out when the code is not within the map

Archez:
> Definitely a situation we want to log it out when the code is not within the map

What log level should we use, `WARN`, `DEBUG`?

Kenix3:
> > Definitely a situation we want to log it out when the code is not within the map
> 
> What log level should we use, `WARN`, `DEBUG`?

I'm thinking WARN.

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -3687,8 +3687,15 @@ static void gfx_step() {
...
-        return;
+    } else if (ucode_handler_index < ucode_handlers.size()) {
+        if (ucode_handlers[ucode_handler_index]->contains(opcode)) {
+            if (ucode_handlers[ucode_handler_index]->at(opcode)(&cmd)) {
+                return;
+            }
+        }
+    } else {
+        // Loaded ucode is out of range of the supported ucode_handlers
+        assert(true);
```
louist103: IMO this should be in the handler for changing ucode. Also should be `assert(false)`

Archez: Updated

louist103 (approved):

PR #492, opened 2024-04-19
Implemented dither noise

KiritoDv:
(cherry picked from commit 7321c1a54903d676b02e0a8bcddd48a78935d40e)

### DIFF
--- a/src/graphic/Fast3D/gfx_opengl.cpp
+++ b/src/graphic/Fast3D/gfx_opengl.cpp
@@ -754,7 +754,7 @@ static struct ShaderProgram* gfx_opengl_create_and_load_new_shader(uint64_t shad
         glUniform1i(sampler_location, 5);
     }
 
-    if(cc_features.opt_alpha_cvg_sel || cc_features.opt_alpha_threshold) {
+    if(cc_features.opt_alpha_threshold) {
         prg->alpha_test_val_location = glGetUniformLocation(shader_program, "alphaTestValue");
         prg->used_alpha_threshold = true;
     } else {
--- a/src/graphic/Fast3D/gfx_opengl.cpp
+++ b/src/graphic/Fast3D/gfx_opengl.cpp
@@ -418,6 +418,22 @@ static struct ShaderProgram* gfx_opengl_create_and_load_new_shader(uint64_t shad
     append_line(fs_buf, &fs_len, "uniform int frame_count;");
     append_line(fs_buf, &fs_len, "uniform float noise_scale;");
 
+    append_line(fs_buf, &fs_len, "vec3 colorDither(float _noise_val, vec3 _color)");
+    append_line(fs_buf, &fs_len, "{");
+    append_line(fs_buf, &fs_len, "    vec3 _noise = vec3(_noise_val, _noise_val, _noise_val);");
+    append_line(fs_buf, &fs_len, "    vec3 threshold = 7.0 / 255.0 * (_noise - 0.5);");
+    append_line(fs_buf, &fs_len, "    _color = clamp(_color + threshold, 0.0, 1.0);");
+    append_line(fs_buf, &fs_len, "    _color.rgb = round(_color.rgb * 32.0) / 32.0;");
+    append_line(fs_buf, &fs_len, "    return _color;");
+    append_line(fs_buf, &fs_len, "}");
+    append_line(fs_buf, &fs_len, "float alphaDither(float _noise, float _alpha)");
+    append_line(fs_buf, &fs_len, "{");
+    append_line(fs_buf, &fs_len, "    float threshold = 7.0 / 255.0 * (_noise - 0.5);");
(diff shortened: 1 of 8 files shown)


src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -3283,11 +3286,15 @@ bool gfx_set_fb_handler_custom(Gfx** cmd0) {
     return false;
 }
 
+float gfx_calculate_noise_scale() {
+    return ((float)gfx_current_dimensions.height / SCREEN_HEIGHT) * 0.5f;
```
Archez: `SCREEN_HEIGHT` -> `gfx_native_dimensions.height`

src/graphic/Fast3D/gfx_opengl.cpp
```diff
@@ -566,8 +586,12 @@ static struct ShaderProgram* gfx_opengl_create_and_load_new_shader(uint64_t shad
     }
 
     if (cc_features.opt_alpha) {
+        append_line(fs_buf, &fs_len, "float cvg = 1.0;");
+        append_line(fs_buf, &fs_len, "float alphaValue = texel.a;");
+
         if (cc_features.opt_alpha_threshold) {
-            append_line(fs_buf, &fs_len, "if (texel.a < 8.0 / 256.0) discard;");
+            append_line(fs_buf, &fs_len, "alphaValue = clamp(alphaValue, 0.0, 1.0);");
+            append_line(fs_buf, &fs_len, "if (alphaValue < alphaTestValue) discard;");
```
Archez, line 604: > Fragment shader compilation failed
ERROR: 0:57: Use of undeclared identifier 'alphaTestValue'

src/graphic/Fast3D/gfx_opengl.cpp
```diff
@@ -566,8 +586,12 @@ static struct ShaderProgram* gfx_opengl_create_and_load_new_shader(uint64_t shad
     }
 
     if (cc_features.opt_alpha) {
+        append_line(fs_buf, &fs_len, "float cvg = 1.0;");
```
Archez: `cvg` appears to be unused

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -3283,11 +3291,15 @@ bool gfx_set_fb_handler_custom(Gfx** cmd0) {
     return false;
 }
 
+float gfx_calculate_noise_scale() {
+    return ((float)gfx_current_game_window_viewport.height / SCREEN_HEIGHT) * 0.5f;
```
Archez: My previous comment meant for this to be
```suggestion
    return ((float)gfx_current_dimensions.height / gfx_native_dimensions.height) * 0.5f;
```

So that it retains what you are replacing from before (specially that we don't use `SCREEN_HEIGHT` here anymore)

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -3886,7 +3898,7 @@ void gfx_run(Gfx* commands, const std::unordered_map<Mtx*, MtxF>& mtx_replacemen
                                             !game_renders_to_framebuffer);
     gfx_rapi->start_frame();
     gfx_rapi->start_draw_to_framebuffer(game_renders_to_framebuffer ? game_framebuffer : 0,
-                                        (float)gfx_current_dimensions.height / gfx_native_dimensions.height);
+                                        gfx_calculate_noise_scale(), g_rdp.alpha_test_value);
```
Archez: Passing in `alpha_test_value` through `start_draw_to_framebuffer` means that the value is only read and set once at the very beginning of the frame.

But you implemented `gfx_dp_set_blend_color` which influences this value, but it'll never be updated to the shader this way.

src/graphic/Fast3D/gfx_opengl.cpp
```diff
@@ -554,9 +581,13 @@ static struct ShaderProgram* gfx_opengl_create_and_load_new_shader(uint64_t shad
...
+        append_line(
+            fs_buf, &fs_len,
+            "texel.a = alphaDither(random(vec3(floor(gl_FragCoord.xy * noise_scale), float(frame_count))), texel.a);");
+    } else if (cc_features.opt_noise) {
         append_line(fs_buf, &fs_len,
-                    "texel.a *= floor(clamp(random(vec3(floor(gl_FragCoord.xy * noise_scale), float(frame_count))) + "
-                    "texel.a, 0.0, 1.0));");
+                    "texel.rgb = colorDither(random(vec3(floor(gl_FragCoord.xy * noise_scale), float(frame_count))), "
+                    "texel.rgb);");
     }
```
Archez, line 591: At least from what I can see in MM, it is possible for aplha dithering and color dithering to be applied at the same time. This `if / if else` does not allow that currently.

We may even need to split `opt_noise` into two values, or something so that we can control which is used.

PR #493, opened 2024-04-19
Libultra implementations

KiritoDv:


Kenix3:
The strategy for handling libultra implementations will be changed in version 2.0.0 of this software. Going to close this for now until there is a good place to PR these.

src/controller/controldevice/controller/Controller.cpp
```diff
@@ -73,6 +75,17 @@ bool Controller::HasConfig() {
...
+bool Controller::IsConnected() {
+    const auto deck = Context::GetInstance()->GetControlDeck();
+    for (uint8_t lusIndex = Blue; lusIndex < Max; lusIndex++) {
+        if (this->HasMappingsForLUSDeviceIndex(static_cast<LUSDeviceIndex>(lusIndex))) {
+            return true;
+        }
+    }
+
+    return false;
+}
```
briaguya0: this will return true if any mapping exists for for a controller, even if the mapping is to a physical device that is disconnected

briaguya0 (changes requested):
the controller/controldeck side of this should be a separate PR. it is not part of the libultra implementation and it the logic in those changes is broken in this PR

Kenix3 (approved):

PR #494, opened 2024-04-19
New Debug Stats window as an overlay with right click options

MelonSpeedruns:


src/window/gui/GuiMenuBar.h
```diff
@@ -7,5 +7,6 @@ namespace LUS {
 class GuiMenuBar : public GuiElement {
   public:
     using GuiElement::GuiElement;
+    float MenuBarHeight;
```
Kenix3, line 10: I don't think we want to store this on the class. Can you remove this part and just hard code it where you currently use this?

PR #496, opened 2024-04-21
move json resource/factory to LUS

briaguya0:
verified working in soh https://github.com/HarbourMasters/Shipwright/pull/4063


Kenix3 (commented):
Why keep the name RawJson instead of calling it Json?

briaguya0:
yeah, `RawJson` isn't the greatest name, it's actually "nlohmann parsed json," i'm ok with just `Json` but if we want to indicate the nlohmann-ness of it in the name somehow that works for me too

Kenix3:
> yeah, `RawJson` isn't the greatest name, it's actually "nlohmann parsed json," i'm ok with just `Json` but if we want to indicate the nlohmann-ness of it in the name somehow that works for me too

Not sure we need to capture that it's nlohmann personally. Can you think of a reason why we would ever use two different parsing libraries at the same time?

briaguya0:
> Can you think of a reason why we would ever use two different parsing libraries at the same time?

i'd expect we wouldn't want to do that, only thing i can think of is if we wanted to move to a different json parsing library it'd be easier to do a deprecation process where we have `NLJson` and `NewParsingLibJson` around at the same time, then drop `NLJson`, but we could probably handle that with feature flags instead. i'll move forward with renaming it to `Json`

Kenix3:
> > Can you think of a reason why we would ever use two different parsing libraries at the same time?
> 
> i'd expect we wouldn't want to do that, only thing i can think of is if we wanted to move to a different json parsing library it'd be easier to do a deprecation process where we have `NLJson` and `NewParsingLibJson` around at the same time, then drop `NLJson`, but we could probably handle that with feature flags instead. i'll move forward with renaming it to `Json`

If the argument is for supporting new and old json libraries in a transition period, my suggest would be to call it `Json` and rename it to `JsonLegacy` when we need that.

Issue #313, opened 2023-06-10
gSPBranchList is broken and crashes any DL that uses it


RevoSucks:
Per a conversation had about it, "gSPBranchList should be the equivalent of gSPDisplayList(), followed by gSPEndDisplayList()". Libultraship's implementation of it can be found here, and it is not a correct implementation:

[](https://github.com/Kenix3/libultraship/blob/f081dace55d6f78052527495745437ee37833dbe/src/graphic/Fast3D/gfx_pc.cpp#L2645-L2667)

Instead of:

```
} else {
    cmd = (Gfx*)seg_addr(cmd->words.w1);
}
```

It should probably be something like:
```
} else {
  cmd++;
  uint64_t hash = ((uint64_t)cmd->words.w0 << 32) + cmd->words.w1;
  cmd = (Gfx*)ResourceGetDataByCrc(hash);
  --cmd; // increase after break
}
```

Thanks for the suggested fix, @Rozelette  .

briaguya0:
would this change break the way `gsSPBranchListOTRFilePath` is being used in https://github.com/HarbourMasters/Shipwright/pull/2962/files ?

Rozelette:
I do not think it should, @briaguya-ai . The erroneous case is `G_DL_OTR_HASH`, so, a `gSPBranchList` to a DL that is referenced to by a hash. `gSPBranchList` to DLs that are referenced to by a filename or segment address should not be affected as they should already work as intended. To be honest, though, `G_DL_OTR_HASH`, `G_DL`, `G_DL_OTR_FILEPATH` are all different opcodes and I do not know in which contexts they are used. As in, I do not know which the `gbi.h` macros use, and the ones the macros we override use, and the ones ZAPD uses.

PR #503, opened 2024-04-22
Fix BranchDL not assigning the new dl to the active command

Archez:
The `BRANCH_DL` code paths weren't reassigning the active command pointer with the loaded DL properly, causing infinite loops when a branch DL is encountered.

### DIFF
--- a/src/graphic/Fast3D/gfx_pc.cpp
+++ b/src/graphic/Fast3D/gfx_pc.cpp
@@ -2914,7 +2914,7 @@ bool gfx_dl_otr_filepath_handler_custom(Gfx** cmd0) {
         g_exec_stack.call(*cmd0, nDL);
     } else {
         if (nDL != nullptr) {
-            cmd = nDL;
+            (*cmd0) = nDL;
             g_exec_stack.branch(*cmd0);
             return true; // shortcut cmd increment
         } else {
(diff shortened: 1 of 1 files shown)


Kenix3:
Is this meant to fix https://github.com/Kenix3/libultraship/issues/313?

Archez:
> Is this meant to fix #313?

No, this is fixing an issue with the BRANCH_DL breaking after the Gfx Refactor Stage 1

I'm not sure if #313 is still valid, I don't fully understand what the original issue is.

Archez:
The issue mentions the use case specifically for the `OTR_HASH` variant of the DL handler, which as of `main` latest looks like:
```c++
bool gfx_dl_otr_hash_handler_custom(Gfx** cmd0) {
    Gfx* cmd = *cmd0;
    if (C0(16, 1) == 0) {
        // Push return address
        (*cmd0)++;

        uint64_t hash = ((uint64_t)(*cmd0)->words.w0 << 32) + (*cmd0)->words.w1;

        Gfx* gfx = (Gfx*)ResourceGetDataByCrc(hash);

        if (gfx != 0) {
            g_exec_stack.call(cmd, gfx);
        }
    } else {
        assert(0 && "????");
        (*cmd0) = (Gfx*)seg_addr((*cmd0)->words.w1);
        return true;
    }
    return false;
}
```

Which appears to be unhandled in the branching case (note the `assert` here)

Kenix3 (approved):

PR #505, opened 2024-04-23
update `GameOverlay` to work when no external fonts have been loaded

### LINKED ISSUE #481: LUS should work without Fonts from the ResourceManager.
These are supplied in SoH by `soh.otr`. LUS should not require this data from an OTR. A fallback that is included with the code is reasonable here.

`GameOverlay::LoadFont`

briaguya0:
Resolves https://github.com/Kenix3/libultraship/issues/481

see https://github.com/HarbourMasters/Shipwright/pull/4067 to see how this can be used

### DIFF
--- a/src/window/gui/GameOverlay.cpp
+++ b/src/window/gui/GameOverlay.cpp
@@ -25,12 +25,12 @@ void GameOverlay::LoadFont(const std::string& name, const std::string& path, flo
     initData->ResourceVersion = 0;
     std::shared_ptr<Font> font = std::static_pointer_cast<Font>(
         Context::GetInstance()->GetResourceManager()->LoadResource(path, false, initData));
-    
+
     if (font == nullptr) {
         SPDLOG_ERROR("Failed to load font: {}", name);
-        return;   
+        return;
     }
-        
+
     mFonts[name] = io.Fonts->AddFontFromMemoryTTF(font->Data, font->DataSize, fontSize);
 }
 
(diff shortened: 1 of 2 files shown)


Archez:
Do we want to consider using one of the fonts baked into LUS/ImGui as the default font for the overlay? Currently with these changes, a port has to provide and load the font externally to be able to leverage the overlay system.

briaguya0:
> Do we want to consider using one of the fonts baked into LUS/ImGui as the default font for the overlay?

it uses the default one if we don't set anything

Archez:
> > Do we want to consider using one of the fonts baked into LUS/ImGui as the default font for the overlay?
> 
> it uses the default one if we don't set anything

Ah ok, I must have missed that part, nvm then

briaguya0:
yeah, it worked using the default font when all i changed was

```diff
void GameOverlay::Init() {
    Context::GetInstance()->GetResourceManager()->GetResourceLoader()->RegisterResourceFactory(
        std::make_shared<ResourceFactoryBinaryFontV0>(), RESOURCE_FORMAT_BINARY, "Font",
        static_cast<uint32_t>(RESOURCE_TYPE_FONT), 0);
    LoadFont("Press Start 2P", "fonts/PressStart2P-Regular.ttf", 12.0f);
    LoadFont("Fipps", "fonts/Fipps-Regular.otf", 32.0f);
-   const std::string defaultFont = mFonts.begin()->first;
    if (!mFonts.empty()) {
+      const std::string defaultFont = mFonts.begin()->first;
        const std::string font = CVarGetString("gOverlayFont", defaultFont.c_str());
        for (auto& [name, _] : mFonts) {
            if (font.starts_with(name)) {
                mCurrentFont = name;
                break;
            }
            mCurrentFont = defaultFont;
        }
    }
}
```

so this really didn't need to touch the default part much

edit for clarification: i tested this by removing the fonts from soh.otr so those calls to loadfont weren't loading anything

PR #527, opened 2024-04-28
Add CMake functionality to allow a port to modify what LUS uses for CVar names

Malkierian:
This utilizes CMake variables and definitions. On LUS's side, the definitions default to what they currently are; no changes without input from a port. They use `set(CACHE)` to allow for being lower priority.

All a port has to do is implement a custom cmake file that defines the same variables using non-cache `set()` so that the variable usage later finds them before looking in the cache for LUS's definitions.

EDIT: As a side note, only the prefixes for `gControllers` and `gAdvancedResolution` are configurable. Their sub-variables remain fixed in LUS.


CMakeLists.txt
```diff
@@ -9,6 +9,9 @@ if(CMAKE_SYSTEM_NAME STREQUAL "iOS")
 endif()
 
 project(libultraship LANGUAGES C CXX)
+
+include(cmake/cvars.cmake)
```
briaguya0: i'm curious as to what happens if this is included in the src cmakelists instead of the root one, it seems like something that only applies in there

cmake/cvars.cmake
```diff
@@ -0,0 +1,50 @@
+set(CVAR_VSYNC_ENABLED "gVsyncEnabled" CACHE STRING "")
+set(CVAR_Z_FIGHTING_MODE "gZFightingMode" CACHE STRING "")
+set(CVAR_NEW_FILE_DROPPED "gNewFileDropped" CACHE STRING "")
+set(CVAR_DROPPED_FILE "gDroppedFile" CACHE STRING "")
```
briaguya0, line 4: not part of this PR, just a note that we should make an issue to not need to use cvars for this

cmake/cvars.cmake
```diff
@@ -0,0 +1,50 @@
...
+set(CVAR_INTERNAL_RESOLUTION "gInternalResolution" CACHE STRING "")
+set(CVAR_MSAA_VALUE "gMSAAValue" CACHE STRING "")
+set(CVAR_SDL_WINDOWED_FULLSCREEN "gSdlWindowedFullscreen" CACHE STRING "")
+set(CVAR_TEXTURE_FILTER "gTextureFilter" CACHE STRING "")
+set(CVAR_CONTROL_NAV "gControlNav" CACHE STRING "")
```
briaguya0: we should probably make this

```suggestion
set(CVAR_IMGUI_CONTROLLER_NAV "gControlNav" CACHE STRING "")
```

or something else that implies it's an ImGui related thing

cmake/cvars.cmake
```diff
@@ -0,0 +1,50 @@
...
+set(CVAR_CONSOLE_ENABLED "gConsoleEnabled" CACHE STRING "")
+set(CVAR_CONTROLLER_CONFIGURATION_ENABLED "gControllerConfigurationEnabled" CACHE STRING "")
+set(CVAR_CONTROLLER_DISCONNECTED_ENABLED "gControllerDisconnectedWindowEnabled" CACHE STRING "")
+set(CVAR_CONTROLLER_REORDERING_ENABLED "gControllerReorderingWindowEnabled" CACHE STRING "")
+set(CVAR_GFX_DEBUGGER_ENABLED "gGfxDebuggerEnabled" CACHE STRING "")
+set(CVAR_STATS_ENABLED "gStatsEnabled" CACHE STRING "")
```
briaguya0:
```suggestion
set(CVAR_CONSOLE_WINDOW_OPEN "gConsoleEnabled" CACHE STRING "")
set(CVAR_CONTROLLER_CONFIGURATION_WINDOW_OPEN "gControllerConfigurationEnabled" CACHE STRING "")
set(CVAR_CONTROLLER_DISCONNECTED_WINDOW_OPEN "gControllerDisconnectedWindowEnabled" CACHE STRING "")
set(CVAR_CONTROLLER_REORDERING_WINDOW_OPEN "gControllerReorderingWindowEnabled" CACHE STRING "")
set(CVAR_GFX_DEBUGGER_WINDOW_OPEN "gGfxDebuggerEnabled" CACHE STRING "")
set(CVAR_STATS_WINDOW_OPEN "gStatsEnabled" CACHE STRING "")
```

i'm open to suggestions here but i think this better describes what these cvars actually do

cmake/cvars.cmake
```diff
@@ -0,0 +1,50 @@
...
+set(CVAR_ENABLE_MULTI_VIEWPORTS "gEnableMultiViewports" CACHE STRING "")
+set(CVAR_LOW_RES_MODE "gLowResMode" CACHE STRING "")
+set(CVAR_SIMULATED_INPUT_LAG "gSimulatedInputLag" CACHE STRING "")
+set(CVAR_ALT_ASSETS "gAltAssets" CACHE STRING "")
+set(CVAR_SINK_ENABLED "gSinkEnabled" CACHE STRING "")
```
briaguya0: i was hoping to see if i could think of a more descriptive name for this but i have no idea what it does, it seems to only be used here

https://github.com/Kenix3/libultraship/blob/a245aaf975f1f3c85ecd63ac3a9d3e8b44949eb8/src/window/gui/ConsoleWindow.cpp#L322-L339

cmake/cvars.cmake
```diff
@@ -0,0 +1,50 @@
...
+set(CVAR_GAME_OVERLAY_FONT "gOverlayFont" CACHE STRING "")
+set(CVAR_OPEN_MENU_BAR "gOpenMenuBar" CACHE STRING "")
```
briaguya0:
```suggestion
set(CVAR_MENU_BAR_OPEN "gOpenMenuBar" CACHE STRING "")
```

cmake/cvars.cmake
```diff
@@ -0,0 +1,50 @@
...
+set(CVAR_BLOCK_CONTROLLERS "gControllers" CACHE STRING "")
+set(CVAR_BLOCK_ADVANCED_RESOLUTION "gAdvancedResolution" CACHE STRING "")
```
briaguya0: using "block" made me think this was a cvar that would *block access* to these or something, i think

```suggestion
set(CVAR_PREFIX_CONTROLLERS "gControllers" CACHE STRING "")
set(CVAR_PREFIX_ADVANCED_RESOLUTION "gAdvancedResolution" CACHE STRING "")
```

might be less confusing

src/controller/controldeck/ControlDeck.cpp
```diff
@@ -9,6 +9,7 @@
 #endif
 #include <imgui.h>
 #include "controller/deviceindex/LUSDeviceIndexMappingManager.h"
+#include "window/gui/Gui.h"
```
briaguya0: now that the cvar macros are coming from cmake this change isn't needed

```suggestion
```

src/controller/controldevice/controller/Controller.h
```diff
@@ -18,6 +18,10 @@
 #include "controller/controldevice/ControlDevice.h"
 #include "controller/controldevice/controller/mapping/keyboard/KeyboardScancodes.h"
 
+#ifndef CVAR_SIMULATED_INPUT_LAG
+#define CVAR_SIMULATED_INPUT_LAG "gSimulatedInputLag"
+#endif
+
```
briaguya0: coming from cmake now

```suggestion
```

briaguya0 (commented):
Looking good! Left a couple naming suggestions and a question

cmake/cvars.cmake
briaguya0: so it seems this functionality was removed/broken at some point, i think for now the best thing to do is just not make a macro for `gSinkEnabled`, and i'll make an issue to figure out what's going on with the ability to send logs to the in game console

cmake/cvars.cmake
Malkierian: That was originally your name suggestion XD

CMakeLists.txt
Malkierian: This appears to work, but would I leave `cvars.cmake` where it is, or move it inside `src`?

briaguya0: i think leaving cvars.cmake where it is should be fine

briaguya0 (approved):

Kenix3 (approved):

cmake/cvars.cmake
briaguya0: https://github.com/Kenix3/libultraship/issues/535

PR #529, opened 2024-04-29
Added a validation to detect if non-uniform thread group sizes are supported

KiritoDv:
This PR Allows support for at least Apple with A10 processors, im doing tests on an iPad 6th Gen

### DIFF
--- a/src/graphic/Fast3D/gfx_metal.cpp
+++ b/src/graphic/Fast3D/gfx_metal.cpp
@@ -205,8 +205,8 @@ bool Metal_NonUniformThreadGroupSupported() {
     // iOS devices with A11 or later support dispatch threads
     return mctx.device->supportsFamily(MTL::GPUFamilyApple4);
 #else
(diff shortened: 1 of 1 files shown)


src/graphic/Fast3D/gfx_metal.cpp
```diff
@@ -199,6 +200,16 @@ bool Metal_IsSupported() {
 #endif
 }
 
+bool Metal_NonUniformThreadGroupSupported() {
+#ifdef __IOS__
+    // iOS devices with A11 or later support dispatch threads
+    return mctx.device->supportsFamily(MTL::GPUFamilyApple4);
```
briaguya0, line 206: would this not return true on macs?

KiritoDv: Not sure, because this validation is for mobile cpus, and this worked as if was true always on macs, so i prefer to just returning true instead of doing a syscall

Kenix3 (approved):

Issue #509, opened 2024-04-24
Clean up the `extern` directory


Kenix3:
We should not be relying on any external repositories committed into LUS.

Supercedes https://github.com/Kenix3/libultraship/issues/500

- [x] [dr_libs](https://github.com/mackron/dr_libs) ~(Fetch a commit from SHA)~ remove until we have a need for it in a `Ship::` resource
- [x] [stb](https://github.com/nothings/stb) (Fetch from a commit SHA) (blocked by https://github.com/Kenix3/libultraship/issues/485)
- [x] [thread-pool](https://github.com/bshoshany/thread-pool) (Fetch from tags)
- [x] [StormLib](https://github.com/ladislav-zezula/StormLib) (We 100% used to have changes, but I don't think we do anymore. Double check before converting.) (Might need to just fetch from tag) (blocked by: https://github.com/Kenix3/libultraship/issues/442)
- [x] [libgfxd](https://github.com/glankk/libgfxd) (Need to fetch from commit SHA) (may conflict with https://github.com/Kenix3/libultraship/issues/521)
- [x] [imgui](https://github.com/ocornut/imgui) (There are not packages for the docking branch, might need to fetch) (There may be some changes, but it is believed they are Switch specific) (blocked by https://github.com/Kenix3/libultraship/issues/485)
- [x] [nlohmann-json](https://github.com/nlohmann/json) (Has all packages we need)
- [x] [spdlog](https://github.com/gabime/spdlog) (Has all packages we need)
- [x] [tinyxml2](https://github.com/leethomason/tinyxml2) ~(Fetch from tags)~ looking into package manager
- [x] StrHash64 (Don't know what repo we took this from, if any. Can just absorb the code)
- [x] ZAPDUtils (https://github.com/Kenix3/libultraship/issues/484)
- [x] metal-cpp (We don't know where this came from, but worst case it could be absorbed into the renderer)

briaguya0:
see https://github.com/Kenix3/libultraship/pull/374 for imgui stuff - i see 2 options for paths forward here

either

* fork imgui and apply our changes from extern there
* pull from sha on fork until https://github.com/Kenix3/libultraship/issues/485
* switch to pulling from upstream sha

or

* keep it in the extern dir until https://github.com/Kenix3/libultraship/issues/485
* switch to pulling from upstream sha

Archez:
[...]

PR #536, opened 2024-04-30
spdlog from package managers

briaguya0:
https://github.com/Kenix3/libultraship/issues/509


src/CMakeLists.txt
```diff
@@ -232,6 +235,7 @@ if (CMAKE_SYSTEM_NAME STREQUAL "Windows")
         _CONSOLE
         _CRT_SECURE_NO_WARNINGS
         ENABLE_DX11
+        NOMINMAX
```
briaguya0, line 238: we use `std::min` and `std::max` which doesn't work if we include `windef.h` (which is included by `windows.h`)

this ensures we can continue to use them regardless of where we include `windows.h` from

i have no idea why the way we were including SPDLOG was making this not a problem for us before, probably just luck of the include order

src/debug/CrashHandler.h
```diff
@@ -13,6 +13,10 @@
 #include <SDL.h>
 #endif
 
+#if _WIN32
+#include <windows.h>
```
briaguya0, line 17: we need this for the `CONTEXT` type which is used in `PrintRegisters` and `PrintStack`

it seems we were getting this from SPDLOG before somehow

src/graphic/Fast3D/gfx_direct3d11.cpp
```diff
@@ -854,8 +854,8 @@ static void gfx_d3d11_update_framebuffer_parameters(int fb_id, uint32_t width, u
     Framebuffer& fb = d3d.framebuffers[fb_id];
     TextureData& tex = d3d.textures[fb.texture_id];
 
-    width = max(width, 1U);
-    height = max(height, 1U);
+    width = ((width) > (1U) ? (width) : (1U));
+    height = ((width) > (1U) ? (width) : (1U));
```
briaguya0, line 858: this is the only place in LUS we used the windows `max` macro. this just replaces it with the implementation from `windef.h`

https://stackoverflow.com/a/30924806

Issue #483, opened 2024-04-19
Building LUS with StormLib should be optional.


Kenix3:
Dependent upon https://github.com/Kenix3/libultraship/issues/442
We should be able to build LUS without StormLib as a dependency at all.

PR #541, opened 2024-05-01
fetchcontent stormlib

briaguya0:
https://github.com/Kenix3/libultraship/issues/509

packages are decently out of date compared to the git repo and `find_package` isn't supported

if stormlib is updated to support `find_package` then this can be reworked, but for now this seems like the simplest way to get back to where we were


cmake/dependencies/common.cmake
```diff
@@ -30,3 +30,14 @@ target_sources(ImGui
 )
 
 target_include_directories(ImGui PUBLIC ${imgui_SOURCE_DIR} ${imgui_SOURCE_DIR}/backends PRIVATE ${SDL2_INCLUDE_DIRS})
+
+# ========= StormLib =============
+if(TRUE)
```
briaguya0, line 35: this will be replaced with an actual condition in https://github.com/Kenix3/libultraship/issues/483

briaguya0:
> image

PR #544, opened 2024-05-02
Splitted f3dex from f3dex2

KiritoDv:


src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -1347,10 +1347,12 @@ static void gfx_sp_tri1(uint8_t vtx1_idx, uint8_t vtx2_idx, uint8_t vtx3_idx, bo
             cross = -cross;
         }
 
+#ifdef F3DEX_GBI_2
```
Kenix3, line 1350: Shouldn't be using IFDEFs. Use the current active GBI.

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -1799,6 +1801,7 @@ static void gfx_calc_and_set_viewport(const Vp_t* viewport) {
 }
 
 static void gfx_sp_movemem_f3dex2(uint8_t index, uint8_t offset, const void* data) {
+#ifdef F3DEX_GBI_2
```
Kenix3, line 1804: There should not be any gbi specific ifdefs or checks inside of handlers. We should split this into two functions, one for f3dex2 and one for f3dex.

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -1826,6 +1830,13 @@ static void gfx_sp_movemem_f3d(uint8_t index, uint8_t offset, const void* data)
         case G_MV_VIEWPORT:
             gfx_calc_and_set_viewport((const Vp_t*)data);
             break;
+// This is a pain but it makes the define easier to read
```
Kenix3, line 1833: There should not be any gbi specific ifdefs or checks inside of handlers. We should split this into two functions, one for f3dex2 and one for f3dex.

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -3604,44 +3614,70 @@ const static std::unordered_map<uint32_t, GfxOpcodeHandlerFunc> otrHandlers = {
     { G_SETINTENSITY, gfx_set_intensity_handler_custom },            // G_SETINTENSITY (0x40)
 };
 
-const static std::unordered_map<uint32_t, GfxOpcodeHandlerFunc> f3dex2Handlers = {
+#ifdef F3DEX_GBI_2
```
Kenix3, line 3617: Why ifdef the handler array? We should be switch arrays based on the renderer state current GBI value.

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -3604,44 +3614,70 @@ const static std::unordered_map<uint32_t, GfxOpcodeHandlerFunc> otrHandlers = {
...
-#else
+    { G_TRI1, gfx_tri1_handler_f3dex2 },
+    { G_TRI2, gfx_tri2_handler_f3dex },
+    // Commands to implement
+    { G_SPNOOP, gfx_spnoop_command_handler_f3dex2 },
+    { G_RDPHALF_1, gfx_stubbed_command_handler },
+};
+#endif
+
+#if (defined(F3DEX_GBI) || defined(F3D_GBI)) && !defined(F3DEX_GBI_2)
```
Kenix3, line 3646: Why ifdef the handler array? We should be switch arrays based on the renderer state current GBI value.

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -3655,9 +3691,13 @@ const static std::unordered_map<uint32_t, GfxOpcodeHandlerFunc> s2dexHandlers =
     { G_RDPHALF_2, gfx_stubbed_command_handler },
 };
 
-const static std::array<const std::unordered_map<uint32_t, GfxOpcodeHandlerFunc>*, UcodeHandlers::ucode_max>
+const static std::array<const std::unordered_map<int8_t, GfxOpcodeHandlerFunc>*, UcodeHandlers::ucode_max>
     ucode_handlers = {
+#ifdef F3DEX_GBI_2
```
Kenix3, line 3696: I don't really think this array should really exist. We should define this when we initialize Fast3D as passed in arguments.

Kenix3 (changes requested):

PR #559, opened 2024-05-06
Added simple gfx trace system

KiritoDv:
This PR adds support for debugging display lists to detect corruption or invalid data

### DIFF
--- a/src/resource/factory/DisplayListFactory.cpp
+++ b/src/resource/factory/DisplayListFactory.cpp
@@ -174,7 +174,7 @@ std::shared_ptr<Ship::IResource> ResourceFactoryBinaryDisplayListV0::ReadResourc
 
         int8_t opcode = (int8_t)(command.words.w0 >> 24);
         bool isExpanded = opcode == G_SETTIMG_OTR_HASH || opcode == G_DL_OTR_HASH || opcode == G_VTX_OTR_HASH ||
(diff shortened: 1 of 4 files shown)


Kenix3:
Add some ifdefs around all of this code and it can get in.

src/resource/factory/DisplayListFactory.cpp
```diff
@@ -145,24 +145,29 @@ std::shared_ptr<Ship::IResource> ResourceFactoryBinaryDisplayListV0::ReadResourc
...
+    size_t idx = 0;
     while (true) {
         Gfx command;
         command.words.w0 = reader->ReadUInt32();
         command.words.w1 = reader->ReadUInt32();
 
-        displayList->Instructions.push_back(command);
-
-        uint8_t opcode = (uint8_t)(command.words.w0 >> 24);
+        int8_t opcode = (int8_t)(command.words.w0 >> 24);
```
briaguya0, line 175: this change from `uint8_t` to `int8_t` hiding in this PR feels a little scary to me. if using `uint8_t` here was a mistake i'd feel better having a separate PR fixing that so we can discuss that change, then this PR can focus solely on *adding* a feature.

KiritoDv: I think that was a mistake i did because i copy/paste some things

PR #565, opened 2024-05-08
Decouple the Window class and Fast3D

### LINKED ISSUE #508: `Window` abstraction
Most of the public methods that exist on Window now can be in the abstract base class. There is some WindowBackend related stuff that would be on the Fast3D Window subclass.

Kenix3:
Resolves #508

### DIFF
--- a/src/Context.cpp
+++ b/src/Context.cpp
@@ -6,6 +6,7 @@
 #include <spdlog/sinks/stdout_color_sinks.h>
 #include "install_config.h"
 #include "debug/GfxDebugger.h"
+#include "graphic/Fast3D/Fast3dWindow.h"
 
 #ifdef _WIN32
 #include <tchar.h>
@@ -24,9 +25,10 @@ std::shared_ptr<Context> Context::GetInstance() {
 
 Context::~Context() {
     SPDLOG_TRACE("destruct context");
+    GetWindow()->SaveWindowToConfig();
+
     // Explicitly destructing everything so that logging is done last.
     mAudio = nullptr;
-    GetWindow()->SaveWindowSizeToConfig(GetConfig());
     mWindow = nullptr;
     mConsole = nullptr;
     mCrashHandler = nullptr;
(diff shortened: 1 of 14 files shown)


src/Context.cpp
```diff
@@ -256,7 +258,7 @@ void Context::InitWindow(std::vector<std::shared_ptr<GuiWindow>> guiWindows) {
         return;
     }
 
-    mWindow = std::make_shared<Window>(guiWindows);
+    mWindow = std::make_shared<Fast::Fast3dWindow>(guiWindows);
```
briaguya0, line 261: this makes it so `Ship::Context` requires fast3d (if only a little bit)

this seems fine but it's not clear to me how we'd want to handle non-fast3d windows going forward

src/graphic/Fast3D/gfx_sdl2.cpp
```diff
@@ -414,6 +414,8 @@ static void gfx_sdl_init(const char* game_name, const char* gfx_api_name, bool s
 
 static void gfx_sdl_close(void) {
     is_running = false;
+    SDL_DestroyRenderer(renderer);
+    SDL_Quit();
```
briaguya0, line 418: would it make sense to just call `gfx_sdl_destroy` here? that way when the
```
// TODO: destroy _any_ resources used by SDL
```
is addressed we wouldn't need to do things in 2 places

src/window/Window.cpp
```diff
@@ -286,21 +69,11 @@ std::shared_ptr<std::vector<WindowBackend>> Window::GetAvailableWindowBackends()
 
 void Window::SetWindowBackend(WindowBackend backend) {
     mWindowBackend = backend;
-    Context::GetInstance()->GetConfig()->SetWindowBackend(GetWindowBackend());
-    Context::GetInstance()->GetConfig()->Save();
+    Ship::Context::GetInstance()->GetConfig()->SetWindowBackend(GetWindowBackend());
+    Ship::Context::GetInstance()->GetConfig()->Save();
```
briaguya0, line 73: not complaining, but we don't actually need `Ship::` here because `Window` is in the ship namespace right? (seeing this just made me do a double take and verify we were actually in the ship namespace here)

briaguya0 (approved):

PR #588, opened 2024-05-16
Clearer errors in ResourceLoader.cpp

MegaMech:
We might not need to output str and hex for type. Not really sure if both is helpful or not.

image
Note that the grammar issue in the screenshot has been fixed.


Kenix3 (approved):
Approved pending clang for at fixes

PR #605, opened 2024-05-26
Add geometry mode for always executing branches

garrettjoecox:
BRANCH_Z can optionally be used in combination with args to conditionally render it. This is used in Majora's Mask for culling scene geometry. This adds an override to always execute the branch.


Kenix3 (approved):

garrettjoecox:
I just changed this to a geometry mode instead of a cvar

PR #618, opened 2024-05-30
Alt Assets Variable

Malkierian:
Creates `altAssetsEnabled` on `ResourceManager`.
Sets `altAssetsEnabled` on `ResourceManager::Init()`.
Swaps all `CVarGetInteger` calls for `CVAR_ALT_ASSETS` to `altAssetsEnabled` references.


src/resource/ResourceManager.h
```diff
@@ -57,6 +59,7 @@ class ResourceManager {
                                                                                  bool loadExact = false);
 
   private:
+    bool altAssetsEnabled = false;
```
briaguya0: tidy won't like this name

```suggestion
    bool mAlternateAssetsEnabled = false;
```

src/resource/ResourceManager.cpp
```diff
@@ -23,6 +23,7 @@ void ResourceManager::Init(const std::vector<std::string>& otrFiles, const std::
     // the extra `- 1` is because we reserve an extra thread for spdlog
     size_t threadCount = std::max(1, (int32_t)(std::thread::hardware_concurrency() - reservedThreadCount - 1));
     mThreadPool = std::make_shared<BS::thread_pool>(threadCount);
+    mAltAssetsEnabled = CVarGetInteger(CVAR_ALT_ASSETS, 0);
```
Archez: Since you are making a dedicated `SetAltAssetsEnabled` method now, I think we could just drop `CVAR_ALT_ASSETS` entirely from the LUS repo and just have ports set the initial value on boot.

Malkierian: You mean remove that line and have it set from the port on init?

Archez: yea, I mean you could explicitly set to false for clarity, but I think it starts as false anyways.

Malkierian: I declared it with false in the header.

briaguya0 (approved):

PR #623, opened 2024-06-01
Fix My Broken Implementation of G_QUAD for f3dex That Never Worked

MegaMech:
The args for G_QUAD to render quadrangles are correct now.


briaguya0:
this PR has the same title as https://github.com/Kenix3/libultraship/pull/599

i assume this is fixing something, would you mind updating the title/description to reflect what this is fixing?

Kenix3:
Looks like it's fixing (or just optimizing) how the quad is generated for the rendering api.

MegaMech:
The arguments were not correct before and I was drawing too many triangles.

Sorry, it was difficult to see if it was drawing correctly because the lighting doesn't work in the scene. But it's good now.

PR #624, opened 2024-06-01
WindowBackend enum stuff

### LINKED ISSUE #576: Remove the WindowBackend enum from Ship code.

briaguya0:
doesn't actually remove the windowbackend enum but updates it to put us in a better spot going forward

resolves https://github.com/Kenix3/libultraship/issues/576

* soh testing pr https://github.com/HarbourMasters/Shipwright/pull/4199

### DIFF
--- a/src/window/Window.cpp
+++ b/src/window/Window.cpp
@@ -81,7 +81,8 @@ bool Window::IsAvailableWindowBackend(int32_t backendId) {
     }
 
     // Verify the backend is available
-    return std::find(mAvailableWindowBackends->begin(), mAvailableWindowBackends->end(), backend) != mAvailableWindowBackends->end();
+    return std::find(mAvailableWindowBackends->begin(), mAvailableWindowBackends->end(), backend) !=
+           mAvailableWindowBackends->end();
 }
 
 void Window::SetWindowBackend(WindowBackend backend) {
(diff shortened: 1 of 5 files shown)


src/window/Window.h
```diff
@@ -9,7 +9,11 @@
 #include "window/gui/Gui.h"
 
 namespace Ship {
-enum class WindowBackend { DX11, DX12, GLX_OPENGL, SDL_OPENGL, SDL_METAL, GX2, BACKEND_COUNT };
+enum class WindowBackend {
+    FAST3D_DXGI_DX11,  // todo: set value
+    FAST3D_SDL_OPENGL, // todo: set value
+    FAST3D_SDL_METAL,  // todo: set value
```
briaguya0: @Kenix3 hoping you have ideas for values here

src/window/Window.cpp
```diff
@@ -67,6 +67,24 @@ std::shared_ptr<std::vector<WindowBackend>> Window::GetAvailableWindowBackends()
...
+
+    // Verify the id is a valid backend enum value
+    switch (backend) {
+        case WindowBackend::FAST3D_DXGI_DX11:
+        case WindowBackend::FAST3D_SDL_OPENGL:
+        case WindowBackend::FAST3D_SDL_METAL:
+            break;
+        default:
+            return false;
+    }
```
briaguya0: we can't just check to see if the value is `>= 0` and `< max` if we're setting values that aren't sequential, we could probably do some fancy template stuff or some fancy macro stuff instead of just throwing the values into a switch here but all of those options feel a bit overengineered for a use case this simple

PR #629, opened 2024-06-02
Fix Flawed Segmented Addr Check

MegaMech:
Not sure if I missed some but essentially:

`(0x05000800 & 1) != true`

### DIFF
--- a/src/graphic/Fast3D/gfx_pc.cpp
+++ b/src/graphic/Fast3D/gfx_pc.cpp
@@ -2589,7 +2589,7 @@ static void gfx_s2dex_rect_copy(F3DuObjSprite* spr) {
 
 static inline void* seg_addr(uintptr_t w1) {
     // Segmented?
-    if (w1 & 1) {
+    if (IS_SEGMENTED(w1)) {
         uint32_t segNum = (uint32_t)(w1 >> 24);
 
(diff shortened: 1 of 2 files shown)


Kenix3:
This change needs to be checked against other ports as it might break them.

```
To give some history:

On N64, you had addresses that were direct pointers to RAM and addresses that were to segments. There were 16 segments, and the first byte of a four byte address would determine the segment number (00-0F). So for example, a segment address could be `0x02000040`, referencing offset `0x40` on segment `2`. On N64 it is easy to distinguish between segment addresses and regular RAM-addresses because RAM is mapped to `0x80000000` which is too high for segments. But on PC the whole address space is used, so a new solution had to be figured out.

Random figured that no pointers in memory should have odd numbers, that everything was at least 2-byte aligned. So if something isn't, the assumption is that it's a segment pointer and not an address into RAM. I guess under certain conditions those path strings do end up being on odd addresses, hence the need for the alignment macro.
```

MegaMech:
Don't merge this, it's not correct yet.

MegaMech:
It appears the issue is that I need to flip the endiannes of my n64 ptr:
`0x0000000000080005`

MegaMech:
Resolved this issue by parsing the data at runtime and converting n64 ptrs to x86.

PR #636, opened 2024-06-11
Fix CC shader argument order for DirectX

Archez:
Fixes incorrect argument ordering from https://github.com/Kenix3/libultraship/pull/628


MegaMech (approved):
Can confirm this fixes a DX crash on launch

Archez:
Windows build is failing due to incompatible build caches on the runners.

briaguya0 (approved):

PR #639, opened 2024-06-20
Modern menu support

Malkierian:
Adds `DrawContents` to GuiElement to allow for calling it from a generic GuiWindow. Implemented in Console, GfxDebugger, and Stats windows.

Adds forced cursor visibility functionality, allowing cursor to remain visible in fullscreen without menu/menubar open.

Adds generic menu functionality (akin to the menubar functionality), including `GetMenu` and `SetMenu`. Draws the menu at the same point as the menubar. Also implements `GetMenuOrMenubarVisible()` to allow for a supplied menu to also trigger gamepad capture or cursor visibility changing when toggled.

Note: this should only be looked at with merge scrutiny if the plan is still to try the menu in a release of 2ship before upstreaming.

### DIFF
--- a/src/graphic/Fast3D/Fast3dWindow.h
+++ b/src/graphic/Fast3D/Fast3dWindow.h
@@ -29,6 +29,8 @@ class Fast3dWindow : public Ship::Window {
     bool IsFullscreen() override;
     bool IsRunning() override;
     const char* GetKeyName(int32_t scancode) override;
+    bool ShouldForceCursorVisibility();
+    void SetForceCursorVisibility(bool visible);
 
     void InitWindowManager();
     void SetTargetFps(int32_t fps);
@@ -47,5 +49,6 @@ class Fast3dWindow : public Ship::Window {
   private:
     GfxRenderingAPI* mRenderingApi;
     GfxWindowManagerAPI* mWindowManagerApi;
+    bool mForceCursorVisibility;
 };
 } // namespace Fast
--- a/src/graphic/Fast3D/Fast3dWindow.h
+++ b/src/graphic/Fast3D/Fast3dWindow.h
@@ -29,8 +29,6 @@ class Fast3dWindow : public Ship::Window {
     bool IsFullscreen() override;
     bool IsRunning() override;
     const char* GetKeyName(int32_t scancode) override;
-    bool ShouldForceCursorVisibility();
-    void SetForceCursorVisibility(bool visible);
 
     void InitWindowManager();
     void SetTargetFps(int32_t fps);
@@ -49,6 +47,5 @@ class Fast3dWindow : public Ship::Window {
   private:
     GfxRenderingAPI* mRenderingApi;
     GfxWindowManagerAPI* mWindowManagerApi;
-    bool mForceCursorVisibility;
 };
 } // namespace Fast
--- a/src/window/gui/Gui.cpp
+++ b/src/window/gui/Gui.cpp
@@ -134,8 +134,7 @@ void Gui::Init(GuiWindowInitData windowImpl) {
         mImGuiIo->ConfigFlags |= ImGuiConfigFlags_ViewportsEnable;
     }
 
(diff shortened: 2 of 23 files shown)


src/graphic/Fast3D/Fast3dWindow.h
```diff
@@ -29,6 +29,8 @@ class Fast3dWindow : public Ship::Window {
     bool IsFullscreen() override;
     bool IsRunning() override;
     const char* GetKeyName(int32_t scancode) override;
+    bool ShouldForceCursorVisibility();
+    void SetForceCursorVisibility(bool visible);
```
briaguya0: why is the cursor visibility stuff in f3d window instead of base window?

src/window/gui/Gui.cpp
```diff
@@ -69,7 +69,8 @@ Gui::Gui(std::vector<std::shared_ptr<GuiWindow>> guiWindows) : mNeedsConsoleVari
     }
 
     if (GetGuiWindow("Input Editor") == nullptr) {
-        AddGuiWindow(std::make_shared<InputEditorWindow>(CVAR_CONTROLLER_CONFIGURATION_WINDOW_OPEN, "Input Editor"));
+        AddGuiWindow(std::make_shared<InputEditorWindow>(CVAR_CONTROLLER_CONFIGURATION_WINDOW_OPEN, "Input Editor",
+                                                         ImVec2(520, 600)));
```
briaguya0: the input editor window didn't have a hardcoded size before, does this make it super wide? also the window itself was titled "Controller Configuration" - does this make it "Input Editor" now?

src/window/gui/Gui.cpp
```diff
@@ -337,13 +340,16 @@ void Gui::DrawMenu() {
 
     ImGui::DockSpace(dockId, ImVec2(0.0f, 0.0f), ImGuiDockNodeFlags_None | ImGuiDockNodeFlags_NoDockingInCentralNode);
 
-    if (ImGui::IsKeyPressed(TOGGLE_BTN) ||
+    if (ImGui::IsKeyPressed(TOGGLE_BTN) || ImGui::IsKeyPressed(ImGuiKey_Escape) ||
         (ImGui::IsKeyPressed(TOGGLE_PAD_BTN) && CVarGetInteger(CVAR_IMGUI_CONTROLLER_NAV, 0))) {
-        GetMenuBar()->ToggleVisibility();
+        if (!ImGui::IsKeyPressed(ImGuiKey_Escape)) {
+            GetMenuBar()->ToggleVisibility();
+        }
```
briaguya0: it seems with this logic using the controller to toggle the menu will always enable the menu**bar** - is that intended?

briaguya0 (approved):
didn't give this the most thorough readthrough but overall this is looking great!

i left a couple non-blocking questions

src/window/gui/Gui.cpp
Malkierian: No, that wasn't intended, actually, but I'm not sure how to make that dynamic instead of just swapping it over.

src/graphic/Fast3D/Fast3dWindow.h
Malkierian: I think only because I didn't realize that some non-virtual functions still existed there.

src/window/gui/Gui.cpp
Malkierian: Yeah, considering the changes I made to how the name displays, that did change it. I think it then is tied to the name used for lookup here. Should I just change it to Controller Configuration here?

briaguya0: I prefer controller config as a name but if others have strong feelings I'm fine with any name

Malkierian: Actually, considering we do support keyboard (and I to want to support mouse eventually), perhaps Input Editor is a better window title?

src/window/gui/Gui.cpp
Malkierian: I figured out *a* solution.

src/window/gui/Gui.cpp
Malkierian: I did remove the initial window size, though. Letting it autosize itself makes more sense.

PR #647, opened 2024-07-25
Revert frame timing changes

Archez:
(based on discussions in discord) It was found that the calculation changes in the `qpc_to_` methods were truncating/overflowing for a lot of values which lead to the vsync calculations to be incorrect and no longer drop frames as needed causing frame slowdown/slowmotion. On some machines this could manifest as early as 30 minutes of elapsed gameplay.

This reverts the changes to how they are in LUS 1.x which does not truncate/overflow.

This also adjusts the `do {} while` loop for `YieldProcessor` to a regular `while` loop so that we aren't yielding when we don't need to.

### DIFF
--- a/src/graphic/Fast3D/gfx_dxgi.cpp
+++ b/src/graphic/Fast3D/gfx_dxgi.cpp
@@ -524,7 +524,8 @@ static uint64_t qpc_to_ns(uint64_t qpc) {
 }
 
 static uint64_t qpc_to_100ns(uint64_t qpc) {
-    return qpc / dxgi.qpc_freq * _100NANOSECONDS_IN_SECOND + qpc % dxgi.qpc_freq * _100NANOSECONDS_IN_SECOND / dxgi.qpc_freq;
+    return qpc / dxgi.qpc_freq * _100NANOSECONDS_IN_SECOND +
+           qpc % dxgi.qpc_freq * _100NANOSECONDS_IN_SECOND / dxgi.qpc_freq;
 }
 
 static bool gfx_dxgi_start_frame(void) {
--- a/src/graphic/Fast3D/gfx_dxgi.cpp
+++ b/src/graphic/Fast3D/gfx_dxgi.cpp
@@ -41,6 +41,9 @@
 #define FRAME_INTERVAL_NS_NUMERATOR 1000000000
 #define FRAME_INTERVAL_NS_DENOMINATOR (dxgi.target_fps)
 
+#define NANOSECOND_IN_SECOND 1000000000
+#define _100NANOSECONDS_IN_SECOND 10000000
+
 using namespace Microsoft::WRL; // For ComPtr
 
 static struct {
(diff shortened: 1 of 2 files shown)


Malkierian (approved):

src/graphic/Fast3D/gfx_dxgi.cpp
```diff
@@ -517,13 +517,11 @@ static void gfx_dxgi_handle_events(void) {
 }
 
 static uint64_t qpc_to_ns(uint64_t qpc) {
-    qpc *= 1000000000;
-    return qpc / dxgi.qpc_freq;
+    return qpc / dxgi.qpc_freq * 1000000000 + qpc % dxgi.qpc_freq * 1000000000 / dxgi.qpc_freq;
```
briaguya0: could be nice to make this a little less magic and throw
```cpp
#define MAGIC_NUMBER_NAME 1000000000
```
in there and replace this with

```suggestion
    return qpc / dxgi.qpc_freq * MAGIC_NUMBER_NAME + qpc % dxgi.qpc_freq * MAGIC_NUMBER_NAME / dxgi.qpc_freq;
```

src/graphic/Fast3D/gfx_dxgi.cpp
```diff
@@ -517,13 +517,11 @@ static void gfx_dxgi_handle_events(void) {
...
 }
 
 static uint64_t qpc_to_100ns(uint64_t qpc) {
-    qpc *= 10000000;
-    return qpc / dxgi.qpc_freq;
+    return qpc / dxgi.qpc_freq * 10000000 + qpc % dxgi.qpc_freq * 10000000 / dxgi.qpc_freq;
```
briaguya0: same comment as above
```cpp
#define MAGIC_NUMBER_NAME 10000000
```

```suggestion
    return qpc / dxgi.qpc_freq * MAGIC_NUMBER_NAME + qpc % dxgi.qpc_freq * MAGIC_NUMBER_NAME / dxgi.qpc_freq;
```

briaguya0 (commented):
non blocking but left some suggestions to move away from magic numbers

i'm not sure what the best names for what i have as `MAGIC_NUMBER_NAME` are, my initial thought was `QPC_PER_NS` and `QPC_PER_100NS` but that feels wrong considering the `qpc_to_ns` magic number is 100x *bigger* than the `qpc_to_100ns` magic number

src/graphic/Fast3D/gfx_dxgi.cpp
Spodi: Your magic number is just the factor from seconds to ns. Or seconds to 100ns steps respectively. It isn't really directly related to the performance counter. ( qpc [ticks] / qpc_freq [ticks/sec] = time [sec] )

Archez: I've updated the PR to add defines for these constants

briaguya0 (approved):

PR #648, opened 2024-07-25
Windows DPI awareness (#365)

Spodi:
Ports need something like this https://github.com/HarbourMasters/Shipwright/pull/4256. "Works" without, but windows will be confused.

* Handle WM_DPICHANGED

* Don't run as DPI-aware process here.

Use a manifest that declares DPI-awareness for the game instead.

* Override SDLs window procedure in Windows

Something is wrong with SDLs original implementation of  WM_GETDPISCALEDSIZE, so pass it to the default system window procedure.

Should fix window scaling when DPI changes.

* Fix building on non-Windows

* Comment some stuff

* Remove now unused run_as_dpi_aware()

### DIFF
--- a/src/graphic/Fast3D/gfx_dxgi.cpp
+++ b/src/graphic/Fast3D/gfx_dxgi.cpp
@@ -99,58 +99,6 @@ static void load_dxgi_library(void) {
     *(FARPROC*)&dxgi.CreateDXGIFactory2 = GetProcAddress(dxgi.dxgi_module, "CreateDXGIFactory2");
 }
 
-template <typename Fun> static void run_as_dpi_aware(Fun f) {
-    // Make sure Windows 8.1 or newer doesn't upscale/downscale the rendered images.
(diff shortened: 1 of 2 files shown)


Malkierian (approved):
Checked out Windows manifest PR for MacReady, checked this out into it, and merged #649. On a DPI monitor set to 150% scaling, I tested the following:

Internal resolution at 200%, MSAA at 2. Went fullscreen, closed, reopened. Started in fullscreen, didn't crash, didn't have zoom issue. Changed to window, resized to be easily draggable, dragged to non-DPI monitor. Resized appropriately, game graphics also resized appropriately. Side note, Windows apparently chooses the resolution/DPI settings of the monitor that the majority of the game window is on for scaling the window. Good to know moving forward.

This should be good to go, as far as I can tell.

Although, to be fair, booting up normal MacReady, I couldn't replicate the DPI IR crash, so I don't know if it actually fixed, but I tried this after testing this PR.

PR #649, opened 2024-07-26
Reintroduce Keyboard resize fix (#385)

Spodi:
Shouldn't break now with with proper DPI awareness on windows


Malkierian (approved):
Merged this into a test build with #648. Works as advertised.

briaguya0 (approved):

PR #651, opened 2024-07-29
Add a missing include in GuiWindows.h

coco875:
Strangely my os complain about unknow defenition of uint32_t so just the import for it

### DIFF
--- a/src/window/gui/GuiWindow.h
+++ b/src/window/gui/GuiWindow.h
@@ -7,6 +7,7 @@
 #include <imgui.h>
 #include <imgui_internal.h>
 #include "window/gui/GuiElement.h"
+#include <stdint.h>
 
 namespace Ship {
 class GuiWindow : public GuiElement {


briaguya0:
could you be more specific about the issue you're facing? what os? also repro steps using a public port repo would be helpful - do you see the same error when building latest dev of https://github.com/HarbourMasters/Shipwright ?

coco875:
when cmake try compile libultraship part he make error about undefined reference of uint32_t, I use fedora 39 with gcc 13.2.1. I will try reproduce it with the repo mentioned

coco875:
I try on Shipwright and don't get the error but the libultraship version on this repo are before https://github.com/Kenix3/libultraship/commit/181465d2abac3bd2914bb58a2f4d2d4016d25708 who introduce uint32_t

coco875:
by updating libultraship to the lastest commit on Shipwright I get the same error 
image

PR #652, opened 2024-07-29
improve support of f3dex

coco875:


coco875:
https://github.com/Kenix3/libultraship/pull/637 fix this issue normally without ifdef

PR #654, opened 2024-08-02
remove some warning

coco875:
warning that I remove:
* remove deplicate macro definition in controller.h. 
* avoid redefine _LANGUAGE_C.
* only add -Wno-implicit-function-declaration when he compile for C or OBJC.
* fix multiple useless typedef.
* add attribute for some unused.
* add cast for when it's not the same type of int (not fixing all).


coco875:
hmm how you want to add attribute macro ? for other os

Kenix3:
> hmm how you want to add attribute macro ? for other os

We don't really have a standard for this. Will need to do some research to see how it's handled in other projects.

louist103:
> hmm how you want to add attribute macro ? for other os

` [[maybe_unused]] ` and ` [[fallthrough]] ` are in C++ as of c++17 and should work on all platforms

Kenix3:
> > hmm how you want to add attribute macro ? for other os
> 
> `[[maybe_unused]]` and `[[fallthrough]]` are in C++ as of c++17 and should work on all platforms

@coco875

Kenix3 (approved):

PR #655, opened 2024-08-08
Add calls to `Update()` for menu and menubar.

Malkierian:
When setting up menu support, I forgot to add a call to the menu's `Update()` function inside the check for an existent menu.
Also, don't know if it was my fault or not, but menubar also didn't have a call to its `Update()` function.

### DIFF
--- a/src/window/gui/Gui.cpp
+++ b/src/window/gui/Gui.cpp
@@ -375,10 +375,12 @@ void Gui::DrawMenu() {
 #endif
 
     if (GetMenuBar()) {
+        GetMenuBar()->Update();
         GetMenuBar()->Draw();
     }
 
     if (GetMenu()) {
+        GetMenu()->Update();
         GetMenu()->Draw();
     }
 


Archez (approved):

briaguya0:
my gut feeling is that calling update in draw is coupling draw and update in a way that is nonstandard for imgui (i think we are better aligned with imgui's recommended use in `GuiWindow`)

@Kenix3 do you remember if we used to have an `UpdateMenu` method? i feel like we might have had one and dropped it because after a refactor it was doing nothing

briaguya0 (approved):
approving because this matches the pattern we use for everything else (see below)
https://github.com/Kenix3/libultraship/blob/09c39083451d0500fe64caddf3d00b663bca9a95/src/window/gui/Gui.cpp#L389-L392

i do think it's worth making a separate issue for figuring out if there's a way to handle update functions we like more

Kenix3:
Yeah this matches how we do it in other places, so it should be the right pattern.

PR #659, opened 2024-08-18
Implemented light factory and otr light support

KiritoDv:


src/resource/type/Light.h
```diff
@@ -0,0 +1,57 @@
...
+    uint8_t col[3];
+    int8_t pad1;
+    uint8_t colc[3];
+    int8_t pad2;
+};
+
+struct LightEntry {
+    Ambient a;
+    LightData l;
+};
```
briaguya0: these names feel extremely abbreviated - why `LightEntry.a` instead of `LightEntry.ambient`? i assume `PointLightN64.col` means color, but what does `colc` mean? using full descriptive names here would be very helpful for readability.

src/resource/type/Light.h
```diff
@@ -0,0 +1,57 @@
...
+  public:
+    using Resource::Resource;
+
+    Light() : Resource(std::shared_ptr<Ship::ResourceInitData>()) {
+    }
+
+    LightEntry* GetPointer();
+    size_t GetPointerSize();
+
+    LightEntry mLight;
```
briaguya0, line 62: this should be private

src/resource/type/Light.h
KiritoDv: I just took it from the gbi so thats probably why but i can rename it

src/resource/type/Light.h
```diff
@@ -0,0 +1,58 @@
...
+namespace LUS {
+
+struct LightN64 {
+    uint8_t Color[3];
+    int8_t Pad1;
+    uint8_t ColorCopy[3];
+    int8_t Pad2;
+    uint8_t Direction[3];
+    int8_t Pad3;
+};
```
briaguya0, line 22: having the padding in the resource types doesn't really fit with existing patterns. i'd think instead the idea would be to move that complexity into the importer

src/resource/factory/LightFactory.cpp
```diff
@@ -0,0 +1,15 @@
...
+
+std::shared_ptr<Ship::IResource> LUS::ResourceFactoryBinaryLightV0::ReadResource(std::shared_ptr<Ship::File> file) {
+    if (!FileHasValidFormatAndReader(file)) {
+        return nullptr;
+    }
+
+    std::shared_ptr<Light> light = std::make_shared<Light>(file->InitData);
+    auto reader = std::get<std::shared_ptr<Ship::BinaryReader>>(file->Reader);
+
+    reader->Read((char*)light->GetPointer(), sizeof(LightEntry));
```
briaguya0, line 12: this part could be spit up into a bunch of `Read{type}` calls to put the data in the appropriate places and also "read away" the padding without storing it anywhere.

briaguya0 (commented):
i assume @Kenix3 has similar thoughts to these but it's worth waiting to see what he thinks before making any changes based on these suggestions

src/resource/factory/LightFactory.cpp
KiritoDv: I tried, this works better, we had a full conversation about it on the Mega's port channel

src/resource/type/Light.h
KiritoDv: The thing with this is that we dont need the gbi, because this is exactly what we do on Gfx and Mtx, but we take the type from the gbi

briaguya0 (approved):

PR #663, opened 2024-08-25
Various warning fixes

Archez:
Various warning fixes including:
* `warning: format not a string literal and no format arguments [-Wformat-security]`
* `warning: ‘typedef’ was ignored in this declaration`
* `warning: ‘XYZ’ will be initialized after [-Wreorder]`
* a couple integer size/sign values


src/config/Config.cpp
```diff
@@ -137,10 +137,9 @@ void Config::EraseBlock(const std::string& key) {
...
         std::vector<std::string> dots = StringHelper::Split(key, ".");
         if (dots.size() > 1) {
-            int curDot = 0;
             for (auto& dot : dots) {
                 if (gjson2.contains(dot)) {
-                    if (curDot == dots.size()) {
+                    if (dots.size() == 0) {
                         gjson2.erase(dot);
                     } else {
                         gjson2 = gjson2[dot];
```
Archez: @Malkierian you added this method, so I just wanted to confirm with you that `curDot` was only ever meant to be 0 and that this change is fine. Otherwise did you mean for `curDot` to be re-assigned in this for-loop for some logic?

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -3848,9 +3848,10 @@ const char* GfxGetOpcodeName(int8_t opcode) {
         } else {
             SPDLOG_CRITICAL("Unhandled OP code: 0x{:X}, for loaded ucode: {}", (uint8_t)opcode,
                             (uint32_t)ucode_handler_index);
-            return nullptr;
         }
     }
+
+    return nullptr;
 }
```
Pepe20129, line 3857: Probably a good idea to also move the `SPDLOG_CRITICAL`.

Archez: The meaning is slightly different, so I put a new log statement as an `else`

src/config/Config.cpp
Archez: Based on discord discussion, seems like this method is possibly incomplete and will be looked into in a follow up PR.

For now, I've reverted the change and only applied the signed int comparison fix to satisfy the warning.

Kenix3 (approved):

PR #664, opened 2024-08-26
Simplifies the Gui rendering setup.

Kenix3:


src/window/gui/Gui.cpp
```diff
@@ -588,8 +537,60 @@ int16_t Gui::GetIntegerScaleFactor() {
...
+    ImVec2 mainPos = ImGui::GetWindowPos();
+    mainPos.x -= windowPosBeforeMenuBar.x;
+    mainPos.y -= windowPosBeforeMenuBar.y;
     ImVec2 size = ImGui::GetContentRegionAvail();
+    gfx_current_dimensions.width = (uint32_t)(size.x * gfx_current_dimensions.internal_mul);
+    gfx_current_dimensions.height = (uint32_t)(size.y * gfx_current_dimensions.internal_mul);
+    gfx_current_game_window_viewport.x = (int16_t)mainPos.x;
+    gfx_current_game_window_viewport.y = (int16_t)mainPos.y;
+    gfx_current_game_window_viewport.width = (int16_t)size.x;
+    gfx_current_game_window_viewport.height = (int16_t)size.y;
```
briaguya0: i realize we were already doing this every frame, but it's gotten me questioning if we *want* to do all of this every frame

src/window/gui/Gui.cpp
```diff
@@ -324,7 +324,7 @@ void Gui::DrawMenu() {
     ImGui::Begin("Main - Deck", nullptr, windowFlags);
     ImGui::PopStyleVar(3);
 
-    ImVec2 topLeftPos = ImGui::GetWindowPos();
+    windowPosBeforeMenuBar = ImGui::GetWindowPos();
```
briaguya0: it's not clear to me from this variable name what "before" means - before the menu bar was visible (as in time)?, above? to the left of?

a less ambiguous variable name would be great here

src/window/gui/Gui.cpp
```diff
@@ -650,6 +651,13 @@ void Gui::RenderViewports() {
     }
 }
 
+void Gui::EndFrame() {
+    ImGui::EndFrame();
+    if (mImGuiIo->ConfigFlags & ImGuiConfigFlags_ViewportsEnable) {
+        ImGui::UpdatePlatformWindows();
+    }
+}
```
briaguya0, line 656: i see you moved this from where it was in the file originally but i can't tell if that moved it into/out of an `#ifdef` or something - is this just a "this feels like where it should be in the file" move?

briaguya0 (commented):
awesome to see imgui stuff removed from fast3d! few comments

src/window/gui/Gui.cpp
Kenix3: Just moved it closer to some other functions.

src/window/gui/Gui.cpp
Kenix3: Probably not, no, but I don't want to change the logic too much in this PR.

src/window/gui/Gui.cpp
```diff
@@ -588,8 +534,60 @@ int16_t Gui::GetIntegerScaleFactor() {
...
+    }
+
+    switch (CVarGetInteger(CVAR_LOW_RES_MODE, 0)) {
+        case 1: { // N64 Mode
+            gfx_current_dimensions.width = 320;
+            gfx_current_dimensions.height = 240;
+            /*
+            const int sw = size.y * 320 / 240;
+            gfx_current_game_window_viewport.x += ((int)size.x - sw) / 2;
+            gfx_current_game_window_viewport.width = sw;*/
```
briaguya0, line 570: i know this was copy/pasted but if we don't need it now we might as well remove it
```suggestion
```

briaguya0 (approved):
assuming this has been tested this LGTM

src/window/gui/Gui.cpp
Kenix3: What makes you think it's unnecessary?

briaguya0: It's commented out

PR #667, opened 2024-09-01
Fix DPI scaling issues on Windows (OpenGL)

GhostlyDark:
Adds an SDL hint to address DPI scaling issues on Windows when using OpenGL, where the window size can appear too large. This can also help with the resolution in general, making things look less crunchy (200% DPI, 100% internal resolution).

before
image

after
Clipboard_09-01-2024_01

### DIFF
--- a/src/graphic/Fast3D/gfx_sdl2.cpp
+++ b/src/graphic/Fast3D/gfx_sdl2.cpp
@@ -302,6 +302,11 @@ static void gfx_sdl_init(const char* game_name, const char* gfx_api_name, bool s
     window_width = width;
     window_height = height;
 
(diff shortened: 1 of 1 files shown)


KiritoDv:
It could be me but the before looks better

GhostlyDark:
> It could be me but the before looks better

On the second screenshot (SDL2 hint active) the resolution on screen with the exact same settings is essentially higher.
If you think the first screenie looks better because of the larger UI...it's not supposed to be this big. DX11 looks pretty much exactly like OpenGL does with the hint, so I don't understand how this can be perceived as a downgrade.

Clipboard_09-01-2024_02

PR #675, opened 2024-09-21
Fix I4 and IA16 tile textures

AloXado320:
Using the same checks from `import_texture_rgba16`. Fixes texture rendering on those formats if they are used with `gDPLoadTextureTile_4b` and `gDPLoadTextureTile` respectively.

Ideally every `import_texture` function should have those checks but I could only test those 2 formats so far.


Kenix3:
Merge pending test against SoH or 2S2H.

MegaMech:
Please delete line 562, 687, 730, 775

Tested on 2Ship. Checked start menu, mirror shield, ikana area, town, out of town area. Checked picking up some items and using some items. No issues found.

MegaMech (approved):

PR #676, opened 2024-09-24
Use alternate stderr suppression for git patch

Archez:
Yet another tweak to the stormilb patch process for visual studio.

The previous solution used the `--quiet` flag on the git cli, which worked, however this was only "recently" introduced (~2 years ago)
In an effort to support older versions of git, and to avoid reports for when someone inevitably runs into it, I'm switching the command to instead redirect stderr output to `nul` and additionally only applying this logic for when the Visual Studio generator is used.

I've tested that this still works for VS setups.


Malkierian:
Crap, now I have to put this precedent in my gamepad patch.

Malkierian:
And I'll have to wait until this is merged, because I think it'd be better to move the definition of `git_hide_output` outside of Stormlib since more than one patch will use it.

Archez:
Closing in favor of #677 implementing the changes

PR #677, opened 2024-09-24
Add ImGui patch for Gamepad Update, Stormlib patch tweak

Malkierian:
In the process of addressing reviews of the modern menu, it was discovered that SDL was implemented differently from WASAPI, and several things were not behaving properly compared to WASAPI. This adds and implements a patch, following the precedent of the Stormlib patch, to remove a return in the SDL implementation preventing gamepad update when the menubar is closed. As far as I can tell, this had no negative effects on gameplay or menubar function with gamepad navigation.

Also incorporates the Stormlib patch tweak from #676 to avoid having to wait for merge times, unifies the variable setting for Visual Studio generator, and uses the tweak format for the gamepad patch, at Archez's suggestion.

### DIFF
--- a/cmake/dependencies/common.cmake
+++ b/cmake/dependencies/common.cmake
@@ -2,11 +2,17 @@ include(FetchContent)
 
 find_package(OpenGL QUIET)
 
+# When using the Visual Studio generator, it is necessary to suppress stderr output entirely so it does not interrupt the patch command.
+# Redirecting to nul is used here instead of the `--quiet` flag, as that flag was only recently introduced in git 2.25.0 (Jan 2022)
+if (CMAKE_GENERATOR MATCHES "Visual Studio")
+	set(git_hide_output 2> nul)
+endif()
+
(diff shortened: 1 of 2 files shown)


cmake/dependencies/common.cmake
```diff
@@ -2,11 +2,22 @@ include(FetchContent)
 
 find_package(OpenGL QUIET)
 
+# When using the Visual Studio generator, it is necessary to suppress stderr output entirely so it does not interrupt the patch command.
+# Redirecting to nul is used here instead of the `--quiet` flag, as that flag was only recently introduced in git 2.25.0 (Jan 2022)
+if (CMAKE_GENERATOR MATCHES "Visual Studio")
+	set(git_hide_output 2> nul)
```
Archez: [nit] Here and below, can you use 4 spaces instead of tabs to match the whitespacing of this file.

Malkierian: Done.

PR #679, opened 2024-10-10
Gui class refactor.

### LINKED ISSUE #507: `Gui` abstraction
* New abstract base class `GuiRenderer`
* `Gui` renamed to `ImGuiRenderer` inherits from `GuiRenderer`
* `LoadGuiTexture` on `Gui` (`ShipDK` namespace) takes in a `LUS::Texture`
  * This could be handled in https://github.com/Kenix3/libultraship/issues/507 by making `ImGuiRenderer` `LUS::ImGuiRenderer` and inheriting from `ShipDK::GuiRenderer`
* In ConsoleWindow we use ImGui::IsKeyPressed. This should be a function of something else.

Kenix3:
Resolves #507


Kenix3:
SoH bump is here: https://github.com/HarbourMasters/Shipwright/pull/4390

Archez:
Did some testing from my end, here are my findings:
* Metal crashes
* OpenGL fails to render if there are any floating windows.

---

For OpenGL the following needs to happen:
`ImGui::RenderPlatformWindowsDefault` deeply calls `SDL_GL_MakeCurrent` for each floating window as a platform implementation. Once done, we have to manually call `SDL_GL_MakeCurrent` with the backed up values from before `ImGui::RenderPlatformWindowsDefault` so that the GL context is ready for the next frame.

I think `Gui::EndFrame` should look like
```c++
void Gui::EndFrame() {
    ImGui::EndFrame();
    if (mImGuiIo->ConfigFlags & ImGuiConfigFlags_ViewportsEnable) {
        WindowBackend backend = Context::GetInstance()->GetWindow()->GetWindowBackend();
        // OpenGL requires extra platform handling on the GL context
        if (backend == WindowBackend::FAST3D_SDL_OPENGL && mImpl.Opengl.Context != nullptr) {
            // Backup window and context before calling RenderPlatformWindowsDefault
            SDL_Window* backupCurrentWindow = SDL_GL_GetCurrentWindow();
            SDL_GLContext backupCurrentContext = SDL_GL_GetCurrentContext();

            ImGui::UpdatePlatformWindows();
            ImGui::RenderPlatformWindowsDefault();

            // Set back the GL context for next frame
            SDL_GL_MakeCurrent(backupCurrentWindow, backupCurrentContext);
        } else {
            ImGui::UpdatePlatformWindows();
            ImGui::RenderPlatformWindowsDefault();
        }
    }
}
```

Which leaves `Gui::DrawFloatingWindows` looking like
```c++
void Gui::DrawFloatingWindows() {
    // Draw the ImGui "viewports" which are the floating windows.
    ImGui::Render();
    ImGuiRenderDrawData(ImGui::GetDrawData());
}
```

And probably this method should renamed to something else? Alternatively, everything in `Gui::EndFrame` could be moved into this method (aside from ImGui::EndFrame).

---

For Metal, the following lines of code from `Metal_NewFrame` need to be executed at the beginning of `gfx_run` (per each interpolated frame, not each game frame)
```c++
int width, height;
SDL_GetRendererOutputSize(renderer, &width, &height);
gfx_metal_setup_screen_framebuffer(width, height);
```

Specifically, the logic in `gfx_metal_setup_screen_framebuffer` sets up required Metal contexts that the render pipeline in `gfx_run` depend on.

One option would be to split those lines from `Metal_NewFrame` into a new method like `Metal_SetupFrame` and expose a new public method on `Gui` like `SetupRendererFrame` to be called in gfx_pc

```c++
void gfx_run(Gfx* commands, const std::unordered_map<Mtx*, MtxF>& mtx_replacements) {
    Ship::Context::GetInstance()->GetWindow()->GetGui()->SetupRendererFrame();
    ...
```
```c++
void Gui::SetupRendererFrame() {
    switch (Context::GetInstance()->GetWindow()->GetWindowBackend()) {
#ifdef __APPLE__
        case WindowBackend::FAST3D_SDL_METAL:
            Metal_SetupFrame(mImpl.Metal.Renderer);
            break;
#endif
        default:
            break;
    }
}
```

src/window/gui/Gui.h
```diff
@@ -72,48 +72,56 @@ class Gui {
...
+    void DrawGame();
+
+    void ImGuiBackendNewFrame();
+    void ImGuiWMNewFrame();
     void ImGuiWMInit();
     void ImGuiBackendInit();
     void ImGuiRenderDrawData(ImDrawData* data);
+
+    bool GetPadBtnTogglesMenu();
+    void SetPadBtnTogglesMenu(bool padBtnTogglesMenu);
```
Archez: You moved `SetPadBtnTogglesMenu` to `protected`, but 2ship calls it externally for the Modern Menu. /cc @Malkierian

Kenix3: For these to be public and maintained long term, I'd rather we set the bad button via these functions instead of the boolean. How is the button to use currently determined? Can we check against that instead of just a boolean living on this class?

Malkierian: This isn't for setting what pad button opens the menu, this is for switching between toggling menubar and menu with said button. If we end up completely yeeting the menubar in the future, then this becomes unnecessary. But for now, the choice is still up to the ports, thus the option needs to be toggleable.

PR #684, opened 2024-10-11
Adds Powershell one-liner to run clang-format to CONTRIBUTING.md

Kenix3:


Malkierian:
This still requires someone downloading it themselves and putting it in the right folder, right? Might want to look at what 2ship does for automatic download, extraction, and placement for it to further streamline things.

Kenix3:
> This still requires someone downloading it themselves and putting it in the right folder, right? Might want to look at what 2ship does for automatic download, extraction, and placement for it to further streamline things.

This works for me out of the box after I installed LLVM for Windows. If we're talking a _full_ answer, I think the solution is a git pre-commit hook that runs something similar to 2ship's script.

briaguya0 (approved):
i see pros and cons to the automatic download method 2ship has - biggest con being that it's more to maintain.

i don't think it's unreasonable to ask LUS contributors to install clang-format, so for now i say :shipit:

Malkierian:
If we're going to require LLVM to be installed, then we should at least unify versioning across projects. 2ship currently needs 13.

Malkierian:
And yes, clang version has been shown to have significant impacts on the interpretation and application of rulesets.

Kenix3:
LUS has used and required 12 for a long time.

Malkierian:
I don't care which version gets settled on, I just want everything to use the same version, so perhaps this method could be setup for 2ship as well and not have to worry about the download.

PR #690, opened 2024-10-15
Updates to Context.cpp for macOS Game mode support

larsy1995:
The needed changes to get Game Mode working for both 2ship and soh. It is dynamic, and with the correct changes to info.plist and packaging-2.cmake, it will work for all future projects as well. 
Moves finding the games user files directory and making the directories out of the shell scripts to allow the macOS builds to be run directly instead of as a child process. 

Needs to be pulled here before, and merged into 2Ship2Harkinian and Shipwright at the same time as merging the PRs for those projects.

### DIFF
--- a/src/Context.cpp
+++ b/src/Context.cpp
@@ -379,7 +379,7 @@ std::string Context::GetAppDirectoryPath(std::string appName) {
         return std::string(fpath);
     }
 #endif
-    
+
 #if defined(__linux__)
     char* fpath = std::getenv("SHIP_HOME");
     if (fpath != NULL) {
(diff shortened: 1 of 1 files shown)


larsy1995:
Moved the folder and file creation out of LUS and into port, the only changes now are the ones needed to get GetAppDirectoryPath working with macOS with the game mode changes.

larsy1995:
I have no idea what that tidy-format, clang-format wants, would be appreciated if anyone told me.

larsy1995:
Tidy-format was successful locally with that commit now.

PR #723, opened 2024-11-21
Add Widescreen related funcs to lus

MegaMech:
Work in progress

Once this PR is approved, I will submit PRs to the other ports to update to reflect these changes.

Please let me know if anything is missing

### DIFF
--- a/src/public/bridge/windowbridge.cpp
+++ b/src/public/bridge/windowbridge.cpp
@@ -39,9 +39,9 @@ bool WindowIsFullscreen(void) {
  * or any situation where a screen coordinate outside the normal N64 bounds of 320x240 are required.
  *
  * How to use:
- * LeftEdgeAlign(0) --> Returns the left most edge of the game render target area
- * LeftEdgeAlign(320) --> Returns the right most edge of the game render target area
- * RightEdgeAlign(320) --> Returns the right most edge of the game render target area
+ * TranslateXRelativeToLeftEdge(0) --> Returns the left most edge of the game render target area
+ * TranslateXRelativeToLeftEdge(320) --> Returns the right most edge of the game render target area
+ * TranslateXRelativeToRightEdge(320) --> Returns the right most edge of the game render target area
  * 
  * Align a rectangle with the left of the screen:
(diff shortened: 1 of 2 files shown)


src/public/bridge/windowbridge.cpp
```diff
@@ -31,4 +31,41 @@ int32_t WindowGetPosY(void) {
...
+ *
+ */
+
+float ScreenGetAspectRatio() {
+    return gfx_current_dimensions.aspect_ratio;
+}
+
+// AdjustXFromLeftEdge 
+float LeftEdgeAlign(float v) {
+    return (SCREEN_WIDTH / 2 - SCREEN_HEIGHT / 2 * OTRGetAspectRatio() + (v));
```
Archez: These `SCREEN_WIDTH` and `SCREEN_HEIGHT` should be replaced with `gfx_native_dimensions.width` and `gfx_native_dimensions.height` respectively.

This is necessary for games that change resolutions mid-game (mm with the bombers notebook).

src/public/bridge/windowbridge.h
```diff
@@ -17,8 +17,16 @@ int32_t WindowGetPosX(void);
 int32_t WindowGetPosY(void);
 bool WindowIsFullscreen(void);
 
+float ScreenGetAspectRatio(void);
+float LeftEdgeAlign(float v);
+int16_t LeftEdgeRectAlign(float v);
+float RightEdgeAlign(float v);
+int16_t RightEdgeRectAlign(float v);
+uint32_t WindowGetGameRenderWidth();
+uint32_t WindowGetGameRenderHeight();
```
Archez, line 26: I know this is WIP still, but wanted to ask the question early.

Since these are coming from the WindowBridge and are `extern "C"` we might want to consider these all having a `Window` prefix to indicate the system they are coming from. Aligning with the recent LUS namespacing clean up efforts.

MegaMech: But it's not really "WindowGetLeftEdge" it's more like "ScreenGetLeftEdge"

I don't mind adding Window, but just consider the name may be confusing.

PR #724, opened 2024-11-22
Allow Ports To Configure Default Button Mappings

MegaMech:
## Replaces hardcoded default button mappings with cmake variables

Replace keyboard button enums with default macros set in `cvars.cmake`. This allows ports to define their own default using `CMake/lus-cvars.cmake` file.

A PR to SOH and 2Ship should not be necessary because I believe they use the default button mappings.


briaguya0 (commented):
overall the logic looks reasonable, but i'm not a huge fan of the `LUS_BTN_*` names

although verbose, i'd prefer something more descriptive such as `LUS_DEFAULT_KB_MAPPING_*`

edit: also, the name of this pr feels a little misleading. LUS already has default keyboard mappings, if i were to name the PR i'd call it "allow ports to configure default keyboard button mappings"

cmake/cvars.cmake
```diff
@@ -20,6 +20,26 @@ set(CVAR_GAME_OVERLAY_FONT "gOverlayFont" CACHE STRING "")
...
 set(CVAR_PREFIX_CONTROLLERS "gControllers" CACHE STRING "")
 set(CVAR_PREFIX_ADVANCED_RESOLUTION "gAdvancedResolution" CACHE STRING "")
+
+set(LUS_BTN_A KbScancode::LUS_KB_X CACHE STRING "")
+set(LUS_BTN_B KbScancode::LUS_KB_C CACHE STRING "")
+set(LUS_BTN_L KbScancode::LUS_KB_E CACHE STRING "")
+set(LUS_BTN_R KbScancode::LUS_KB_R CACHE STRING "")
+set(LUS_BTN_Z KbScancode::LUS_KB_Z CACHE STRING "")
+set(LUS_BTN_START KbScancode::LUS_KB_SPACE CACHE STRING "")
+set(LUS_BTN_STICKUP LUS_KB_W CACHE STRING "")
```
briaguya0: feels slightly odd to have some with `KbScancode::` and some without. it seems that's leftover from before (based on changes in `AxisDirectionMappingFactory`) - but this feels like a nice opportunity to add some consistency

cmake/cvars.cmake
```diff
@@ -43,4 +63,23 @@ add_compile_definitions(
 	CVAR_MENU_BAR_OPEN="${CVAR_MENU_BAR_OPEN}"
 	CVAR_PREFIX_CONTROLLERS="${CVAR_PREFIX_CONTROLLERS}"
 	CVAR_PREFIX_ADVANCED_RESOLUTION="${CVAR_PREFIX_ADVANCED_RESOLUTION}"
-)
\ No newline at end of file
+
+	LUS_DEFAULT_KB_MAPPING_A=${LUS_DEFAULT_KB_MAPPING_A}
```
briaguya0, line 67: it looks like all the lines above this have `${}` surrounded by `""`

i'm not sure if that's required syntax for cmake or not but if these don't break as

```suggestion
	LUS_DEFAULT_KB_MAPPING_A="${LUS_DEFAULT_KB_MAPPING_A}"
```

etc i'd think we'd want to do that for consistency

MegaMech: The quotes result in syntax errors. It passes the physical string to the function.

Kenix3: I agree with Bria's naming suggestion here. Can we get this in?

MegaMech: I'm trying it again just-in-case it does work this time, even though it won't. :P

briaguya0 (approved):

Kenix3:
> overall the logic looks reasonable, but i'm not a huge fan of the `LUS_BTN_*` names
> 
> although verbose, i'd prefer something more descriptive such as `LUS_DEFAULT_KB_MAPPING_*`
> 
> edit: also, the name of this pr feels a little misleading. LUS already has default keyboard mappings, if i were to name the PR i'd call it "allow ports to configure default keyboard button mappings"

I agree with Bria's naming suggestion here. Can we get this in?

cmake/cvars.cmake
MegaMech: image

MegaMech:
Done! (Albeit, I feel like I'm now in a sort of time distortion because I can't get in again what I've already gotten in)

MegaMech:
Don't merge yet. Starship wants controller defaults

briaguya0:
> Don't merge yet. Starship wants controller defaults

that doesn't need to be the same PR though right?

briaguya0:
> Done! (Albeit, I feel like I'm now in a sort of time distortion because I can't get in again what I've already gotten in)

i suggested "allow ports to configure default keyboard button mappings" and the title is currently "Allow Keyboard Button Mapping Configuration In Games"

"Allow Keyboard Button Mapping Configuration In Games" is ambiguous, **players** can already configure keyboard button mappings. this PR is replacing hardcoded **default** keyboard button mappings with cmake variables. come to think of it that'd be another reasonable title - "replace hardcoded default keyboard button mappings with cmake variables"

src/controller/controldevice/controller/mapping/factories/ButtonMappingFactory.cpp
```diff
@@ -147,74 +153,74 @@ ButtonMappingFactory::CreateDefaultSDLButtonMappings(ShipDeviceIndex shipDeviceI
...
             }
             break;
         case BTN_CLEFT:
             mappings.push_back(std::make_shared<SDLAxisDirectionToButtonMapping>(shipDeviceIndex, portIndex, BTN_CLEFT,
-                                                                                 SDL_CONTROLLER_AXIS_RIGHTX, -1));
+                                                                                 LUS_DEFAULT_CONT_MAPPING_CLEFT, -1));
             if (isGameCube) {
                 mappings.push_back(std::make_shared<SDLButtonToButtonMapping>(shipDeviceIndex, portIndex, BTN_CLEFT,
-                                                                              SDL_CONTROLLER_BUTTON_Y));
+                                                                              LUS_DEFAULT_CONT_MAPPING_CLEFT2));
```
briaguya0: these would benefit from more descriptive names. a consumer of LUS shouldn't have to know that `CLEFT2` is for an `SDLButton` only on gamecube controllers and `CLEFT` is for an `SDLAxisDirection`

cmake/cvars.cmake
```diff
@@ -20,6 +20,44 @@ set(CVAR_GAME_OVERLAY_FONT "gOverlayFont" CACHE STRING "")
...
+set(LUS_DEFAULT_KB_MAPPING_CUP KbScancode::LUS_KB_ARROWKEY_UP CACHE STRING "")
+set(LUS_DEFAULT_KB_MAPPING_CDOWN KbScancode::LUS_KB_ARROWKEY_DOWN CACHE STRING "")
+set(LUS_DEFAULT_KB_MAPPING_CLEFT KbScancode::LUS_KB_ARROWKEY_LEFT CACHE STRING "")
+set(LUS_DEFAULT_KB_MAPPING_CRIGHT KbScancode::LUS_KB_ARROWKEY_RIGHT CACHE STRING "")
+set(LUS_DEFAULT_KB_MAPPING_DUP KbScancode::LUS_KB_T CACHE STRING "")
+set(LUS_DEFAULT_KB_MAPPING_DDOWN KbScancode::LUS_KB_G CACHE STRING "")
+set(LUS_DEFAULT_KB_MAPPING_DLEFT KbScancode::LUS_KB_F CACHE STRING "")
+set(LUS_DEFAULT_KB_MAPPING_DRIGHT KbScancode::LUS_KB_H CACHE STRING "")
+
+set(LUS_DEFAULT_CONT_MAPPING_A SDL_CONTROLLER_BUTTON_A CACHE STRING "")
```
briaguya0: curious as to what others think about the naming here. considering these are used specifically for `SDL`, i think i'd prefer `LUS_DEFAULT_SDL_MAPPING_*` over `LUS_DEFAULT_CONT_MAPPING_*`

src/controller/controldevice/controller/mapping/factories/ButtonMappingFactory.cpp
```diff
@@ -147,74 +153,74 @@ ButtonMappingFactory::CreateDefaultSDLButtonMappings(ShipDeviceIndex shipDeviceI
...
             if (isGameCube) {
                 mappings.push_back(std::make_shared<SDLButtonToButtonMapping>(shipDeviceIndex, portIndex, BTN_CDOWN,
-                                                                              SDL_CONTROLLER_BUTTON_RIGHTSHOULDER));
+                                                                              LUS_DEFAULT_CONT_MAPPING_CDOWN2));
             }
             break;
         case BTN_CLEFT:
             mappings.push_back(std::make_shared<SDLAxisDirectionToButtonMapping>(shipDeviceIndex, portIndex, BTN_CLEFT,
-                                                                                 SDL_CONTROLLER_AXIS_RIGHTX, -1));
+                                                                                 LUS_DEFAULT_CONT_MAPPING_CLEFT, -1));
```
briaguya0: the `-1` is important here as it is what determines that we mean left on the x axis instead of right. `LUS_DEFAULT_CONT_MAPPING_CLEFT` is only the axis part of `AxisDirection`

briaguya0 (changes requested):
the controller side of things feels a bit more complex because it uses both `SDLButtonToButtonMapping` and `SDLAxisDirectionToButtonMapping`

the current implementation allows ports to
* set any SDL **button** to:
  * any N64 button (including dpad and c buttons)
* kind of configure SDL axes for the c buttons (axis vs axis direction issues)

the current implementation does not allow port to
* set SDL buttons to analog stick axis directions
* set SDL axis directions to any non-c-stick buttons

my opinion is the controller stuff requires quite a bit more architectural thinking than the keyboard stuff does, and it should be split out into a separate PR

src/controller/controldevice/controller/mapping/factories/ButtonMappingFactory.cpp
MegaMech: I was copying the name of BTN_CLEFT

cmake/cvars.cmake
MegaMech: Sure! But for me I didn't realize at first that SDL referred to a physical gamepad.

src/controller/controldevice/controller/mapping/factories/ButtonMappingFactory.cpp
MegaMech: My bad, I didn't see the conditional gamecube statement.

MegaMech:
> my opinion is the controller stuff requires quite a bit more architectural thinking

Would any port actually want to override where the analogue stick goes? I think usually with the different brands of controllers the main issue is the face buttons are different like A/B and X/Square or whatever.

At the very least, with starship's release coming up, I think the point is moreso about having the ability to set what they need, and not necessarily a full blown implementation.

briaguya0:
>  I think the point is moreso about having the ability to set what they need, and not necessarily a full blown implementation.

i still think the controller part should be a separate PR. the axis direction vs button stuff is more complex than the keyboard stuff. let's land the keyboard stuff and figure out what an MVP for controller stuff looks like after.

src/controller/controldevice/controller/mapping/factories/ButtonMappingFactory.cpp
briaguya0: you replaced `SDL_CONTROLLER_AXIS_RIGHTX` with `LUS_DEFAULT_CONT_MAPPING_CLEFT`

`SDL_CONTROLLER_AXIS_RIGHTX` is used for both left ***and*** right

Kenix3 (approved):

briaguya0:
this shouldn't land as-is. the controller stuff either needs to be reworked or removed. the keyboard stuff seems ok as-is, but if we want to do controller stuff moving away from cmake vars is a better architectural decision.

briaguya0 (approved):
:shipit:

briaguya0:
even if we decide to add more robust (non-cmake-based) default setting in the future, having this shouldn't get in the way.

copying my previous note from discord here so it's in a more permanent place

> if we want to go that route moving away from cmake vars makes sense. i think the move would be to update `InitControlDeck` to optionally take in default mappings and store them somewhere
> 
> the stored default mappings can't be the same class as the saved mappings, for example <https://github.com/Kenix3/libultraship/blob/80c2c5e9c6a5681428f759100c2428c5e41f333c/src/controller/controldevice/controller/mapping/factories/AxisDirectionMappingFactory.cpp#L81-L90> has
> 
> ```cpp
> std::make_shared<KeyboardKeyToAxisDirectionMapping>(portIndex, stick, LEFT, LUS_KB_A),
> ```
> 
> `portIndex` and `stick` aren't part of the hardcoded defaults, so we'd need a way to store just the direction and the keyboard code in that case.
> 
> `SDLAxisDirectionToAxisDirectionMapping` is a little more complicated <https://github.com/Kenix3/libultraship/blob/80c2c5e9c6a5681428f759100c2428c5e41f333c/src/controller/controldevice/controller/mapping/factories/AxisDirectionMappingFactory.cpp#L104-L112>
> 
> we have some logic in there for setting the axis based on which stick it is, `stick == LEFT_STICK ? 0 : 2` is saying "if this is the left stick, map axis 0, if it's the right stick, map axis 2" and likewise `stick == LEFT_STICK ? 1 : 3` is saying ""if this is the left stick, map axis 1, if it's the right stick, use axis 3"
> 
> for button mappings we'd just need to make sure we pass in/store something that can be filtered by bitmask
> 
> for keyboard <https://github.com/Kenix3/libultraship/blob/80c2c5e9c6a5681428f759100c2428c5e41f333c/src/controller/controldevice/controller/mapping/factories/ButtonMappingFactory.cpp#L73-L130> we switch on bitmask, we'd want to just filter the stored defaults instead and then apply them to the appropriate `portIndex`
> 
> for SDL button mappings <https://github.com/Kenix3/libultraship/blob/80c2c5e9c6a5681428f759100c2428c5e41f333c/src/controller/controldevice/controller/mapping/factories/ButtonMappingFactory.cpp#L132-L222> this actually opens up quite a bit of room for improvement
> 
> we currently switch on bitmask and then also have some logic to map x and y on gamecube controllers to c buttons like vc/gc oot/mm have.  we could instead store defaults in a way that they can be generic and apply to any SDL controller, and also store specific defaults for known controllers
> 
> i feel like trying to do all of this in cmake would be a nightmare

PR #725, opened 2024-11-23
multiplayer disconnected controller fixes

briaguya0:


src/controller/deviceindex/ControllerDisconnectedWindow.cpp
```diff
@@ -139,15 +139,10 @@ void ControllerDisconnectedWindow::DrawElement() {
...
-    ImGui::OpenPopup("Controller Disconnected");
-    if (ImGui::BeginPopupModal("Controller Disconnected", NULL, ImGuiWindowFlags_AlwaysAutoResize)) {
-        // todo: don't use UINT8_MAX-1 to mean we don't know what controller was disconnected
-        if (mPortIndexOfDisconnectedController == UINT8_MAX - 1) {
-            DrawUnknownOrMultipleControllersDisconnected();
-        } else {
-            DrawKnownControllerDisconnected();
-        }
-        ImGui::EndPopup();
+    if (mPortIndexOfDisconnectedController == UINT8_MAX - 1) {
```
briaguya0, line 142: i was getting a ton of mismatched open/close popup errors, i assume something changed with modern menu to make it so this can just be simplified - seems to work having it this way now

src/controller/deviceindex/ShipDeviceIndexMappingManager.cpp
```diff
@@ -409,6 +409,8 @@ void ShipDeviceIndexMappingManager::HandlePhysicalDeviceDisconnectSinglePlayer(i
 }
 
 void ShipDeviceIndexMappingManager::HandlePhysicalDeviceDisconnectMultiplayer(int32_t sdlJoystickInstanceId) {
+    auto portIndexOfPhysicalDeviceThatHasBeenDisconnected =
+        GetPortIndexOfDisconnectedPhysicalDevice(sdlJoystickInstanceId);
```
briaguya0, line 413: `GetShipDeviceIndexOfDisconnectedPhysicalDevice` closes the controller

https://github.com/Kenix3/libultraship/blob/80c2c5e9c6a5681428f759100c2428c5e41f333c/src/controller/deviceindex/ShipDeviceIndexMappingManager.cpp#L546

so by the time this was being called 

https://github.com/Kenix3/libultraship/blob/80c2c5e9c6a5681428f759100c2428c5e41f333c/src/controller/deviceindex/ShipDeviceIndexMappingManager.cpp#L481

`mController` would already be null

https://github.com/Kenix3/libultraship/blob/80c2c5e9c6a5681428f759100c2428c5e41f333c/src/controller/controldevice/controller/mapping/sdl/SDLMapping.cpp#L145-L146

meaning it always didn't know what device was disconnected

moving this up here makes it not a problem

PR #726, opened 2024-11-23
readme update to test ios ci

briaguya0:


briaguya0:
fixed by https://github.com/Kenix3/libultraship/pull/728

PR #729, opened 2024-12-01
Added path into texture metadata

KiritoDv:


Kenix3:
What is the purpose of this? The texture object already stores the Resource which has the ResourceInitData which includes the path.

PR #730, opened 2024-12-02
Changed format on gfx trace

KiritoDv:


src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -4011,8 +4011,14 @@ static void gfx_step() {
 
 #ifdef USE_GBI_TRACE
     if (cmd->words.trace.valid && CVarGetInteger("gEnableGFXTrace", 0)) {
-        SPDLOG_INFO("Trace File: {}", cmd->words.trace.file);
-        SPDLOG_INFO("Trace Line: {}", cmd->words.trace.idx);
+#define TRACE                                  \
```
Kenix3, line 4014: Why the extra define here? It's also never terminated.

PR #733, opened 2024-12-04
Fix LoadTile Height for f3dex 0.95 games

MegaMech:


Malkierian:
As a note, this is still being debated as it only fixes things for one port and could adversely affect others based on f3dex 1.

PR #737, opened 2024-12-08
Optimize filtered `ListFiles()` to only iterate once.

Malkierian:
Previously, the filtered ListFiles would call the unfiltered ListFiles, which would iterate over the whole of the hash map, then would iterate on the return from the unfiltered one to add matches to a separate list that gets returned. This completely removes the unfiltered function, moves the logic to the filtered function, and adds a check for empty filter to always add something when the unfiltered ListFiles would have been called.

### DIFF
--- a/src/resource/archive/ArchiveManager.cpp
+++ b/src/resource/archive/ArchiveManager.cpp
@@ -67,21 +67,17 @@ bool ArchiveManager::HasFile(uint64_t hash) {
 }
 
 std::shared_ptr<std::vector<std::string>> ArchiveManager::ListFiles(const std::string& filter) {
-    auto list = ListFiles();
-    auto result = std::make_shared<std::vector<std::string>>();
-
-    std::copy_if(list->begin(), list->end(), std::back_inserter(*result),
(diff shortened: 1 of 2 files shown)


src/resource/archive/ArchiveManager.cpp
```diff
@@ -67,19 +67,11 @@ bool ArchiveManager::HasFile(uint64_t hash) {
 }
 
 std::shared_ptr<std::vector<std::string>> ArchiveManager::ListFiles(const std::string& filter) {
-    auto list = ListFiles();
-    auto result = std::make_shared<std::vector<std::string>>();
-
-    std::copy_if(list->begin(), list->end(), std::back_inserter(*result),
```
briaguya0, line 73: @Kenix3 do you remember the reasoning behind doing a copy here?

briaguya0 (approved):
seems reasonable to me! left a question for kenix but i'd be happy to just :shipit:

src/resource/archive/ArchiveManager.cpp
Kenix3: The original code that looks like a copy here was just how we copied from one list to another. It's copy with this lambda filter. If Malk's way is faster then the new code is kosher.

PR #740, opened 2024-12-11
Uses a struct in in mResourceCache in the ResourceManager.

### LINKED ISSUE #687: `ResourceManager` load and unload resource functions include an Archive.
- [x] Archive will need a Map compatible comparison operator. They are the same if the Archive path matches.
- [x] Add a new struct for resource cache lookup: shared ptr to Archive, String cache key, and String resource name. The struct will need a Map compatible comparison operator. Will string match the two string and string match the Archive names.
- [x] ResourceManager function that take in a cache key and a resource path should take in the struct as well.
- [x] ResourceManager functions have a new alternate version of functions that just take in the name like we had before.

Kenix3:
Resolves #687

### DIFF
--- a/src/resource/ResourceManager.h
+++ b/src/resource/ResourceManager.h
@@ -34,7 +34,7 @@ struct ResourceIdentifier {
     size_t mHash;
 };
 
-struct ResourceCacheDataHash {
+struct ResourceIdentifierHash {
     size_t operator()(const ResourceIdentifier& rcd) const;
 };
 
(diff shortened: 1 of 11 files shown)


src/resource/ResourceManager.h
```diff
@@ -8,19 +8,29 @@
...
 struct File;
 
-// Resource manager caches any and all files it comes across into memory. This will be unoptimal in the future when
-// modifications have gigabytes of assets. It works with the original game's assets because the entire ROM is 64MB and
-// fits into RAM of any semi-modern PC.
+struct ResourceCacheData {
+    // Path can either be a Path or a Search Mask including globs depending on usage.
+    const std::string Path = "";
+    const uintptr_t Owner = 0;
+    const std::shared_ptr<Archive> Archive = nullptr;
```
briaguya0: looks like the linux build doesn't like this name

PR #741, opened 2024-12-11
Mouse mapping support

lightmanLP:
Support for binding mouse keys

- [x] update dxgi mouse capture to make it compatible with sdl
- [x] fix fullscreen dxgi mouse capture behavior

co-authored-by: @lilacLunatic

### DIFF
--- a/src/controller/controldeck/ControlDeck.cpp
+++ b/src/controller/controldeck/ControlDeck.cpp
@@ -51,6 +51,19 @@ bool ControlDeck::ProcessKeyboardEvent(KbEventType eventType, KbScancode scancod
     return result;
 }
 
+bool ControlDeck::ProcessMouseEvent(bool isPressed, MouseBtn button) {
+    bool result = false;
+    for (auto port : mPorts) {
+        auto controller = port->GetConnectedController();
+
+        if (controller != nullptr) {
+            result = controller->ProcessMouseEvent(isPressed, button) || result;
+        }
+    }
+
+    return result;
+}
+
 bool ControlDeck::AllGameInputBlocked() {
     return !mGameInputBlockers.empty();
 }
@@ -64,6 +77,11 @@ bool ControlDeck::GamepadGameInputBlocked() {
 bool ControlDeck::KeyboardGameInputBlocked() {
     // block keyboard input when typing in imgui
     return AllGameInputBlocked() || ImGui::GetIO().WantCaptureKeyboard;
+}
+
+bool ControlDeck::MouseGameInputBlocked() {
+    // block mouse input when mouse capture (hidden & window-focused mouse) mode disabled
+    return AllGameInputBlocked() || !Context::GetInstance()->GetWindow()->IsMouseCaptured();
 }
 
 std::shared_ptr<Controller> ControlDeck::GetControllerByPort(uint8_t port) {
--- a/src/controller/controldeck/ControlDeck.cpp
+++ b/src/controller/controldeck/ControlDeck.cpp
@@ -79,18 +79,20 @@ bool ControlDeck::GamepadGameInputBlocked() {
 
 bool ControlDeck::KeyboardGameInputBlocked() {
     // block keyboard input when typing in imgui
-    return AllGameInputBlocked() || ImGui::GetIO().WantCaptureKeyboard;
+    ImGuiWindow *activeIDWindow = ImGui::GetCurrentContext()->ActiveIdWindow;
+    return AllGameInputBlocked() ||
+        (activeIDWindow != NULL && activeIDWindow->ID != Context::GetInstance()->GetWindow()->GetGui()->GetMainGameWindowID()) ||
+        ImGui::GetTopMostPopupModal() != NULL; //ImGui::GetIO().WantCaptureKeyboard, but ActiveId check altered
 }
 
 bool ControlDeck::MouseGameInputBlocked() {
     // block mouse input when user interacting with gui
     // TODO: check perfomance
-    static ImGuiID gameWindowID = ImGui::FindWindowByName("Main Game")->ID;
     ImGuiWindow *window = ImGui::GetCurrentContext()->HoveredWindow;
     if (window == NULL) {
         return true;
     }
-    return AllGameInputBlocked() || (window->ID != gameWindowID);
+    return AllGameInputBlocked() || (window->ID != Context::GetInstance()->GetWindow()->GetGui()->GetMainGameWindowID());
 }
 
 std::shared_ptr<Controller> ControlDeck::GetControllerByPort(uint8_t port) {
--- a/src/controller/controldeck/ControlDeck.cpp
+++ b/src/controller/controldeck/ControlDeck.cpp
@@ -35,6 +35,7 @@ void ControlDeck::Init(uint8_t* controllerBits) {
     // if we don't have a config for controller 1, set default keyboard bindings
     if (!mPorts[0]->GetConnectedController()->HasConfig()) {
         mPorts[0]->GetConnectedController()->AddDefaultMappings(ShipDeviceIndex::Keyboard);
+        mPorts[0]->GetConnectedController()->AddDefaultMappings(ShipDeviceIndex::Mouse);
     }
 
     Context::GetInstance()->GetWindow()->GetGui()->GetGuiWindow("Controller Reordering")->Show();
--- a/src/controller/controldeck/ControlDeck.cpp
+++ b/src/controller/controldeck/ControlDeck.cpp
@@ -10,6 +10,8 @@
 #include <imgui.h>
 #include "controller/deviceindex/ShipDeviceIndexMappingManager.h"
 
+#include "spdlog/spdlog.h"
+
 namespace Ship {
 
 ControlDeck::ControlDeck(std::vector<CONTROLLERBUTTONS_T> additionalBitmasks) : mSinglePlayerMappingMode(false) {
@@ -80,8 +82,14 @@ bool ControlDeck::KeyboardGameInputBlocked() {
 }
 
 bool ControlDeck::MouseGameInputBlocked() {
-    // block mouse input when mouse capture (hidden & window-focused mouse) mode disabled
-    return AllGameInputBlocked() || !Context::GetInstance()->GetWindow()->IsMouseCaptured();
+    // block mouse input when user interacting with gui
+    // TODO: check perfomance
+    static ImGuiID gameWindowID = ImGui::FindWindowByName("Main Game")->ID;
+    ImGuiWindow *window = ImGui::GetCurrentContext()->HoveredWindow;
+    if (window == NULL) {
+        return true;
+    }
+    return AllGameInputBlocked() || (window->ID != gameWindowID);
 }
 
 std::shared_ptr<Controller> ControlDeck::GetControllerByPort(uint8_t port) {
--- a/src/controller/controldeck/ControlDeck.cpp
+++ b/src/controller/controldeck/ControlDeck.cpp
@@ -79,20 +79,22 @@ bool ControlDeck::GamepadGameInputBlocked() {
 
 bool ControlDeck::KeyboardGameInputBlocked() {
     // block keyboard input when typing in imgui
-    ImGuiWindow *activeIDWindow = ImGui::GetCurrentContext()->ActiveIdWindow;
+    ImGuiWindow* activeIDWindow = ImGui::GetCurrentContext()->ActiveIdWindow;
     return AllGameInputBlocked() ||
-        (activeIDWindow != NULL && activeIDWindow->ID != Context::GetInstance()->GetWindow()->GetGui()->GetMainGameWindowID()) ||
-        ImGui::GetTopMostPopupModal() != NULL; //ImGui::GetIO().WantCaptureKeyboard, but ActiveId check altered
+           (activeIDWindow != NULL &&
+            activeIDWindow->ID != Context::GetInstance()->GetWindow()->GetGui()->GetMainGameWindowID()) ||
+           ImGui::GetTopMostPopupModal() != NULL; // ImGui::GetIO().WantCaptureKeyboard, but ActiveId check altered
 }
 
 bool ControlDeck::MouseGameInputBlocked() {
     // block mouse input when user interacting with gui
     // TODO: check perfomance
-    ImGuiWindow *window = ImGui::GetCurrentContext()->HoveredWindow;
+    ImGuiWindow* window = ImGui::GetCurrentContext()->HoveredWindow;
     if (window == NULL) {
         return true;
     }
-    return AllGameInputBlocked() || (window->ID != Context::GetInstance()->GetWindow()->GetGui()->GetMainGameWindowID());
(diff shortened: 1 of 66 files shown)


src/controller/controldeck/ControlDeck.cpp
```diff
@@ -63,7 +79,22 @@ bool ControlDeck::GamepadGameInputBlocked() {
 
 bool ControlDeck::KeyboardGameInputBlocked() {
     // block keyboard input when typing in imgui
-    return AllGameInputBlocked() || ImGui::GetIO().WantCaptureKeyboard;
+    ImGuiWindow* activeIDWindow = ImGui::GetCurrentContext()->ActiveIdWindow;
+    return AllGameInputBlocked() ||
+           (activeIDWindow != NULL &&
+            activeIDWindow->ID != Context::GetInstance()->GetWindow()->GetGui()->GetMainGameWindowID()) ||
+           ImGui::GetTopMostPopupModal() != NULL; // ImGui::GetIO().WantCaptureKeyboard, but ActiveId check altered
+}
```
briaguya0, line 86: just curious - why does the keyboard game input blocking logic need to change for mouse support?

src/controller/controldeck/ControlDeck.cpp
```diff
@@ -63,7 +79,22 @@ bool ControlDeck::GamepadGameInputBlocked() {
...
+
+bool ControlDeck::MouseGameInputBlocked() {
+    // block mouse input when user interacting with gui
+    // TODO: check perfomance
```
briaguya0: since this is only called within mouse mappings this shouldn't have any performance impact on existing stuff, so even if the performance isn't great at the moment that shouldn't block merging this

src/window/MouseMeta.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#include <string>
+
+namespace Ship {
+enum class MouseBtn { LEFT, MIDDLE, RIGHT, BACKWARD, FORWARD, MOUSE_BTN_COUNT, MOUSE_BTN_UNKNOWN };
+static std::string mouseBtnNames[7] = { "MouseLeft",    "MouseMiddle",     "MouseRight",       "MouseBackward",
+                                        "MouseForward", "MOUSE_BTN_COUNT", "MOUSE_BTN_UNKNOWN" };
+} // namespace Ship
```
briaguya0: window feels like a slightly odd place for this file, we have keyboard scancodes in https://github.com/Kenix3/libultraship/blob/main/src/controller/controldevice/controller/mapping/keyboard/KeyboardScancodes.h

src/controller/controldevice/controller/mapping/mouse/MouseKeyToAnyMapping.cpp
```diff
@@ -0,0 +1,35 @@
+#include "MouseKeyToAnyMapping.h"
+#include "Context.h"
+
+#include "utils/StringHelper.h"
+#include "window/gui/IconsFontAwesome4.h"
+
+namespace Ship {
+MouseKeyToAnyMapping::MouseKeyToAnyMapping(MouseBtn button)
```
briaguya0: we should probably standardize on mouse button vs mouse key, `MouseKeyToAnyMapping(MouseBtn` seems a little odd

i'd lean towards `MouseButton` instead of `MouseKey` for all the names but i'm open to mousekey too

src/controller/controldevice/controller/ControllerButton.cpp
```diff
@@ -178,9 +181,10 @@ bool ControllerButton::AddOrEditButtonMappingFromRawPress(CONTROLLERBUTTONS_T bi
     mUseKeydownEventToCreateNewMapping = true;
     if (mKeyboardScancodeForNewMapping != LUS_KB_UNKNOWN) {
         mapping = std::make_shared<KeyboardKeyToButtonMapping>(mPortIndex, bitmask, mKeyboardScancodeForNewMapping);
-    }
-
-    if (mapping == nullptr) {
+    } else if (!ImGui::IsAnyItemHovered() && mMouseButtonForNewMapping != MouseBtn::MOUSE_BTN_UNKNOWN) {
+        // TODO: I dont think direct ImGui calls should be here
```
briaguya0: i was about to comment on `#include <imgui.h>` saying this exact thing, looks like you beat me to it!

src/controller/controldevice/controller/ControllerStick.cpp
```diff
@@ -272,9 +276,11 @@ bool ControllerStick::AddOrEditAxisDirectionMappingFromRawPress(Direction direct
     if (mKeyboardScancodeForNewMapping != LUS_KB_UNKNOWN) {
         mapping = std::make_shared<KeyboardKeyToAxisDirectionMapping>(mPortIndex, mStickIndex, direction,
                                                                       mKeyboardScancodeForNewMapping);
-    }
-
-    if (mapping == nullptr) {
+    } else if (!ImGui::IsAnyItemHovered() && mMouseButtonForNewMapping != MouseBtn::MOUSE_BTN_UNKNOWN) {
+        // TODO: I dont think direct ImGui calls should be here (again)
```
briaguya0: agreed (again)

src/controller/controldevice/controller/ControllerStick.cpp
```diff
@@ -319,9 +326,13 @@ void ControllerStick::UpdatePad(int8_t& x, int8_t& y) {
...
 bool ControllerStick::ProcessKeyboardEvent(KbEventType eventType, KbScancode scancode) {
-    if (mUseKeydownEventToCreateNewMapping && eventType == LUS_KB_EVENT_KEY_DOWN) {
-        mKeyboardScancodeForNewMapping = scancode;
-        return true;
+    if (mUseKeydownEventToCreateNewMapping) {
+        if (eventType == LUS_KB_EVENT_KEY_DOWN) {
+            mKeyboardScancodeForNewMapping = scancode;
+            return true;
+        } else {
+            mKeyboardScancodeForNewMapping = LUS_KB_UNKNOWN;
```
briaguya0: just curious: why does this need to get set every time now?

src/controller/controldevice/controller/ControllerStick.cpp
```diff
@@ -339,6 +350,31 @@ bool ControllerStick::ProcessKeyboardEvent(KbEventType eventType, KbScancode sca
     return result;
 }
 
+bool ControllerStick::ProcessMouseEvent(bool isPressed, MouseBtn button) {
+    if (mUseKeydownEventToCreateNewMapping) {
```
briaguya0: i feel like it would make sense to split this up instead of using the same flag for mouse button presses and keydowns

src/graphic/Fast3D/gfx_dxgi.cpp
```diff
@@ -364,29 +388,41 @@ static LRESULT CALLBACK gfx_dxgi_wnd_proc(HWND h_wnd, UINT message, WPARAM w_par
             onkeyup(w_param, l_param);
             break;
         case WM_LBUTTONDOWN:
+            on_mouse_button_down(0);
             dxgi.mouse_pressed[0] = true;
```
briaguya0: would it work to move the `dxgi.mouse_pressed[0] = true;` etc. lines into `on_mouse_button_down`/`on_mouse_button_up`?

src/graphic/Fast3D/gfx_dxgi.cpp
```diff
@@ -550,14 +591,19 @@ static bool gfx_dxgi_get_mouse_state(uint32_t btn) {
...
+        gfx_dxgi_set_cursor_visibility(true);
         ReleaseCapture();
     }
+    dxgi.is_mouse_captured = capture;
 }
 
 static bool gfx_dxgi_is_mouse_captured() {
-    return (GetCapture() != NULL);
+    return dxgi.is_mouse_captured;
 }
```
briaguya0, line 609: curious about this change, mostly wondering if there was a reason we didn't just use `dxgi.is_mouse_captured` before

briaguya0 (commented):
overall stuff is looking good! left a few comments (some you already had todos for)

i wasn't able to test this out locally (now that the resourcefilter stuff landed it seems the soh side of that will need to land before testing this is straightforward), but i was hoping to test out what the mapping experience - specifically what happens when this popup is visible

image

and someone clicks on (or misses when trying to click on) the cancel button

src/controller/controldeck/ControlDeck.cpp
lightmanLP: If you hold mouse down, `ActiveId` changes from zero and locks keyboard inputs

src/controller/controldevice/controller/ControllerStick.cpp
lightmanLP: I dont exactly know why, but:
```
somehow previously pressed key sometimes binds after clicking on binding button (how...?)
steps to reproduce:
1. press key to add new mapping
2. reopen menu via esc
3. open mapping window again
also works with clicking cancel button (click being buffered, lol)
```

lightmanLP: My assumtion was it just keeps something after, so I cleared it out

src/controller/controldeck/ControlDeck.cpp
briaguya0: so this is codifying "we want to block keyboard input when mouse is held down"? the previous logic here was "this is when we don't want to send keyboard input to the game" (if we're blocking all game input or if someone is typing in imgui)

src/graphic/Fast3D/gfx_dxgi.cpp
lightmanLP: This variable doest exist before, I updated capture implementation to match with sdl one

src/controller/controldevice/controller/ControllerStick.cpp
briaguya0: is this an issue you can repro on LUS without these mouse mapping additions? if so a quick tiny PR explaining the issue and the fix would be wonderful!

src/controller/controldeck/ControlDeck.cpp
lightmanLP: ImGui blocks in-game input via WantCaptureKeyboard when you holding down mouse button

lightmanLP: My changes replicates WantCaptureKeyboard, but without this block

src/graphic/Fast3D/gfx_dxgi.cpp
lightmanLP: Yeah, sure

src/controller/controldevice/controller/ControllerStick.cpp
lightmanLP: Tried it, reproducible on release version of 2s2h. I'll make pr

lightmanLP:
Cancel button click should work without any problems, you can test it here https://github.com/lightmanLP/2ship2harkinian/tree/mouse-mapping-test
P.S. slightly outdated because of release-version LUS

src/controller/controldevice/controller/ControllerStick.cpp
lightmanLP: It can create only one mapping anyway, so Im not sure how separate flag should work

src/controller/controldevice/controller/ControllerStick.cpp
lightmanLP: https://github.com/Kenix3/libultraship/pull/767

src/controller/controldevice/controller/ControllerStick.cpp
briaguya0: i guess just a rename would work then since mouse doesn't have keydown events

briaguya0:
> Cancel button click should work without any problems, you can test it here https://github.com/lightmanLP/2ship2harkinian/tree/mouse-mapping-test P.S. slightly outdated because of release-version LUS

the cancel button does seem to work as expected, but i am still somewhat concerned about always capturing mouse when reading input for mappings

accidental mouse inputs are decently common, so having some safeguards would be nice

one example i noticed when testing this out was if someone accidentally double clicks on an existing mapping it'll replace the mapping

https://github.com/user-attachments/assets/7ad05c7a-9fe7-4609-872d-871ad967cfca

i'd like to hear what others think about this (is it a problem we should worry about? if it's a problem we should worry about, what would a good solution be?)

src/controller/controldevice/controller/mapping/ControllerMapping.h
```diff
@@ -7,7 +7,8 @@ namespace Ship {
 
 #define MAPPING_TYPE_GAMEPAD 0
 #define MAPPING_TYPE_KEYBOARD 1
-#define MAPPING_TYPE_UNKNOWN 2
+#define MAPPING_TYPE_MOUSE 2
+#define MAPPING_TYPE_UNKNOWN 3
```
Kenix3: Unknown should probably be hard coded to something like -1

briaguya0: that's was definitely a me thing, i don't remember if we hit snags with that in the initial implementation or not

lightmanLP:
> one example i noticed when testing this out was if someone accidentally double clicks on an existing mapping it'll replace the mapping

It might be fixed by spawning binding popup offsetted from cursor and checking for cursor to be in popup boundaries for mouse binding

src/controller/controldevice/controller/mapping/ControllerMapping.h
lightmanLP: Seems like it wouldnt cause any issues now on 2ship and soh, so Im changing it

briaguya0:
> It might be fixed by spawning binding popup offsetted from cursor and checking for cursor to be in popup boundaries for mouse binding

that seems reasonable to me!

it doesn't solve the "tried to click the cancel button and missed" problem but that's probably fine - if people start reporting that as a bug once this ships we can look into other solutions

Kenix3 (approved):

lightmanLP:
> It might be fixed by spawning binding popup offsetted from cursor and checking for cursor to be in popup boundaries for mouse binding

Implemented
For testing purposes: https://github.com/lightmanLP/2ship2harkinian/actions/runs/12521779818

lightmanLP:
Made version of 2S2H with latest LUS and mouse mapping on top
https://github.com/HarbourMasters/2ship2harkinian/pull/910

briaguya0:
@lightmanLP I just tested the latest build from <https://github.com/lightmanLP/2ship2harkinian/actions/workflows/main.yml> (`387fb85`) and it seems like using the mouse in the popped out input editor is completely broken (but it's working perfectly in the `esc` menu)

is that something you've tested/been able to repro?

lightmanLP:
Just downloaded latest, everything seems fine for me. Can you explain what exactly happened?
Only issue with popped out editor that I bumped in always is mapping window doesn't work when it's outside the game window (and its known issue for keyboard, same for mouse).

briaguya0:
after trying again i was unable to reproduce the issue

briaguya0 (approved):
:shipit:

PR #744, opened 2024-12-12
Remove explicit void in functions with no parameters

### LINKED ISSUE #698: Drop explicit void when no parameters

Kenix3:
Resolves #698


briaguya0 (approved):
is this something clang-format can check for?

Kenix3:
> is this something clang-format can check for?

Good question. Format probably can.

Kenix3:
Actually doesn't appear to be supported:
https://github.com/microsoft/clang/blob/master/docs/ClangFormatStyleOptions.rst

Issue #736, opened 2024-12-07
[Feature Request] Async options for UnloadDirectory


Malkierian:
The file listing in the Directory commands is very process intensive, and "unloading" assets should never have to block the game loops, so an option to have that happen on a separate thread rather than on the main one would be nice to have.

briaguya0:
i'm curious about the

> file listing in the Directory commands is very process intensive

part of this

in `LoadDirectoryAsync` the file listing part isn't async

https://github.com/Kenix3/libultraship/blob/99d78d61380048bbcec2f8364f5d876681fd982e/src/resource/ResourceManager.cpp#L226-L239

so the following that pattern to "make unloading async" (just adding unloads to a queue instead of unloading on the same thread) wouldn't address the file listing issue

it's also probably worth looking into ways to make the `ListFiles` more performant as well

getting some measurements on 

https://github.com/Kenix3/libultraship/blob/99d78d61380048bbcec2f8364f5d876681fd982e/src/resource/archive/ArchiveManager.cpp#L79-L85

and

https://github.com/Kenix3/libultraship/blob/99d78d61380048bbcec2f8364f5d876681fd982e/src/resource/archive/ArchiveManager.cpp#L69-L77

would be good

if the performance issue is only in the version with a filter, we could look into adding (less robust but more performant) alternatives to glob (after reading https://research.swtch.com/glob i don't think we can get a much faster glob implementation than https://github.com/Kenix3/libultraship/blob/99d78d61380048bbcec2f8364f5d876681fd982e/src/utils/glob.c)

Malkierian:
I think the issue is much more prevalent in the filter version, yes.

PR #752, opened 2024-12-13
ResourceFilter, Directory Function Renames, Multifilter Functionality

Malkierian:
This PR creates a new struct, `ResourceFilter`, which holds include and exclude lists as well as owner and parent options to be passed to `LoadResource` as part of the `ResourceIdentifier`. It also renames and fleshes out all Directory commands to be named with Resources instead, and threads all Resources functionality to also encapsulate `ListFiles` calls, which addresses #736. `ListFiles` calls are unified now, eventually passing a ResourceFilter to the base function.

### DIFF
--- a/src/resource/ResourceManager.h
+++ b/src/resource/ResourceManager.h
@@ -85,11 +85,11 @@ class ResourceManager {
     LoadDirectoryAsync(const std::string& searchMask, BS::priority_t priority = BS::pr::normal);
     std::shared_ptr<std::vector<std::shared_future<std::shared_ptr<IResource>>>>
     LoadDirectoryAsyncWithExclude(const std::vector<std::string>& includeMasks,
-                                  const std::vector<std::string>& excludeMasks, uintptr_t owner = 0,
+                                  const std::vector<std::string>& excludeMasks,
                                   BS::priority_t priority = BS::pr::normal);
     void DirtyDirectory(const std::string& searchMask);
     void UnloadDirectoryWithExclude(const std::vector<std::string>& includeMasks,
-                                    const std::vector<std::string>& excludeMasks, uintptr_t owner);
+                                    const std::vector<std::string>& excludeMasks);
     void UnloadDirectory(const std::string& searchMask);
 
     bool OtrSignatureCheck(const char* fileName);
--- a/src/resource/ResourceManager.h
+++ b/src/resource/ResourceManager.h
@@ -18,13 +18,23 @@
 namespace Ship {
 struct File;
 
+struct ResourceFilter {
+    ResourceFilter(const std::list<std::string> includeMasks, const std::list<std::string> excludeMasks,
+        const uintptr_t owner, const std::shared_ptr<Archive> parent);
+
+    const std::list<std::string> IncludeMasks;
+    const std::list<std::string> ExcludeMasks;
+    const uintptr_t Owner = 0;
+    const std::shared_ptr<Archive> Parent = nullptr;
+};
+
 struct ResourceIdentifier {
-    friend class ResourceIdentifierHash;
+    friend struct ResourceIdentifierHash;
 
     ResourceIdentifier(const std::string& path, const uintptr_t owner, const std::shared_ptr<Archive> parent);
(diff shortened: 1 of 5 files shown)


src/resource/ResourceManager.h
```diff
@@ -18,13 +20,23 @@
 namespace Ship {
 struct File;
 
+struct ResourceFilter {
+    ResourceFilter(const std::list<std::string> includeMasks, const std::list<std::string> excludeMasks,
```
Kenix3: Can we not make the lists that are passed into the constructors here references and then copy into new containers on construct?

Malkierian: Not that I could figure out, that was the very issue that was causing that problem I was talking about on Discord last night where it would end up with garbage data in the containers while iterating through them in `ListFiles`.

Malkierian: Changing them to vector references didn't work, either, though I would have thought it would from the research I did on the issue saying that lists can have reference passing issues with non-contiguous memory, etc.

Kenix3: What about const pointers instead? Raw pointers.

Kenix3: LUS would have to copy anyway... I don't think we want LUS relying on memory blocks allocated by the caller.

In this setup the optimizer _should_ be able to see the double copy and chop one out. I can see why it's extra rough in debug build.

Malkierian: Somehow, no, that doesn't work either. This is an example of what's going on with it, when this is supposed to be a list with a size of 4
image

Malkierian: Ah, I see now what you were talking about before, and I realize I didn't copy the `ResourceIdentifier` precedent exactly. Yes, that works just fine. So if they're not references in the arguments, does that mean they would be copied twice or something? What's the benefit to having references passed in but copied to internal storage?

src/resource/archive/ArchiveManager.cpp
```diff
@@ -64,10 +64,41 @@ bool ArchiveManager::HasFile(uint64_t hash) {
...
 }
 
-std::shared_ptr<std::vector<std::string>> ArchiveManager::ListFiles(const std::string& filter) {
+std::shared_ptr<std::vector<std::string>> ArchiveManager::ListFiles() {
+    return ListFiles({}, {});
+}
+
+std::shared_ptr<std::vector<std::string>> ArchiveManager::ListFiles(const std::string& searchMask) {
+    return ListFiles({ searchMask }, {});
+}
```
briaguya0: do we need both of these or could the line in `ArchiveManager.h` just change? 
```diff
- std::shared_ptr<std::vector<std::string>> ListFiles(const std::string& searchMask);
+ std::shared_ptr<std::vector<std::string>> ListFiles(const std::string& searchMask = "");
```

src/resource/ResourceManager.h
Kenix3: Yeah it will copy once into the function, and then it will copy a second time to make one inside of LUS. This is the literal code. I _hope_ release optimizations handle this.

src/resource/archive/ArchiveManager.cpp
Malkierian: Doing so would require me to change how the main function checks for non-existent filters, as a list initialized with `""` I believe doesn't count for `List::empty()`. Or, I guess, add a local initialization to be able to choose between feeding searchMask to it or not based on if it's `""`

briaguya0: i think i'm just imagining times when a port might do something like
```cpp
std::string mySearchMask = "";
if (customLogic) {
  mySearchMask += customLogicPath;
}
ListFiles(mySearchMask);
```
so handling the case where `customLogic` is false there the same as someone just calling `ListFiles()` seems reasonable

Malkierian: Eh, I guess it's easy enough to change the way the searchMask function passes the searchMask to the main one so that it doesn't add a size 0 string to the list before passing it.

briaguya0 (approved):
LGTM

PR #753, opened 2024-12-13
make `Context` not need `Fast3dWindow`

### LINKED ISSUE #577: Do not require Context to use Fast3dWindow.
How are we going to accomplish this since the Window is an abstract base class?
Basically the idea is to pass a Window in for the Context constructor, and the createInstance method.

briaguya0:
resolves https://github.com/Kenix3/libultraship/issues/577

soh testing pr: https://github.com/HarbourMasters/Shipwright/pull/4682

### DIFF
--- a/src/Context.cpp
+++ b/src/Context.cpp
@@ -74,23 +74,16 @@ Context::Context(std::string name, std::string shortName, std::string configFile
     : mConfigFilePath(std::move(configFilePath)), mName(std::move(name)), mShortName(std::move(shortName)) {
 }
 
-void Context::Init(const std::vector<std::string>& archivePaths, const std::unordered_set<uint32_t>& validHashes,
+bool Context::Init(const std::vector<std::string>& archivePaths, const std::unordered_set<uint32_t>& validHashes,
                    uint32_t reservedThreadCount, AudioSettings audioSettings, std::shared_ptr<Window> window) {
-    InitLogging();
-    InitConfiguration();
-    InitConsoleVariables();
-    InitResourceManager(archivePaths, validHashes, reservedThreadCount);
-    InitControlDeck();
-    InitCrashHandler();
-    InitConsole();
-    InitWindow(window);
-    InitAudio(audioSettings);
-    InitGfxDebugger();
+    return InitLogging() && InitConfiguration() && InitConsoleVariables() &&
+           InitResourceManager(archivePaths, validHashes, reservedThreadCount) && InitControlDeck() &&
+           InitCrashHandler() && InitConsole() && InitWindow(window) && InitAudio(audioSettings) && InitGfxDebugger();
 }
 
-void Context::InitLogging() {
+bool Context::InitLogging() {
     if (GetLogger() != nullptr) {
-        return;
+        return true;
     }
 
     try {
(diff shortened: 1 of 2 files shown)


src/Context.cpp
```diff
@@ -255,12 +254,17 @@ void Context::InitConsole() {
...
-void Context::InitWindow(std::vector<std::shared_ptr<GuiWindow>> guiWindows) {
+void Context::InitWindow(std::shared_ptr<Window> window) {
     if (GetWindow() != nullptr) {
         return;
     }
 
-    mWindow = std::make_shared<Fast::Fast3dWindow>(guiWindows);
+    if (window == nullptr) {
+        // todo: better error - quit?
+        SPDLOG_ERROR("CONTEXT NEEDS WINDOW BADLY");
```
briaguya0: this is why it's a draft

Kenix3: We definitely want to clean up the message here.

src/Context.cpp
```diff
@@ -42,15 +41,14 @@ Context::~Context() {
     spdlog::shutdown();
 }
 
-std::shared_ptr<Context> Context::CreateInstance(const std::string name, const std::string shortName,
-                                                 const std::string configFilePath,
-                                                 const std::vector<std::string>& archivePaths,
-                                                 const std::unordered_set<uint32_t>& validHashes,
-                                                 uint32_t reservedThreadCount, AudioSettings audioSettings) {
+std::shared_ptr<Context>
```
Kenix3, line 44: We can check for missing objects in this function and return a nullptr for the Context. We could then have the game check for existence (maybe create a C bridge that returns a book?). If not exist then it's on the game to exit itself since the game will always control `main()`

briaguya0: hmm, that doesn't handle the `CreateUninitializedInstance` use case though. if a consumer uses `CreateUninitializedInstance` then calls a few `Init*`s for specific things, then calls `Init` they could still end up with a broken context

Kenix3: That's a fair rebuttal, but I think the initial thought is the direction we would want to go. Do we return true/false anywhere on whether or not the `Context` initialized successfully after `Init*`

briaguya0: not currently, all the `Init`s are void

https://github.com/Kenix3/libultraship/blob/be703a997b5eb189188ec036223a2238f0c10328/src/Context.h#L40-L41

https://github.com/Kenix3/libultraship/blob/be703a997b5eb189188ec036223a2238f0c10328/src/Context.h#L57-L67

Kenix3: Do you think it'd be too kitchen-sinky to correct that and have it return a success/failure state? Then it can bubble up more.
I think it is desired functionality to be able to get a success/failure state out of the Context.

briaguya0: i don't think that'd be too bad, we could just update

https://github.com/Kenix3/libultraship/blob/be703a997b5eb189188ec036223a2238f0c10328/src/Context.cpp#L79-L91

to be 

```cpp
void Context::Init(const std::vector<std::string>& archivePaths, const std::unordered_set<uint32_t>& validHashes,
                   uint32_t reservedThreadCount, AudioSettings audioSettings) {
    return InitLogging() && InitConfiguration() ...etc;
}
```

do we have a good pattern of returning success/fail somewhere? i'd prefer not to hardcode `0`/`1` if possible

Kenix3: https://github.com/Kenix3/libultraship/blob/be703a997b5eb189188ec036223a2238f0c10328/src/public/bridge/resourcebridge.cpp#L37-L39
This here returns a 0 if we can't do the thing. I'd be perfectly happy with bool and true/false.

briaguya0: hmm, but that's 0 for size...

i guess true/false is fine - just feels a little odd since it's backwards from exit codes where 0 is good (and returning `false` to mean everything initialized successfully feels odd too)

Kenix3: I'm good with that. This is a class initialization. You could have the classes return true/false for `did init` and then the static function swap the return codes since it's the "entry point"

briaguya0: cool `true` for init success it is then, i'll get on it

src/Context.cpp
```diff
@@ -42,15 +41,14 @@ Context::~Context() {
...
+Context::CreateInstance(const std::string name, const std::string shortName, const std::string configFilePath,
+                        const std::vector<std::string>& archivePaths, const std::unordered_set<uint32_t>& validHashes,
+                        uint32_t reservedThreadCount, AudioSettings audioSettings, std::shared_ptr<Window> window) {
     if (mContext.expired()) {
         auto shared = std::make_shared<Context>(name, shortName, configFilePath);
         mContext = shared;
-        shared->Init(archivePaths, validHashes, reservedThreadCount, audioSettings);
+        shared->Init(archivePaths, validHashes, reservedThreadCount, audioSettings, window);
```
Kenix3: Here you should probably return nullptr on failure to init?

PR #755, opened 2024-12-14
shared pointers for `Get/SetArchives`

### LINKED ISSUE #714: change `GetArchives` to return shared pointer to copy of vector
https://github.com/Kenix3/libultraship/blob/00c6bcc88286173ce22b8a6656373d3aa5830c62/src/resource/archive/ArchiveManager.h#L24

follow pattern from

https://github.com/Kenix3/libultraship/blob/00c6bcc88286173ce22b8a6656373d3aa5830c62/src/resource/archive/ArchiveManager.cpp#L79-L85

### LINKED ISSUE #715: update `SetArchives` to take shared pointer instead of reference
https://github.com/Kenix3/libultraship/blob/00c6bcc88286173ce22b8a6656373d3aa5830c62/src/resource/archive/ArchiveManager.h#L25

briaguya0:
resolves https://github.com/Kenix3/libultraship/issues/714
resolves https://github.com/Kenix3/libultraship/issues/715

### DIFF
--- a/src/resource/archive/ArchiveManager.cpp
+++ b/src/resource/archive/ArchiveManager.cpp
@@ -31,7 +31,7 @@ void ArchiveManager::Init(const std::vector<std::string>& archivePaths,
 
 ArchiveManager::~ArchiveManager() {
     SPDLOG_TRACE("destruct archive manager");
-    SetArchives({});
+    SetArchives(nullptr);
 }
 
 bool ArchiveManager::IsLoaded() {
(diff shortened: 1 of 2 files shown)


src/resource/archive/ArchiveManager.cpp
```diff
@@ -118,8 +122,15 @@ size_t ArchiveManager::RemoveArchive(std::shared_ptr<Archive> archive) {
...
-void ArchiveManager::SetArchives(const std::vector<std::shared_ptr<Archive>>& archives) {
-    mArchives = archives;
+void ArchiveManager::SetArchives(std::shared_ptr<std::vector<std::shared_ptr<Archive>>> archives) {
+    mArchives.clear();
+
+    if (archives) {
+        for (const auto& archive : *archives) {
+            mArchives.push_back(archive);
+        }
+    }
```
briaguya0, line 132: @Kenix3 is this what you were thinking here?

Kenix3: This looks exactly like what we were thinking. It looks like we decided to pass lists around into and out of LUS as shared pointers.

Kenix3 (approved):
Ultimately I would like to standardize on how we pass containers around. Would like to discuss if shared pointers or const references is better. For now, this does the work of getting us to consistency even if we don't have the final solution.

Issue #580, opened 2024-05-09
Class headers should be included via the `"ship/"` or `"libultraship/"` directory.


Kenix3:
Most of this should be able to be done in cmake. This can be done after the libultraship/shipdk split.

briaguya0:
Steps:
1. Restructure `/src` directory to have `fast`/`lus`/`ship` subdirectories, put appropriate files in each spot
2. Move all headers out of `/src` into `/include` mirroring the directory structure of `/src`

briaguya0:
blocked by https://github.com/Kenix3/libultraship/issues/591

briaguya0:
https://github.com/Kenix3/libultraship/blob/main/src/resource/ResourceType.h needs to turn into 2 files

PR #757, opened 2024-12-14
move resources out of LUS namespace

### LINKED ISSUE #591: Remove most LUS resource types/factories
* only keep the ones used by fast3d
* move the ones fast3d needs to `Fast::` namespace

briaguya0:
resolves https://github.com/Kenix3/libultraship/issues/591

otrexporter: https://github.com/HarbourMasters/OTRExporter/pull/30
soh: https://github.com/HarbourMasters/Shipwright/pull/4685

### DIFF
--- a/src/graphic/Fast3D/gfx_pc.cpp
+++ b/src/graphic/Fast3D/gfx_pc.cpp
@@ -890,15 +890,15 @@ static void import_texture_raw(int tile, bool importReplacement) {
 
     uint16_t width = metadata->width;
     uint16_t height = metadata->height;
-    LUS::TextureType type = metadata->type;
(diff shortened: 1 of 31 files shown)


Kenix3:
Do we think this should come with moving the files as well? I'm good either way, but include paths will change when they are moved into the Fast3d system that we intend to split out completely during 2.x.x

briaguya0:
> Do we think this should come with moving the files as well? I'm good either way, but include paths will change when they are moved into the Fast3d system that we intend to split out completely during 2.x.x

i figured moving files around could all happen in https://github.com/Kenix3/libultraship/issues/580, that way we only have one giant "import everything from new places" change instead of 2 waves of it

Kenix3 (approved):
Sounds good to me!

PR #758, opened 2024-12-14
split `ControlDeck` between `Ship::` and `LUS::`

### LINKED ISSUE #706: split `ControlDeck` between ship and LUS
anything that touches

https://github.com/Kenix3/libultraship/blob/00c6bcc88286173ce22b8a6656373d3aa5830c62/src/controller/controldeck/ControlDeck.h#L35-L36

should move to the LUS version that inherits from the ship version

briaguya0:
resolves https://github.com/Kenix3/libultraship/issues/706

soh https://github.com/HarbourMasters/Shipwright/pull/4687


src/Context.cpp
```diff
@@ -218,7 +218,7 @@ void Context::InitControlDeck(std::vector<CONTROLLERBUTTONS_T> additionalBitmask
         return;
     }
 
-    mControlDeck = std::make_shared<ControlDeck>(additionalBitmasks);
+    mControlDeck = std::make_shared<LUS::ControlDeck>(additionalBitmasks);
```
briaguya0: reason this is in draft

Kenix3: Hmm. What is the additional bitmasks doing again?

We could push this to the derived class. If we do that, we want some kind of internal overload able init function that's call from the main one.

briaguya0: we don't have a derived context

Kenix3: I meant for the ControlDeck class.

briaguya0: even without `additionalBitmasks` we can't make an instance of an abstract class
```
error: invalid new-expression of abstract class type ‘Ship::ControlDeck’
```

Kenix3 (approved):

PR #761, opened 2024-12-16
Fix missing inputs on Metal shader for blended textures

Archez:
These two lines in the Metal shader for blended textures had mismatched sprintf format chars and input args. Crash was not observed on Debug builds because UB resolved as `0`, but in Release the UB could be any value.


briaguya0 (approved):

briaguya0:
approved, but just want to confirm that it should be `i` instead of `0` (you mentioned on debug it was always `0`)

Archez:
Correct, `i` is for the tile ID. Regarding blended textures, those all have coincidentally been on tile 0 in practice which is why the issue wasn't observed in Debug.

PR #776, opened 2024-12-29
optimise import texture

coco875:
don't know how to make gfx_table.cpp compile instead of include it but else it improve performance of import texture in general
before modification:
[starship_bench_old_ci.txt](https://github.com/user-attachments/files/18268033/starship_bench_old_ci.txt)

after modification:
[starship_bench.txt](https://github.com/user-attachments/files/18268034/starship_bench.txt)


shizukachan:
This code really shouldn't be in the hot path, as libultraship has a texture cache that normally avoids re-importing textures. However, since Starship does per-frame whole-texture-cache invalidation in its fox_std_lib.c file when scroll and mottle effects are triggered, the texture caching doesn't work and we have to re-import the textures.

You can stub the invalidate function (eg. make it return immediately without doing the invalidate) to see other (relatively) low hanging fruit. This will break all scroll/mottle effects in Starship, but continue allowing performance onion peeling in libultraship.

coco875:
then in this case this will fall in "micro optimisation" and I think it can be useful to other port who also invalidate cache, or just import texture

coco875:
> This code really shouldn't be in the hot path, as libultraship has a texture cache that normally avoids re-importing textures. However, since Starship does per-frame whole-texture-cache invalidation in its fox_std_lib.c file when scroll and mottle effects are triggered, the texture caching doesn't work and we have to re-import the textures.
> 
> You can stub the invalidate function (eg. make it return immediately without doing the invalidate) to see other (relatively) low hanging fruit. This will break all scroll/mottle effects in Starship, but continue allowing performance onion peeling in libultraship.

you are right starship have now avoid to stop to invalid all texture frequently, and the performance is much better, don't know if it's still worst have big clunky table for convertion.

PR #785, opened 2025-01-07
Allows LUS to set the screen dimensions programmatically.

Kenix3:


src/graphic/Fast3D/Fast3dWindow.cpp
```diff
@@ -62,13 +62,13 @@ void Fast3dWindow::Init() {
     uint32_t width, height;
     int32_t posX, posY;
 
-    isFullscreen = Ship::Context::GetInstance()->GetConfig()->GetBool("Window.Fullscreen.Enabled", false) || gameMode;
+    isFullscreen = Ship::Context::GetInstance()->GetConfig()->GetBool("Window.Fullscreen.Enabled", 0) || gameMode;
```
briaguya0: is the plan to standardize on `0` instead of `false` for `GetBool`?

src/graphic/Fast3D/Fast3dWindow.cpp
```diff
@@ -62,13 +62,13 @@ void Fast3dWindow::Init() {
...
     posX = Ship::Context::GetInstance()->GetConfig()->GetInt("Window.PositionX", 100);
     posY = Ship::Context::GetInstance()->GetConfig()->GetInt("Window.PositionY", 100);
 
     if (isFullscreen) {
-        width = Ship::Context::GetInstance()->GetConfig()->GetInt("Window.Fullscreen.Width", gameMode ? 1280 : 1920);
-        height = Ship::Context::GetInstance()->GetConfig()->GetInt("Window.Fullscreen.Height", gameMode ? 800 : 1080);
+        width = Ship::Context::GetInstance()->GetConfig()->GetInt("Window.Fullscreen.Width", 1920);
```
briaguya0: need to test this in gamescope. i still don't like that we have resolution tied to gamescope, just because someone is using gamescope doesn't mean they're on a steam deck, but seeing what breaks and if we have ways around it with the other changes in this PR will be good

edit: looks like we're looking for `steamdeck` in the release variant id so it's not _just_ gamescope we're checking for, but still feels a bit tacked on

Kenix3:
https://github.com/Kenix3/libultraship/pull/1097

PR #789, opened 2025-01-10
fix linux build validation

briaguya0:
as of writing, the `ubuntu-latest` gh runner image is now `ubuntu-24.04` https://github.com/actions/runner-images/blob/976232da217887825e7541c2635bf39cbbf22654/README.md?plain=1#L23

gh runners support the latest 3 versions of gcc https://github.com/actions/runner-images/blob/976232da217887825e7541c2635bf39cbbf22654/README.md?plain=1#L124

the latest available in `24.04` is `14`, meaning our options are `12`, `13`, and `14`

i bumped it to `12` - if we'd rather validate builds using `13` or `14` i'm happy to look into that


briaguya0:
image

PR #790, opened 2025-01-10
Move init data logic to resource loader

coco875:


briaguya0:
it looks like this has some conflicts. would you be able to resolve those and make a testing PR so we can verify these changes in a released port?

coco875:
ok, will probably do this tomorrow

coco875:
the PR: https://github.com/HarbourMasters/Starship/pull/170

PR #791, opened 2025-01-16
Edit reset hotkey to support Mac after ImGui 1.91

Archez:
ImGui 1.91 changed behavior of how Super and Ctrl work on MacOS by swapping the keys for various interactions. (e.g. cmd+r for reset), this means we can drop the ifdef for Mac regarding the alternate reset hotkey.

### DIFF
--- a/src/window/gui/Gui.cpp
+++ b/src/window/gui/Gui.cpp
@@ -504,21 +504,13 @@ void Gui::DrawMenu() {
         }
     }
 
-#if __APPLE__
-    if ((ImGui::IsKeyDown(ImGuiKey_LeftSuper) || ImGui::IsKeyDown(ImGuiKey_RightSuper)) &&
-        ImGui::IsKeyPressed(ImGuiKey_R, false)) {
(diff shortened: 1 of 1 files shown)


briaguya0:
how does this compare to turning `ConfigMacOSXBehavior` off?

briaguya0 (approved):

Archez:
> how does this compare to turning `ConfigMacOSXBehavior` off?

The first iteration of this PR attempted that. Although that did make the original hotkey behavior work again on Mac, it also changed a couple other behaviors in ImGui such as "Select all" changing Cmd+A -> Ctrl+A, "Copy/Cut/Paste" changing Cmd+C/X/V ->Ctrl+C/X/V, the hot key for tabbing between imgui windows, and a couple other things. These are all undesirable for a Mac user, so instead switching our hotkey for Cmd+R to use the same Ctrl+R one as windows allows ImGui to redirect that back to Cmd+R behind the scenes.

Kenix3 (approved):

PR #793, opened 2025-01-19
simplify controllers

### LINKED ISSUE #792: simplify controllers
we can narrow down the issues people are having with the controller rework to 2 things:
* mappings are associated to a specific device (N64 A -> "xbox controller that is currently SDL0" A) 
  * this also means we have to maintain some fancy logic to maintain those connections
* OS support (especially windows) for controllers is trash
  * we can't trust that devices send the correct number of connect/disconnect events
  * we can't trust that we don't get duplicates of any given controller
  * basically we just can't trust that we have reasonable information about what controllers exist/are connected at any given time

so while the current controller system provides awesome flexibility when everything is working as expected, it fails a lot harder than simpler systems do when things fall apart (which seems to be the rule not the exception on windows)

given all of this - i think the most reasonable solution is to simplify things in the following ways:
* update all SDL mappings to not have an associated SDL device
* set up many:many for n64 controller port to sdl device

from a user perspective this means:
* you can no longer have different mappings from different devices on one n64 port - no "A on my xbox controller and triangle on my playstation controller" - it'd just be both SOUTH and NORTH on both
* no more listing disconnected devices
* where we currently show the list of devices (connected and disconnected) - we can just have toggles to enable/disable devices on each port

what i haven't figured out yet:
* rumble strength saving per device (is it a problem if we lose this?)
* analog deadzone/sensitivity per device (is it a problem if we lose this?)
* LED settings per device (and how to handle devices that support/don't support it)
* gyro settings per device (and how to handle devices that support/don't support it)

i think these are all solveable problems. this should allow us to remove quite a bit of supporting code, and should make the problems people face with controllers a lot less painful

briaguya0:
resolves https://github.com/Kenix3/libultraship/issues/792

soh testing pr:
* https://github.com/HarbourMasters/Shipwright/pull/4914

### `//todo`
- [ ] decide if there's a better way to handle the little vector thing https://github.com/Kenix3/libultraship/pull/793#discussion_r1921640280
- [x] filter devices by port https://github.com/Kenix3/libultraship/pull/793#discussion_r1921640637
  * i think https://github.com/Kenix3/libultraship/pull/793/commits/96e7d20d44d8013983d3633b67aa0189793d90ca is good enough for now - once we start trying to account for edge cases we add complexity to what should be happy paths. i'm trying to avoid that because that's how we got to the point of needing the massive simplification in the first place.
- [x] actually test the built-in input editor window (i've made sure all the changes i made to the soh one made it over but i've only tested the soh input editor window)
  - works as expected


src/controller/controldevice/controller/mapping/sdl/SDLAxisDirectionToAxisDirectionMapping.cpp
```diff
@@ -8,39 +8,46 @@
...
+
+        if ((mAxisDirection == POSITIVE && axisValue < 0) || (mAxisDirection == NEGATIVE && axisValue > 0)) {
+            normalizedValues.push_back(0.0f);
+            continue;
+        }
+
+        // scale {-32768 ... +32767} to {-MAX_AXIS_RANGE ... +MAX_AXIS_RANGE}
+        // and use the absolute value of it
+        normalizedValues.push_back(fabs(axisValue * MAX_AXIS_RANGE / MAX_SDL_RANGE));
+    }
```
briaguya0, line 40: not sure what to do about this one

src/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
```diff
@@ -0,0 +1,45 @@
...
+}
+
+ConnectedPhysicalDeviceManager::~ConnectedPhysicalDeviceManager() {
+}
+
+std::unordered_map<int32_t, SDL_GameController*>
+ConnectedPhysicalDeviceManager::GetConnectedSDLGamepadsForPort(uint8_t portIndex) {
+    // todo: filter somehow
+    return mConnectedSDLGamepads;
+}
```
briaguya0: this is the big todo - now that mappings aren't tied to devices this means that if you have mappings on all the ports every controller will send inputs to every port

i'd like to find a way to handle this without introducing the type of complexity that led to the issues we have with the pre-simplified system

briaguya0: calling this one good as of https://github.com/Kenix3/libultraship/commit/96e7d20d44d8013983d3633b67aa0189793d90ca

Kenix3 (approved):

PR #794, opened 2025-01-20
After patch update cache fix

lightmanLP:
When `FetchContent` updates dependency from git, it simply keeps patched changes as unstaged ones. It is fine in case when patch is compatible with updated patch, but otherwise the build will fail until the cache is cleared from unstaged changes.
Thats why I added additional check to remove unstaged changes and reapply patch, if there is a problem.


Archez:
This won't work because the patch command is run directly in a terminal shell context. Parenthesis don't work for order of operations in shell execution.

lightmanLP:
Remade it into cmake script (Im not that good with cmake, so it might be a little sketchy)

Kenix3 (approved):

PR #798, opened 2025-01-25
Raw mouse input support

lightmanLP:
Mouse capture uses raw input now.
P.S.: also fixes mouse delta inconsistency with opengl, when mouse capture is disabled

Test build:
https://github.com/lightmanLP/2ship2harkinian/tree/raw-mouse-input-test
https://github.com/lightmanLP/2ship2harkinian/actions/runs/12960612950

co-authored-by: @lilacLunatic

### DIFF
--- a/src/graphic/Fast3D/gfx_dxgi.cpp
+++ b/src/graphic/Fast3D/gfx_dxgi.cpp
@@ -1,4 +1,4 @@
-#if defined(ENABLE_DX11) || defined(ENABLE_DX12)
+#if defined(ENABLE_DX11) || defined(ENABLE_DX12) || true
 
 #include <stdint.h>
 #include <math.h>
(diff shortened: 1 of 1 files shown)


MegaMech:
Tested on [content deleted] using DX, and OpenGL with SetMouseCapture on/off, and it works

MegaMech (approved):

briaguya0:
the PR description has

> P.S.: also fixes mouse delta inconsistency with opengl, when mouse capture is disabled

but the only file changed in this PR is `gfx_dxgi.cpp`. is that change in a different PR?

lightmanLP:
I meant that current DX implementation is inconsistent with OpenGL one, because DX implementation should also return mouse delta, when mouse capture is disabled.
So the fix only required changes on the DX side.

PR #801, opened 2025-01-27
Fix ListFiles(string) causing non-matches for calls to list all

Malkierian:
I made two mistakes with the previous code, where I was checking for .size() > 0 on a blank string instead of !empty(), and passing in an initializer with the original string argument rather than the constructed list that ignores blank strings.


briaguya0 (approved):
:shipit:

PR #805, opened 2025-01-29
remove unordered_map for attr handler

coco875:
change ucode_attr_handlers to an array of array instead of map to no longer compute hash.
originally the code was just made to make it work without taking in account performance or cost of an hastable.

profile make in debug mode. (profiler make with callgrind use callgrind_annotate or kcachegrind to visualise it)
before: 
[old_bench.txt](https://github.com/user-attachments/files/19096440/old_bench.txt)
image

after:
[bench.txt](https://github.com/user-attachments/files/19096553/bench.txt)
image

### DIFF
--- a/src/graphic/Fast3D/gfx_pc.cpp
+++ b/src/graphic/Fast3D/gfx_pc.cpp
@@ -148,19 +148,14 @@ static map<string, MaskedTextureEntry> masked_textures;
 
 static UcodeHandlers ucode_handler_index = ucode_f3dex2;
 
-const static std::unordered_map<Attribute, std::any> f3dex2AttrHandler = {
-    { MTX_PROJECTION, F3DEX2_G_MTX_PROJECTION }, { MTX_LOAD, F3DEX2_G_MTX_LOAD },     { MTX_PUSH, F3DEX2_G_MTX_PUSH },
-    { MTX_NOPUSH, F3DEX_G_MTX_NOPUSH },          { CULL_FRONT, F3DEX2_G_CULL_FRONT }, { CULL_BACK, F3DEX2_G_CULL_BACK },
-    { CULL_BOTH, F3DEX2_G_CULL_BOTH },
+const static uint32_t f3dex2AttrHandler[] = {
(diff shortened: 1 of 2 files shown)


briaguya0:
could you please provide a PR description? i know there were discussions about the performance implications of this change, it'd be good to have those documented as part of this PR

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -171,10 +166,10 @@ static constexpr std::array ucode_attr_handlers = {
     &f3dex2AttrHandler, // ucode_s2dex
 };
 
-template <typename T> static constexpr T get_attr(Attribute attr) {
+static uint32_t get_attr(Attribute attr) {
     const auto ucode_map = ucode_attr_handlers[ucode_handler_index];
-    assert(ucode_map->contains(attr) && "Attribute not found in the current ucode handler");
-    return std::any_cast<T>(ucode_map->at(attr));
+    // assert(ucode_map->contains(attr) && "Attribute not found in the current ucode handler");
```
briaguya0, line 171: why is this commented out?

coco875: ucode_map are no longer a map so we no longer need to do the assert.

coco875:
> i know there were discussions about the performance implications of this change, it'd be good to have those documented as part of this PR

what do you mean by document it ? because it's just change an unordored_map<int, int> where the key are an enum with limited range and static so I just change it to use instead an array so not so much change.

briaguya0:
> what do you mean by document it ? 

the *what* aspect of this PR is clear, but the *why* is not documented.

i remember conversations about the *why* relating to performance. it'd be good to document performance gains from this change and weigh them against the original reasoning behind the architectural decision to use a map (it looks like these maps were added in https://github.com/Kenix3/libultraship/pull/557/ and updated in https://github.com/Kenix3/libultraship/pull/567/ - @KiritoDv do you have thoughts on this one?)

coco875:
originally it use an unordered_map because @KiritoDv want just to make it work, without really caring about performance, by changing to just an array to allow to access more directly to the ressource (so avoid all process of an hash table). I will ping you in the private channel where we discuss about that.

Kenix3 (approved):

PR #806, opened 2025-01-29
Update gfx_sdl2.cpp to fix macOS Metal Stutter/Slowdown.

ReddestDream:
Fixes stutter/inability to lock 60 consistently on macOS using Metal. May also provide performance benefits under OpenGL and Linux. This is the weird 57-58 fps issue noticed on Metal graphics on Macs for a long while.

Thanks to @Spodi, @Malkierian, ProxySaw, and Fenrir for their help on Discord getting this figured out! :D

### DIFF
--- a/src/graphic/Fast3D/gfx_sdl2.cpp
+++ b/src/graphic/Fast3D/gfx_sdl2.cpp
@@ -615,10 +615,8 @@ static inline void sync_framerate_with_timer() {
 
     const int64_t next = previous_time + 10 * FRAME_INTERVAL_US_NUMERATOR / FRAME_INTERVAL_US_DENOMINATOR;
     int64_t left = next - t;
-#ifdef _WIN32
     // We want to exit a bit early, so we can busy-wait the rest to never miss the deadline
     left -= 15000UL;
-#endif
     if (left > 0) {
 #ifndef _WIN32
         const timespec spec = { 0, left * 100 };
@@ -632,13 +630,15 @@ static inline void sync_framerate_with_timer() {
 #endif
     }
 
-#ifdef _WIN32
     t = qpc_to_100ns(SDL_GetPerformanceCounter());
     while (t < next) {
-        YieldProcessor(); // TODO: Find a way for other compilers, OSes and architectures
+#ifdef _WIN32
+        YieldProcessor();
+#else
+        sched_yield();
+#endif
         t = qpc_to_100ns(SDL_GetPerformanceCounter());
     }
-#endif
     t = qpc_to_100ns(SDL_GetPerformanceCounter());
     if (left > 0 && t - next < 10000) {
         // In case it takes some time for the application to wake up after sleep,


larsy1995:
I can confirm that this one works.  Tested it a fair bit, game runs at a smooth fps except during loading/compilation stutters which is expected as of now.

ReddestDream:
Yeah. This does not fix shader compile stutters unfortunately. This is just for the main slowdown/stutter issue.

larsy1995:
Also, fyi, you've @ a random person instead of me, which is fair since you didn't know my GitHub account.

src/graphic/Fast3D/gfx_sdl2.cpp
```diff
@@ -615,10 +615,8 @@ static inline void sync_framerate_with_timer() {
 
     const int64_t next = previous_time + 10 * FRAME_INTERVAL_US_NUMERATOR / FRAME_INTERVAL_US_DENOMINATOR;
     int64_t left = next - t;
-#ifdef _WIN32
```
briaguya0, line 618: just trying to track down this change to see if it makes sense to have this on all platforms

```cpp
#ifdef _WIN32
    // We want to exit a bit early, so we can busy-wait the rest to never miss the deadline
    left -= 15000UL;
#endif
```
was added in https://github.com/Kenix3/libultraship/pull/330

blame prior to that was https://github.com/Kenix3/libultraship/commit/d600f1088e21a88040ce0e1e404b58a93c404ca9#diff-d1ca6db8c1db9fb811a90932b7e7e9d4d4bbea0b345cd67c4c73ea5c25a54628

```diff
-    t = SDL_GetPerformanceCounter();
+    t = qpc_to_100ns(SDL_GetPerformanceCounter());

-    const int64_t next = qpc_to_100ns(previous_time) + 10 * FRAME_INTERVAL_US_NUMERATOR / FRAME_INTERVAL_US_DENOMINATOR;
-    const int64_t left = next - qpc_to_100ns(t);
+    const int64_t next = previous_time + 10 * FRAME_INTERVAL_US_NUMERATOR / FRAME_INTERVAL_US_DENOMINATOR;
+    const int64_t left = next - t;
```

briaguya0 (commented):
it looks like this changes behavior not just on mac but also on linux, which leads into my main question of: where has this been tested?

linking a testing PR on a port would be great.

src/graphic/Fast3D/gfx_sdl2.cpp
ghost: Only tested on an M1 Air, macOS 15.2 so far. Will test on Steam Deck soon.

ghost: Why am I on my school account suddenly?…

larsy1995: I can also test it on Windows if that is preferable?

ReddestDream:
> it looks like this changes behavior not just on mac but also on linux, which leads into my main question of: where has this been tested?

It's been tested on Apple Silicon Macs. My personal M1 MBA and that of another person, Fenrir, on Discord (who is also [larsy1995](https://github.com/larsy1995)).

I'm hoping to test on Linux later, but if anyone can help me test, it would be appreciated. My understanding is that  sched_yield() is the correct POSIX equivalent for YieldProcessor, so that should work there as well.

>   left -= 15000UL;

Regarding this, @Spodi suggested this was also necessary to go along with the first part. I agree it also needs testing outside of just Apple Silicon Macs.

larsy1995:
Here’s the build with those changes based on my Game Mode PR: 
https://github.com/larsy1995/Shipwright/actions/runs/13037943737

ReddestDream:
> Here’s the build with those changes based on my Game Mode PR: https://github.com/larsy1995/Shipwright/actions/runs/13037943737

If anyone can help test this build, particularly on Linux and Intel Macs, it would be greatly appreciated. There is no expected behavior change on Windows, but confirmation is always appreciated. Thanks!

src/graphic/Fast3D/gfx_sdl2.cpp
```diff
@@ -632,13 +630,15 @@ static inline void sync_framerate_with_timer() {
...
-        YieldProcessor(); // TODO: Find a way for other compilers, OSes and architectures
+#ifdef _WIN32
+        YieldProcessor();
+#else
+        sched_yield();
+#endif
         t = qpc_to_100ns(SDL_GetPerformanceCounter());
     }
-#endif
     t = qpc_to_100ns(SDL_GetPerformanceCounter());
```
briaguya0, line 642: it seems strange to have this extra
```cpp
t = qpc_to_100ns(SDL_GetPerformanceCounter());
```
in here.

it seems it was always an odd "extra" call for windows, and it makes sense that it existed for other platforms because all of the other ones were inside the windows ifdef

briaguya0 (commented):
i think it might make sense to use mac ifdefs in here to isolate the changes, and then after that lands opening another PR removing the ifdefs and giving people on linux a chance to test it out

ReddestDream:
Okay. Let me clean up and redo. Thanks!

PR #807, opened 2025-01-29
Update gfx_sdl2.cpp to fix macOS Metal Slowdown.

ReddestDream:
Fixes stutter/inability to lock 60 consistently on macOS using Metal. May also provide performance benefits under OpenGL and Linux. This is the weird 57-58 fps issue noticed on Metal graphics on Macs for a long while.

Thanks to @Spodi, @Malkierian, ProxySaw, and Fenrir for their help on Discord getting this figured out! :D

**Note:** This version has been fixed to exclude other operating systems aside from macOS from modification.


larsy1995:
I reviewed and made changes to that code myself, it should be solid. It is for all intents and purposes the same as the original code, but with added support for macOS and only macOS.

larsy1995:
Here is a working build:
https://github.com/larsy1995/Shipwright/actions/runs/13041184423

briaguya0 (approved):

PR #809, opened 2025-02-03
Fixed missing trace on expanded commands

KiritoDv:


briaguya0 (commented):
i'm not the biggest fan of having this block in 2 spots in this method, but i can't think of a great way to avoid that

i guess one way would be to make a little helper method that wraps `displayList->Instructions.push_back(command);` and includes the `GBI_TRACE` stuff, but that kind of feels like overkill

KiritoDv:
Yeah im not too happy but for 3 lines i dont think its worth it

Issue #317, opened 2023-06-20
Window size is not correctly persisted for Mac (probably retina screens)


garrettjoecox:
As of #288 , window size is now persisted through application restarts. On Mac, potentially only on retina screens, this value is not persisted correctly.

When the application is closed, the value stored is double what it should be.

Repro:
Open `<game>.json`
Set `Window.Height` and `Window.Width` to 500
Open application and observe it booted with the correct width/height.
Observe value in json file has not been touched.
Close application
Observe value in json has been doubled
Re-open application and observe the width/height is now double what it should be

Spodi:
There is probably some DPI scaling issue here. Aren't Retina screens usually set to 2x scaling?
Sadly i don't know SDL enough to try and fix this. Also I don't have access to a Mac with such a display to test anything.

Spodi:
Found some discussion about that: https://discourse.libsdl.org/t/high-dpi-mode/34411
Maybe that helps someone that can actually test this stuff...

ReddestDream:
Since Scaling Fix (https://github.com/Kenix3/libultraship/pull/811) is now merged, this issue should now be fixed. Thanks!

PR #811, opened 2025-02-05
macOS Scaling Fix

larsy1995:
This PR fixes more or less all issues related to macOS scaling issues. 
This fixes N64 mode, it fixes ProxySaws issue where the window size wouldn't persist between restarts, but would double in size. This also happened if you went in and out of the built in fullscreen (f11).
This also allows us to re-enable advanced resolution options in ship and 2ship, with the ability to set custom resolutions and aspect ratios, use pixel perfect mode and more. 
Thanks a lot to @ReddestDream for being the idea guy and helping come up with ideas on where to look and what to do as well as the logic needed and for finding the fix to get macOS to always use the framebuffer fixing our last standing issue.

### DIFF
--- a/src/graphic/Fast3D/gfx_pc.cpp
+++ b/src/graphic/Fast3D/gfx_pc.cpp
@@ -4125,14 +4125,14 @@ bool gfx_is_frame_ready() {
     return gfx_wapi->is_frame_ready();
 }
 
-bool viewport_matches_render_resolution(){
+bool viewport_matches_render_resolution() {
 #ifdef __APPLE__
     // Always treat the viewport as not matching the render resolution on mac
     // to avoid issues with retina scaling.
     return false;
 #else
-    if(gfx_current_dimensions.width == gfx_current_game_window_viewport.width &&
-       gfx_current_dimensions.height == gfx_current_game_window_viewport.height) {
+    if (gfx_current_dimensions.width == gfx_current_game_window_viewport.width &&
+        gfx_current_dimensions.height == gfx_current_game_window_viewport.height) {
         return true;
     }
     return false;
--- a/src/graphic/Fast3D/gfx_pc.cpp
+++ b/src/graphic/Fast3D/gfx_pc.cpp
@@ -4125,6 +4125,13 @@ bool gfx_is_frame_ready() {
     return gfx_wapi->is_frame_ready();
 }
 
+#ifdef __APPLE__
+// macOS should use framebuffer like different_size in all cases. This flag Fixes issue with 1:1 on Retina.
+constexpr bool is_macOS = true;
+#else
+constexpr bool is_macOS = false;
+#endif
+
 void gfx_start_frame() {
     gfx_wapi->get_dimensions(&gfx_current_window_dimensions.width, &gfx_current_window_dimensions.height,
                              &gfx_current_window_position_x, &gfx_current_window_position_y);
@@ -4158,9 +4165,9 @@ void gfx_start_frame() {
 
     bool different_size = gfx_current_dimensions.width != gfx_current_game_window_viewport.width ||
                           gfx_current_dimensions.height != gfx_current_game_window_viewport.height;
-    if (different_size || gfx_msaa_level > 1) {
+    if (is_macOS || different_size || gfx_msaa_level > 1) {
         game_renders_to_framebuffer = true;
-        if (different_size) {
+        if (is_macOS || different_size) {
             gfx_rapi->update_framebuffer_parameters(game_framebuffer, gfx_current_dimensions.width,
                                                     gfx_current_dimensions.height, gfx_msaa_level, true, true, true,
                                                     true);
@@ -4171,7 +4178,7 @@ void gfx_start_frame() {
                                                     gfx_current_window_dimensions.height, gfx_msaa_level, false, true,
                                                     true, true);
         }
-        if (gfx_msaa_level > 1 && different_size) {
+        if (gfx_msaa_level > 1 && (is_macOS || different_size)) {
             gfx_rapi->update_framebuffer_parameters(game_framebuffer_msaa_resolved, gfx_current_dimensions.width,
                                                     gfx_current_dimensions.height, 1, false, false, false, false);
         }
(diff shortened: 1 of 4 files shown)


ReddestDream:
**Note:** The following issue is also fixed by this PR: https://github.com/Kenix3/libultraship/issues/317

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -4162,9 +4165,9 @@ void gfx_start_frame() {
 
     bool different_size = gfx_current_dimensions.width != gfx_current_game_window_viewport.width ||
                           gfx_current_dimensions.height != gfx_current_game_window_viewport.height;
-    if (different_size || gfx_msaa_level > 1) {
+    if (is_macOS || different_size || gfx_msaa_level > 1) {
```
briaguya0: i'm not the biggest fan of adding these `is_macOS` checks in here. this block before the changes is already not the easiest to follow, and adding those in makes it harder to follow

maybe just replacing `different_size` with `some_name_that_describes_what_happens_when_different_size_is_set` and having
```cpp
#ifdef __APPLE__
    bool some_name_that_describes_what_happens_when_different_size_is_set = true;
#else
    bool some_name_that_describes_what_happens_when_different_size_is_set = 
                          gfx_current_dimensions.width != gfx_current_game_window_viewport.width ||
                          gfx_current_dimensions.height != gfx_current_game_window_viewport.height;
#endif
```
would do the trick

larsy1995: The is_macOS flag is just to ensure that the frame buffer is always used, otherwise the rendering breaks when the viewport and the internal resolution matches because of how the macOS handles the retina scaling.

larsy1995: The addition of the flag instead of making any other changes was to keep it as non invasive as possible.

larsy1995: That said, I thought I had used the flag several more places, but I work on like 3 different builds at the same time, so I might have forgotten to push those.

larsy1995: Okay, no it is in all three intended places, I just struggle to see the changes on git.

larsy1995: But I'll try that change you suggested, looking at it is should probably work.

briaguya0: > The is_macOS flag is just to ensure that the frame buffer is always used

this is missing some nuance. it's ensuring the framebuffer is used in the way that `different_size` ensures the framebuffer is used, as opposed to the way just having `gfx_msaa_level > 1` ensures it is used.

i'm not arguing the logic here, what you have seems functional and sound. i'm just trying to make it so it's easier to read this block of code and understand why we'd make it in to any given if/else.

also, assuming these 3 uses are the only places `is_macOS` is needed in this file, what we really need is a name for a variable that represents `is_macOS || different_size` - which is what i was trying to get at with my `some_name_that_describes_what_happens_when_different_size_is_set` comment

larsy1995: would "viewportDoesNotMatchResolution" be a proper name for the variable?

larsy1995: I do not disagree with you here, I'm all for the purpose of functions and variables being explicitly clear in their names. As I said, we just didn't want to make any invasive changes. 
I do feel that viewportDoesNotMatchResolution is a much better name.

briaguya0: that seems reasonable, maybe something like

```cpp
#ifdef __APPLE__
    // always treat the viewport as not matching the render resolution on mac
    // to avoid issues with retina scaling
    bool viewport_matches_render_resolution = false;
#else
    bool viewport_matches_render_resolution = 
                          gfx_current_dimensions.width == gfx_current_game_window_viewport.width &&
                          gfx_current_dimensions.height == gfx_current_game_window_viewport.height;
#endif
```

and then replace the `different_size` checks with `!viewport_matches_render_resolution`

larsy1995: I do like that actually. So going by this, we prefer snake_case in this project then, yeah?

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -4241,7 +4244,7 @@ void gfx_run(Gfx* commands, const std::unordered_map<Mtx*, MtxF>& mtx_replacemen
             bool different_size = gfx_current_dimensions.width != gfx_current_game_window_viewport.width ||
                                   gfx_current_dimensions.height != gfx_current_game_window_viewport.height;
 
-            if (different_size) {
+            if (is_macOS || different_size) {
```
briaguya0: might be an opportunity to move the duplicated `different_size` logic out into a function or macro here

src/graphic/Fast3D/gfx_pc.cpp
briaguya0: `snake_case` in the `Fast3D` code specifically - everywhere else in LUS `camelCase` is the standard

src/graphic/Fast3D/gfx_pc.cpp
larsy1995: Already on it, was thinking the same thing while going over this.

larsy1995: bool viewport_matches_render_resolution(){
#ifdef __APPLE__
    // Always treat the viewport as not matching the render resolution on Mac
    // to avoid issues with retina scaling.
    return false;
#else
    if(gfx_current_dimensions.width == gfx_current_game_window_viewport.width &&
       gfx_current_dimensions.height == gfx_current_game_window_viewport.height) {
        return true;
    }
    return false;
#endif
}
and replaced every instance of different_size with !viewport_matches_render_resolution(); 
I'll make the commit when I am done with lectures.

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -4175,7 +4182,7 @@ void gfx_start_frame() {
                                                     gfx_current_window_dimensions.height, gfx_msaa_level, false, true,
                                                     true, true);
         }
-        if (gfx_msaa_level > 1 && different_size) {
+        if (gfx_msaa_level > 1 && (!viewport_matches_render_resolution())) {
```
briaguya0: very much just a nit but
```suggestion
        if (gfx_msaa_level > 1 && !viewport_matches_render_resolution()) {
```

briaguya0 (approved):
tiny nit but LGTM!

larsy1995:
Yeah, mostly if it compiles I won't see those minor things.

PR #812, opened 2025-02-07
Default MaximumFrameLatency increased

Spodi:
Defaults jitter fix (MaximumFrameLatency) to be on (2), instead of off (1), if nothing is set by a port. Keeping it at 1 has a lower latency, but I think it's more sensible to get a smoother experience by default.

### DIFF
--- a/src/graphic/Fast3D/gfx_dxgi.cpp
+++ b/src/graphic/Fast3D/gfx_dxgi.cpp
@@ -462,7 +462,7 @@ void gfx_dxgi_init(const char* game_name, const char* gfx_api_name, bool start_i
     dxgi.qpc_freq = qpc_freq.QuadPart;
 
     dxgi.target_fps = 60;
-    dxgi.maximum_frame_latency = 1;
+    dxgi.maximum_frame_latency = 2;
 
     // Use high-resolution timer by default on Windows 10 (so that NtSetTimerResolution (...) hacks are not needed)
     dxgi.timer = CreateWaitableTimerExW(nullptr, nullptr, CREATE_WAITABLE_TIMER_HIGH_RESOLUTION, TIMER_ALL_ACCESS);


briaguya0:
Digging into this a bit https://github.com/search?q=repo%3AKenix3%2Flibultraship%20maximum_frame_latency&type=code

#### "apply"
https://github.com/Kenix3/libultraship/blob/b7040484f185de8409dfd559f37a43da50d60a28/src/graphic/Fast3D/gfx_dxgi.cpp#L111-L129

#### the line being changed in this PR
https://github.com/Kenix3/libultraship/blob/b7040484f185de8409dfd559f37a43da50d60a28/src/graphic/Fast3D/gfx_dxgi.cpp#L465

#### a bunch of logic in swap buffers end
https://github.com/Kenix3/libultraship/blob/b7040484f185de8409dfd559f37a43da50d60a28/src/graphic/Fast3D/gfx_dxgi.cpp#L866-L909

#### what is called when ports set this using `Fast3dWindow::SetMaximumFrameLatency`
https://github.com/Kenix3/libultraship/blob/b7040484f185de8409dfd559f37a43da50d60a28/src/graphic/Fast3D/gfx_dxgi.cpp#L925-L927

#### a call to apply when creating swapchain
https://github.com/Kenix3/libultraship/blob/b7040484f185de8409dfd559f37a43da50d60a28/src/graphic/Fast3D/gfx_dxgi.cpp#L966-L996

there seems to be a *lot* here - if nothing else 

https://github.com/Kenix3/libultraship/blob/b7040484f185de8409dfd559f37a43da50d60a28/src/graphic/Fast3D/gfx_dxgi.cpp#L875

probably needs to be updated to `2` as well, but any explanation of *why* so much logic is needed for what seems like something we should just set and forget would be really helpful

briaguya0:
after rereading this

> https://github.com/Kenix3/libultraship/blob/b7040484f185de8409dfd559f37a43da50d60a28/src/graphic/Fast3D/gfx_dxgi.cpp#L875
> 
> probably needs to be updated to `2` as well, but any explanation of _why_ so much logic is needed for what seems like something we should just set and forget would be really helpful

it seems that spot can stay `1`

breaking down

https://github.com/Kenix3/libultraship/blob/b7040484f185de8409dfd559f37a43da50d60a28/src/graphic/Fast3D/gfx_dxgi.cpp#L871-L887

current latency value is stored in local var
https://github.com/Kenix3/libultraship/blob/b7040484f185de8409dfd559f37a43da50d60a28/src/graphic/Fast3D/gfx_dxgi.cpp#L871-L874

it's set to 1 to "run out" the existing latency frames or something?
https://github.com/Kenix3/libultraship/blob/b7040484f185de8409dfd559f37a43da50d60a28/src/graphic/Fast3D/gfx_dxgi.cpp#L875-L882

it is set back to the new applied latency value (which was stored in the local var)
https://github.com/Kenix3/libultraship/blob/b7040484f185de8409dfd559f37a43da50d60a28/src/graphic/Fast3D/gfx_dxgi.cpp#L883-L886

briaguya0 (approved):
i still feel the ability to change this at runtime is adding a *ton* of complexity we likely don't need, but from a "minimal change maximum impact" perspective this seems like a good one

Spodi:
> https://github.com/Kenix3/libultraship/blob/b7040484f185de8409dfd559f37a43da50d60a28/src/graphic/Fast3D/gfx_dxgi.cpp#L871-L874
> 
> it's set to 1 to "run out" the existing latency frames or something?
> 
> https://github.com/Kenix3/libultraship/blob/b7040484f185de8409dfd559f37a43da50d60a28/src/graphic/Fast3D/gfx_dxgi.cpp#L875-L882
> 
> it is set back to the new applied latency value (which was stored in the local var)
> 
> https://github.com/Kenix3/libultraship/blob/b7040484f185de8409dfd559f37a43da50d60a28/src/graphic/Fast3D/gfx_dxgi.cpp#L883-L886

Yes, setting it to 1 and than waiting in a loop is to get the latency down again. This is the "undocumented stuff" I talked about. It's never mentioned in the official docs, that you can't just set it at runtime and latency will stay up instead.
In fact the previous aproach here was to create a completely new swapchain. Even the one that wrote the old approach was seemingly supprised, that latency won't go down automatically (according to the comments).
Creating a new swapchain had the problem of confusing ReShade, SK and whatever else is out there. And sometimes also did strange things on its own, when the jitter fix option was changed too fast and therefore multiple swapchains where created in a quick succesion.

PR #813, opened 2025-02-08
Dxgi: Remove unnecessary flushes and cleanup

Spodi:
- Clean up after https://github.com/Kenix3/libultraship/pull/325
Most of that probably got optimized out anyway.

- removes unnecessary flush commands
  - the one I added myself in #325
    there was already one before in `mRapi->EndFrame();` and nothing renders in between.
  - and https://github.com/Kenix3/libultraship/blob/02bb77ef253e2de0969fd2cb36ad2e870677d18d/src/graphic/Fast3D/gfx_direct3d11.cpp#L833-L835
`swap_chain->Present(mVsyncEnabled, DXGI_PRESENT_ALLOW_TEARING)` in `mWapi->SwapBuffersBegin();` is already an implicit `Flush()`; so no need to do it twice in a row, with nothing else in between happening.

Removing the flushes maybe even helps the lower end windows devices. As those flushes can in theory be pretty demanding.
https://learn.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11devicecontext-flush
> If an application calls this method when not necessary, it incurs a performance penalty. Each call to Flush incurs a significant amount of overhead. 

 

I should have done that ages ago...
Should not change any behavior. But as always: please test before merging.

### DIFF
--- a/src/graphic/Fast3D/backends/gfx_direct3d11.cpp
+++ b/src/graphic/Fast3D/backends/gfx_direct3d11.cpp
@@ -724,7 +724,6 @@ void GfxRenderingAPIDX11::EndFrame() {
 }
 
 void GfxRenderingAPIDX11::FinishRender() {
-    mContext->Flush();
 }
 
 int GfxRenderingAPIDX11::CreateFramebuffer() {
--- a/src/graphic/Fast3D/backends/gfx_dxgi.cpp
+++ b/src/graphic/Fast3D/backends/gfx_dxgi.cpp
@@ -876,19 +876,6 @@ void GfxWindowBackendDXGI::SwapBuffersBegin() {
     mVsyncEnabled = Ship::Context::GetInstance()->GetConsoleVariables()->GetInteger(CVAR_VSYNC_ENABLED, 1) ? 1 : 0;
 
     LARGE_INTEGER t;
-    ComPtr<ID3D11Device> mDevice;
-    mSwapChainDevice.As(&mDevice);
-
-    if (mDevice != nullptr) {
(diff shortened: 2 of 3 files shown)


briaguya0:
> please test before merging

please make a testing PR on a port (and cross link them) so people can verify this

Spodi:
Testbuild: https://github.com/HarbourMasters/Shipwright/pull/5024

Spodi:
Converting to a draft for now, as I need to test "flush" situation more. The results of my quick tests where inconsistent.

Spodi:
Seems to be good without the flushing.

I need to hold back not to also remove this flush here https://github.com/Kenix3/libultraship/blob/02bb77ef253e2de0969fd2cb36ad2e870677d18d/src/graphic/Fast3D/gfx_direct3d11.cpp#L833-L835 but i dunno if it serves a purpose or not. Stuff looks fine without, but could be something specific.

EDIT: I removed it and it seems fine 🤷

Spodi:
Tesbuild is out of date. will update this comment if it isn't anymore.

EDIT: Up to date now (https://github.com/HarbourMasters/Shipwright/pull/5024)

louist103:
I can confirm that 2ship can successfully builds with this and this build works fine in Wine

Spodi:
This maybe even helps the lower end windows devices. As those flushes can in theory be pretty demanding.

https://learn.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11devicecontext-flush
> If an application calls this method when not necessary, it incurs a performance penalty. Each call to Flush incurs a significant amount of overhead.

serprex (approved):

PR #814, opened 2025-02-09
fix rumble mappings not loading from config

briaguya0:
some leftover shipdeviceindex stuff was making it so the factory saw rumble mappings in the config as invalid. this PR removes that as it's no longer used


briaguya0:
image

PR #815, opened 2025-02-10
Add method to call for shutting down the ResourceManager thread pool

Malkierian:
This adds a method in `ResourceManager` to allow for setting up a thread pool shutdown to prevent crashes when exiting while still loading assets. I have a wait time of 2 seconds here between pausing and purging to prevent deadlocks, more than that is unnecessary (especially if it's only for resource threads, as I think it should be). Unfortunately, because of the way shutdowns work, this has to be called manually from the port before nulling `Context`. This is demonstrated in the asset preload PR: https://github.com/HarbourMasters/Shipwright/pull/4652

The reasoning for this PR is that I needed to be able to shut down the resource thread pool because of the way preloading would queue up a lot of loads at the start, and would crash if you tried to exit while in the opening sequence. This does not do any of the work of exposing the `ResourceManager` thread pool because there's still some ambiguity around the scope of that work (should only be used for non-critical tasks in my opinion; no save or config writing). The reason I needed to do it on the preload PR is because I couldn't get the crash to happen on latest develop. It's easiest to recreate with the 4K Reloaded pack. To let it trigger, comment the line `OTRGlobals::Instance->context->GetResourceManager()->ShutDownThreadPool();` in OTRGlobals (should be line 1248), and close the program as soon as you see the Nintendo logo. Uncomment it to see it not crash under the same circumstances. I *think* it still happens in release mode, but you'll get the stack trace in debug in VS.

### DIFF
--- a/src/resource/ResourceManager.cpp
+++ b/src/resource/ResourceManager.cpp
@@ -410,4 +410,10 @@ void ResourceManager::SetAltAssetsEnabled(bool isEnabled) {
     mAltAssetsEnabled = isEnabled;
 }
 
+void ResourceManager::ShutDownThreadPool() {
+    mThreadPool->pause();
+    mThreadPool->wait_for(std::chrono::duration<double>(2));
+    mThreadPool->purge();
+}
+
 } // namespace Ship
--- a/src/resource/ResourceManager.cpp
+++ b/src/resource/ResourceManager.cpp
@@ -291,14 +291,14 @@ ResourceManager::GetCachedResource(std::variant<ResourceLoadError, std::shared_p
 }
 
 std::shared_ptr<std::vector<std::shared_ptr<IResource>>>
-ResourceManager::LoadResourcesProcess(const ResourceFilter& filter) {
+ResourceManager::LoadResourcesProcess(const ResourceFilter& filter, bool exact) {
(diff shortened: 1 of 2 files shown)


src/resource/ResourceManager.cpp
```diff
@@ -410,4 +410,10 @@ void ResourceManager::SetAltAssetsEnabled(bool isEnabled) {
     mAltAssetsEnabled = isEnabled;
 }
 
+void ResourceManager::ShutDownThreadPool() {
+    mThreadPool->pause();
+    mThreadPool->wait_for(std::chrono::duration<double>(2));
```
briaguya0, line 416: i see you mentioned
> I have a wait time of 2 seconds here between pausing and purging to prevent deadlocks

that feels a little bit magic to me, are we just assuming we'll never have a pool where 2 seconds isn't enough time? have you looked into alternative ways to mitigate deadlock that don't rely on a hardcoded delay?

briaguya0: @Kenix3's suggestion is to just pass the amount of time to wait as a function param

Kenix3:
Merge pending CI succeeding. Let me know on Discord if you need help.

PR #816, opened 2025-02-11
Remove load tile check

MegaMech:
This line is not needed. Other tiles are valid input


briaguya0:
could you update the description of this PR to explain why this should be removed?

briaguya0:
@MegaMech the new description still doesn't help me understand *why* this should be removed

briaguya0:
after doing more research, we found that `G_TX_LOADTILE` is suggestive not prescriptive, and we can load into other places without issue

Issue #661, opened 2024-08-20
Macos - Fullscreen through LUS does not work properly


Archez:
On macos, fullscreening using the LUS method does not match the OS fullscreen function.

Using the LUS method has the following issues:
* Other monitors get blacked out
* Unable to use OS features like Window switching, Desktop Space switching, 
* The fullscreen goes beyond the "safe area" of the screen (window renders behind the screen notch)
* Activating LUS fullscreen while the OS fullscreen is active leads to crashing and potentially stalling of the whole OS

Using the OS fullscreen method:
* LUS does not think it is fullscreen and therefore does not hide the mouse cursor

Presumably this also impacts the new Game Mode feature in newer macOS, which requires being fullscreened by the OS.

larsy1995:
I’m going to add here that the LUS full screen mode actually does activate game mode, but the rest is valid.

larsy1995:
A PR that fixes this is coming soon.

Archez:
Addressed by #817 and #855

PR #817, opened 2025-02-13
Mac fullscreen fix

larsy1995:
Changes windowed mode on macOS to use native macOS fullscreen instead of SDL to enable macOS menubar access while in borderless fullscreen mode.

This workaround is needed until we update to SDL3 and can use SDL_HINT_VIDEO_MAC_FULLSCREEN_MENU_VISIBILITY for this functionality.
Also fixes crashing when going from native macOS fullscreen to SDL.

Also changes the default fullscreen behaviour from exclusive to borderless windowed.
This will need a separate PR in each game to change this visually, as well as to disable the fullscreen toggle while already in a fullscreen mode to avoid unintended behaviour.

Thanks to @ReddestDream for all the help testing and coming up with the workaround with disabling the toggle while in a fullscreen mode.


ReddestDream:
**Note:** This PR works best with and was designed with the macOS Scaling Fix in mind: https://github.com/Kenix3/libultraship/pull/811

ReddestDream:
**Also note:** This PR was designed to address this issue with the input of @Archez: https://github.com/Kenix3/libultraship/issues/661

src/graphic/Fast3D/gfx_sdl2.cpp
```diff
@@ -251,6 +252,30 @@ static void set_fullscreen(bool on, bool call_callback) {
         SDL_SetWindowPosition(wnd, posX, posY);
         SDL_SetWindowSize(wnd, window_width, window_height);
     }
+#if defined(__APPLE__)
+    // Current implementation of the fullscreening with native macOS fullscreen for Windowed Mode and
+    // SDL for Exclusive Fullscreen. This code can and will be changed when we upgrade to SDL3 because of the
+    // ability to use SDL_HINT_VIDEO_MAC_FULLSCREEN_MENU_VISIBILITY to get the menubar working instead of this
+    // workaround
+    bool useNativeMacOSFullscreen = !CVarGetInteger(CVAR_SDL_WINDOWED_FULLSCREEN, 0);
```
briaguya0, line 260: my biggest issue with this PR is that it is changing the meaning of `CVAR_SDL_WINDOWED_FULLSCREEN`

ReddestDream: That is understandable. [larsy1995](https://github.com/larsy1995) (Fenrir) can say more, but my understanding is that this was done to conform with what @Archez suggested on Discord: "It might make sense for the LUS fullscreen to use the OS one when "windowed fullscreen" is off, and the SDL one when on." We followed this suggestion for the implementation here since this is a solution to the issue by @Archez here https://github.com/Kenix3/libultraship/issues/661. If it turns out that this isn't the best way, please let us know. Thanks!

ghost: How about we rename the CVAR? To make an emphasis on the fact that it changes the behavior of the fullscreen mode, not just borderless window mode on and off?

larsy1995: At least that way it would be clearer on macOS and not make it confusing for other devs. A universal changeFullscreenMode CVar.

Archez: In the context and use cases of our ports, is there any reason why anyone would want "exclusive fullscreen" over the native behavior for MacOS? I only originally made that suggestion as a "if both matter", but I'm not convinced at this point that both matter.

It seems all this could just be simplified on MacOS (hide the windowed fullscreen toggle/ignore the CVar).

briaguya0: > In the context and use cases of our ports, is there any reason why anyone would want "exclusive fullscreen" over the native behavior for MacOS? I only originally made that suggestion as a "if both matter", but I'm not convinced at this point that both matter.
> 
> It seems all this could just be simplified on MacOS (hide the windowed fullscreen toggle/ignore the CVar).

yeah, i suggested the same thing (just always use native on mac) in https://github.com/HarbourMasters/Shipwright/pull/5060

ReddestDream: I'm not convinced that both matter either. IIRC, we kept the exclusive fullscreen because it was able to use "Direct" rendering while the native only used "Composited," but I'm not really sure that matters in the end.

ReddestDream: I got confirmation from [larsy1995](https://github.com/larsy1995) (Fenrir) that, yes, we kept the exclusive (SDL) fullscreen because it was the only way to get Direct rendering. Again, if that matters. Also, when SDL3 is available, as noted above, there may even be further streamlining possible.

ReddestDream: Just as a note, other SDL apps (_e.g.,_ DOSBox-X) on macOS offer two fullscreen options kinda like what we have here. Again, it's a question of how much it matters and what makes sense for the project, clarity, consistency, etc.

larsy1995: Anyhow, this fix also makes the f11 button exit native fullscreen as well. I agree the code would be cleaner with just native borderless and is what I wanted from the start personally, but I just followed the specs given by Archez in that issue ReddestDream linked earlier. 
I just need the go ahead and I'll make the changes to both this and the menubars asap.

larsy1995:
I was not ready with this tbh, I have a more updated version of this fix on my computer right now, with native fullscreen only, but I had not committed it yet.

PR #819, opened 2025-02-14
Fix and improve the console window

Archez:
This fixes and addresses some issues with the Console Window:

* After the recent ImGui bump, the `bind` command would throw an assert due to incorrect ImGuiKey values being passed into the name function. This has been updated to use the name enum key range.
* After the recent ImGui bump, it was exposed that the output log had conflicting ImGui ID's. This was due to a bad string copy that was bringing over a null terminator which prevented the existing `##` ID from applying. This removes that null terminator copy so the string works correctly.
* Added `unbind` command that can unbind both normal and toggle commands.
* Fixed help text for `get` command.
* Clearing the console now properly clears other values like the previously selected/highlighted output text.
* Pressing the down arrow when highlighting output text now works properly (selects one output line down).
* The output text now only checks for up/down presses if the output text table is focused.
* Pressed up/down in the console input field now behaves more like a standard terminal experience.
  * Pressing up will autofill the most recently submitted command. Pressing up again will go to the command before that, so-on until you hit the very first command.
  * Pressing down, when having pressed up previously, will cycle back towards the more recent commands, until you get to an empty prompt.
* Dispatching a command will reset the "previous history index"
* The autocomplete hint tooltip now displays above the input text, instead of below. Previously, due to issues with ImGui directly, tooltips that display outside their parent window causes undesirable effects, particularly in that it unfocuses the input field, making entering commands on a floating console window **infuriating**. Having the tooltip presented above the input, keeps it within its window and stops this behavior.

### DIFF
--- a/src/window/gui/ConsoleWindow.cpp
+++ b/src/window/gui/ConsoleWindow.cpp
@@ -572,7 +572,6 @@ void ConsoleWindow::Dispatch(const std::string& line) {
 int ConsoleWindow::CallbackStub(ImGuiInputTextCallbackData* data) {
     const auto instance = static_cast<ConsoleWindow*>(data->UserData);
     const bool emptyHistory = instance->mHistory.empty();
-    const int historyIndex = instance->mHistoryIndex;
     auto console = Context::GetInstance()->GetConsole();
     std::string history;
 
@@ -591,9 +590,9 @@ int ConsoleWindow::CallbackStub(ImGuiInputTextCallbackData* data) {
             if (emptyHistory) {
                 break;
             }
-            if (historyIndex > 0) {
+            if (instance->mHistoryIndex > 0) {
                 instance->mHistoryIndex -= 1;
-            } else if (historyIndex < 0) {
+            } else if (instance->mHistoryIndex < 0) {
                 instance->mHistoryIndex = static_cast<int>(instance->mHistory.size()) - 1;
             }
             data->DeleteChars(0, data->BufTextLen);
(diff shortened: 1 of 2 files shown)


src/window/gui/ConsoleWindow.cpp
```diff
@@ -533,23 +591,29 @@ int ConsoleWindow::CallbackStub(ImGuiInputTextCallbackData* data) {
             if (emptyHistory) {
                 break;
             }
-            if (historyIndex < static_cast<int>(instance->mHistory.size()) - 1) {
-                instance->mHistoryIndex += 1;
+            if (historyIndex > 0) {
```
briaguya0: reading through this i'm finding it's not the easiest thing to keep track of `historyIndex` vs `instance->mHistoryIndex`

i'm thinking maybe renaming `historyIndex` could help but i can't think of a better name off the top of my head

briaguya0 (approved):
one non-blocking comment about a variable name but overall LGTM!

src/window/gui/ConsoleWindow.cpp
Archez: Honestly having the two variables was a source of some of the bad behavior (updating the instance var but still using the local var later).

I contemplated just removing the local var entirely, I'll just do that now.

briaguya0: the change looks great!

one last question before i reapprove: are there any situations where we were *relying* on the local var not changing?

Archez: Yeah nothing relies on the local var. Before my changes some of the code used it, but it was undesirable and counter intuitive to the user (read: pressing an arrow key to cycle the autofill, but it not taking affect until pressing the key twice because local var was desynced with instance var)

briaguya0 (approved):
Even happier with the new changes!

PR #820, opened 2025-02-14
Fix O2R log statement not compiling

Archez:
I don't know why the github builds did not complain about this, but the use of `GetLastError()` should not be in the O2RArchive class, as that method is strictly from StormLib and pertains to the OTRArchive class.

Removing the function reference and updating the log statement allowed LUS to build correctly for me.


KiritoDv:
Oh i thought i was the only one

briaguya0 (approved):

PR #822, opened 2025-02-16
V-Sync toggle in SDL (OpenGL + Metal)

Spodi:
Works without changes on the port side, but similar changes like this are recommended: https://github.com/HarbourMasters/Shipwright/pull/5074 (This can also be used as Test-PR)

### DIFF
--- a/src/graphic/Fast3D/gfx_sdl2.cpp
+++ b/src/graphic/Fast3D/gfx_sdl2.cpp
@@ -662,13 +662,13 @@ static inline void sync_framerate_with_timer() {
 }
 
 static void gfx_sdl_swap_buffers_begin() {
-    if (vsync_enabled !=
-        Ship::Context::GetInstance()->GetConsoleVariables()->GetInteger(CVAR_VSYNC_ENABLED, 1)) {
+    if (vsync_enabled != Ship::Context::GetInstance()->GetConsoleVariables()->GetInteger(CVAR_VSYNC_ENABLED, 1)) {
         // Make sure only 0 or 1 is set.
-        vsync_enabled =
-            !!Ship::Context::GetInstance()->GetConsoleVariables()->GetInteger(CVAR_VSYNC_ENABLED, 1);
+        vsync_enabled = !!Ship::Context::GetInstance()->GetConsoleVariables()->GetInteger(CVAR_VSYNC_ENABLED, 1);
     }
     SDL_GL_SetSwapInterval(vsync_enabled);
+    SDL_RenderSetVSync(renderer, vsync_enabled);
+
     sync_framerate_with_timer();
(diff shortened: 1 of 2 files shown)


src/graphic/Fast3D/gfx_sdl2.cpp
```diff
@@ -662,6 +662,13 @@ static inline void sync_framerate_with_timer() {
 }
 
 static void gfx_sdl_swap_buffers_begin() {
+    if (vsync_enabled != Ship::Context::GetInstance()->GetConsoleVariables()->GetInteger(CVAR_VSYNC_ENABLED, 1)) {
+        // Make sure only 0 or 1 is set.
+        vsync_enabled = !!Ship::Context::GetInstance()->GetConsoleVariables()->GetInteger(CVAR_VSYNC_ENABLED, 1);
+    }
+    SDL_GL_SetSwapInterval(vsync_enabled);
+    SDL_RenderSetVSync(renderer, vsync_enabled);
```
briaguya0: i feel like we shouldn't be calling these every time we call `SDL_GL_SwapWindow`. this really should only happen when the cvar changes

Spodi: where and how would i do that?

Spodi: oh, or do you mean setting the swap interval? I thought you ment getting the cvar.

Spodi: changed that

src/graphic/Fast3D/gfx_sdl2.cpp
```diff
@@ -662,6 +662,14 @@ static inline void sync_framerate_with_timer() {
 }
 
 static void gfx_sdl_swap_buffers_begin() {
+    if (vsync_enabled != Ship::Context::GetInstance()->GetConsoleVariables()->GetInteger(CVAR_VSYNC_ENABLED, 1)) {
+        // Make sure only 0 or 1 is set.
+        vsync_enabled =
+            (Ship::Context::GetInstance()->GetConsoleVariables()->GetInteger(CVAR_VSYNC_ENABLED, 1) ? 1 : 0);
+        SDL_GL_SetSwapInterval(vsync_enabled);
+        SDL_RenderSetVSync(renderer, vsync_enabled);
+    }
```
briaguya0: this does seem *better* than calling `SDL_GL_SetSwapInterval` and `SDL_RenderSetVSync` every time, but it'd still prefer to find a way to not need to check the cvar all the time like this

Spodi: Yeah, but I have no idea how. Any suggestion?

PR #824, opened 2025-02-22
Change GL_RGB --> GL_RGBA8

MegaMech:
This seems to be a spelling error since the other renderers use rgba8 or some variation of (bgra8).

This change will be necessary if the colour id framebuffer were to be implemented as it's based on a uint32_t instead of a 24 bit value.


Archez:
These changes cause a significant difference in how OpenGL appears to render transparent textures: Example 
image
The clouds on the map are darkened, the HUD A and C buttons are darkened. Water boxes, road paths, etc are all darkened.

This is how it looked before the changes: 
image

MegaMech:
I noticed there's another area that needs an rgba change to truly see if it works or not. In I think gfx_pc

Archez:
> I noticed there's another area that needs an rgba change to truly see if it works or not. In I think gfx_pc

In my test above, I had changed one more rgba spot in the OpenGL file that missed, in update_fb_parameters

PR #825, opened 2025-02-23
Add Colour Id Framebuffer

MegaMech:
This is needed for object picking in the editor.

API Changes:
* `update_framebuffer_parameters` now has an image_data argument. Use NULL if it's not needed.

Todo:
* Modify shader


MegaMech:
Closed because gave up and used vert collision mesh instead

PR #821, opened 2025-02-15
Moved shaders to prism lang

KiritoDv:
This PR ports the shader files to use [prism](https://github.com/KiritoDv/prism-processor) which is a preprocessor so we can store them on otrs that can be modified by ports

PR #827, opened 2025-02-23
Colouridandshaders

MegaMech:
no merge

This is just a dup of
https://github.com/Kenix3/libultraship/pull/825

with included
https://github.com/Kenix3/libultraship/pull/821

For easier work


MegaMech:
Closed because gave up and used vert collision mesh instead

PR #831, opened 2025-03-10
Add surround 5.1 support

dcampora:
This MR adds support for 5.1 audio in both the SDL and WASAPI backend. If no option has been selected, stereo will be picked by default.


briaguya0 (commented):
overall this is looking great! my only comments are about using `AudioSurroundSetting` more or less as a bool to mean stereo or 5.1

i don't immediately see a use case for something other than 5.1 and stereo, but i still think it'd be nice to not treat the cvar as a bool here.

this would mean:
* `AudioSurroundSetting` -> `AudioChannelsSetting` (or something along those lines)
* `AudioPlayer::GetAudioSurround` -> `AudioPlayer::GetAudioChannels` (i don't *love* this name but as long as it's still returning an `AudioChannelsSetting` instead of an int it should be fine
* `AudioPlayer::SetAudioSurround` -> `AudioPlayer::SetAudioChannels` (i don't *love* this name but as long as it's still taking in an `AudioChannelsSetting` instead of an int it should be fine
* update `Config::GetCurrentAudioSurround()` and `GetNumAudioChannels()` to be use switch/case instead of ternaries

PR #832, opened 2025-03-12
Add surround 5.1 support

tortugaveloz:
This MR adds support for 5.1 audio in both the SDL and WASAPI backend. If no option has been selected, stereo will be picked by default.


tortugaveloz:
@briaguya-ai I applied your comments [from here](https://github.com/Kenix3/libultraship/pull/831#pullrequestreview-2672711489).

ThreeDeeJay:
Interesting. So does it create the surround mix by grabbing the sounds' 3D coordinates relative to the camera/listener point?
Or is it like an overhaul/modified version of the ancient Dolby Surround mode with/without encoding/decoding?
Either way it's a great addition, though in the future you might wanna consider [OpenAL](https://github.com/kcat/openal-soft) (which [even SDL recommends](https://github.com/libsdl-org/SDL_mixer/blob/e777b524e8f854c6c6590e2b5766bb1cee52e016/include/SDL3_mixer/SDL_mixer.h#L1285-L1287)) for more/better mixing modes ([7.1, Dolby Atmos](https://github.com/kcat/openal-soft/blob/6d220834b0f6b409c863900aa69309c69d0e68d4/alsoftrc.sample#L58-L59), [3D audio over headphones](https://github.com/kcat/openal-soft/blob/6d220834b0f6b409c863900aa69309c69d0e68d4/alsoftrc.sample#L110-L111) like [this](https://youtu.be/zO1XuqJUPu0), etc.) as well as other advanced effects like ([EAX](https://github.com/kcat/openal-soft/blob/6d220834b0f6b409c863900aa69309c69d0e68d4/al/eax/api.cpp#L302-L327)/[EFX](https://github.com/kcat/openal-soft/blob/6d220834b0f6b409c863900aa69309c69d0e68d4/include/AL/efx-presets.h#L38-L401)) reverb 👀👌

Kenix3 (approved):

cmake/cvars.cmake
```diff
@@ -20,6 +20,7 @@ set(CVAR_GAME_OVERLAY_FONT "gOverlayFont" CACHE STRING "")
 set(CVAR_MENU_BAR_OPEN "gOpenMenuBar" CACHE STRING "")
 set(CVAR_PREFIX_CONTROLLERS "gControllers" CACHE STRING "")
 set(CVAR_PREFIX_ADVANCED_RESOLUTION "gAdvancedResolution" CACHE STRING "")
+set(CVAR_SURROUND_AUDIO "gSurroundAudio" CACHE STRING "")
```
briaguya0: the cvar name here still implies a boolean surround/not to me. could you update this to match the other name changes?

tortugaveloz: Renamed to gAudioChannelsSetting.

tortugaveloz:
@ThreeDeeJay that's up to the game itself. In Starship, I added the following (https://github.com/HarbourMasters/Starship/pull/174):

- Voices from the center speaker.
- SFX calculated spatially from where the sound comes from, attributed to the speakers accordingly.
- BGM played on both front and rear stereo.
- Subwoofer plays < 80 Hz.
- Reverb on rears.

And thanks for the suggestions. For most of those however I would need to have a compatible device to make sure I'm doing the right thing.

src/public/bridge/audiobridge.h
```diff
@@ -6,12 +6,16 @@
 #include "stdint.h"
 #include "stddef.h"
 
+typedef enum AudioChannelsSetting { stereo, surround51 } AudioChannelsSetting;
```
briaguya0:
```suggestion
typedef enum AudioChannelsSetting { stereo, surround51, max } AudioChannelsSetting;
```

src/config/Config.cpp
```diff
@@ -226,6 +226,16 @@ AudioBackend Config::GetCurrentAudioBackend() {
     return AudioBackend::SDL;
 }
 
+AudioChannelsSetting Config::GetCurrentAudioSurround() {
+    int surround = GetInt("CVars.gAudioChannelsSetting");
```
briaguya0: sorry i missed this one before, having this get use a hardcoded string will make it so ports changing the string for `CVAR_AUDIO_CHANNELS_SETTING` won't work.

```suggestion
    int surround = GetInt(CVAR_AUDIO_CHANNELS_SETTING);
```

also, we should probably have a default in here

```suggestion
    int surround = GetInt(CVAR_AUDIO_CHANNELS_SETTING, 0);
```

and we should probably avoid using magic numbers

```suggestion
    int32_t surround = GetInt(CVAR_AUDIO_CHANNELS_SETTING, static_cast<int32_t>(AudioChannelsSetting::stereo));
```

and we should probably not just rely on `stereo` being `0`

```suggestion
    int32_t surround = GetInt(CVAR_AUDIO_CHANNELS_SETTING, static_cast<int32_t>(AudioChannelsSetting::max));
```

src/config/Config.cpp
```diff
@@ -226,6 +226,16 @@ AudioBackend Config::GetCurrentAudioBackend() {
...
+    switch (surround) {
+        case 1:
+            return AudioChannelsSetting::surround51;
+        default:
+            return AudioChannelsSetting::stereo;
+    }
```
briaguya0: we should avoid magic numbers here

```suggestion
    switch (surround) {
        case AudioChannelsSetting::surround51:
            return AudioChannelsSetting::surround51;
        case AudioChannelsSetting::stereo:
        case AudioChannelsSetting::max:
        default:
            return AudioChannelsSetting::stereo;
    }
```

it's also possible we could just simplify this entire method to just `static_cast` directly to the enum from the int

```cpp
AudioChannelsSetting Config::GetCurrentAudioSurround() {
    return static_cast<AudioChannelsSetting>(GetInt(CVAR_AUDIO_CHANNELS_SETTING, static_cast<int32_t>(AudioChannelsSetting::max));
}
```

it's also possible implicit casting will work too but i'm not sure what rules all the ports set up for all the different compilers and if any might complain about it

```cpp
AudioChannelsSetting Config::GetCurrentAudioSurround() {
    return GetInt(CVAR_AUDIO_CHANNELS_SETTING, static_cast<int32_t>(AudioChannelsSetting::max));
}
```

src/config/Config.cpp
```diff
@@ -226,6 +226,16 @@ AudioBackend Config::GetCurrentAudioBackend() {
     return AudioBackend::SDL;
 }
 
+AudioChannelsSetting Config::GetCurrentAudioSurround() {
```
briaguya0:
```suggestion
AudioChannelsSetting Config::GetCurrentAudioChannelsSetting() {
```

src/config/Config.h
```diff
@@ -67,6 +67,7 @@ class Config {
     void SetCurrentAudioBackend(AudioBackend backend);
     WindowBackend GetWindowBackend();
     void SetWindowBackend(WindowBackend backend);
+    AudioChannelsSetting GetCurrentAudioSurround();
```
briaguya0:
```suggestion
    AudioChannelsSetting GetCurrentAudioChannelsSetting();
```

briaguya0 (commented):
sorry for all the back and forth on this, i think i found everything i want to comment on this time!

tortugaveloz:
There is seemingly an error in gfx_opengl.cpp that is unrelated to this PR. CC @briaguya-ai

ThreeDeeJay:
> @ThreeDeeJay that's up to the game itself. In Starship, I added the following ([HarbourMasters/Starship#174](https://github.com/HarbourMasters/Starship/pull/174)):
> 
>     * Voices from the center speaker.
> 
>     * SFX calculated spatially from where the sound comes from, attributed to the speakers accordingly.
> 
>     * BGM played on both front and rear stereo.
> 
>     * Subwoofer plays < 80 Hz.
> 
>     * Reverb on rears.
> 
> 
> And thanks for the suggestions. For most of those however I would need to have a compatible device to make sure I'm doing the right thing.

OpenAL Soft is done completely in software, as the name implies. So it works on any sound card, and is also multiplatform by the way 👌. By default I uses whatever speaker configuration the Windows playback device is set to, and HRTF (3D audio for any stereo headphones) could be enabled via config file or implemented as an audio option in the GUI. The other stuff is also optional.

briaguya0:
> There is seemingly an error in gfx_opengl.cpp that is unrelated to this PR. CC @briaguya-ai

latest `main` passed CI, not sure what you're running into

tortugaveloz:
Ah, the typedef enum does not put things into a namespace. Since it is used in C, we can't use enum class. I'll give enum names a prefix. Funny it doesn't fail locally.

tortugaveloz:
It should be fixed now.

PR #834, opened 2025-03-13
Add command gSPSegmentInterp

NEstelami:
This is used in 2Ship (and will be used in SoH) to fix interpolation issues with horses. It adds a new argument type for `gsMoveWd`, `G_MW_SEGMENT_INTERP`, which sets a segment only on a specific interpolated frame. This allows for certain data to be substituted during interpolation. In the case of 2Ship, this data is skinned vertices.


Kenix3 (approved):
Merge pending CI fixes

PR #837, opened 2025-03-16
fix triangle with xml for f3dex

coco875:


MegaMech:
Confirmed this works in port

MegaMech (approved):

PR #838, opened 2025-03-18
Implemented dynamic shader loading

KiritoDv:


include/libultraship/libultra/gbi.h
```diff
@@ -2791,11 +2791,19 @@ typedef union Gfx {
 #define gsSPGrayscale(state) \
     { (_SHIFTL(G_SETGRAYSCALE, 24, 8)), (state) }
 
-#define gsSPLoadShader(shader, type) gsDma1p(G_LOAD_SHADER, shader, 0, type)
-#define gsSPUnloadShader() gsDma1p(G_LOAD_SHADER, 0, 0, 0)
+#define gsSPSetShader(shader, frag, vtx)          \
```
Rozelette, line 2797: `shader` is no longer needed

include/libultraship/libultra/gbi.h
```diff
@@ -2791,11 +2791,19 @@ typedef union Gfx {
...
+    { (_SHIFTL(G_LOAD_SHADER, 24, 8)), (vtx) }, { \
+        0, (frag)                                 \
+    }
 
-#define gSPLoadShader(pkt, shader, type) gDma1p(pkt, G_LOAD_SHADER, shader, 0, type)
-#define gSPUnloadShader(pkt) gDma1p(pkt, G_LOAD_SHADER, 0, 0, 0)
+#define gSPSetShader(pkt, frag, vtx)                   \
```
Rozelette, line 2802: Consider changing this so that you set the fragment and vertex shaders separately, as what it seemed to be like previously . You already need 2 `Gfx`s, so there is no increase in size. Also, there will likely be situations where you want to change the custom fragment shader but keep the vertex one, or maybe you don't even know what the vertex shader is currently.

src/graphic/Fast3D/gfx_cc.cpp
```diff
@@ -82,4 +78,6 @@ void gfx_cc_get_features(uint64_t shader_id0, uint32_t shader_id1, struct CCFeat
     if (cc_features->used_textures[1] && shader_id1 & SHADER_OPT(TEXEL1_BLEND)) {
         cc_features->used_blend[1] = true;
     }
+
+    cc_features->shader_id = (shader_id1 >> 16) & 0xFFFF;
```
Rozelette: The numbers should be constants. Consider
```
constexpr uint32_t ShaderIdSize = 16;
constexpr uint32_t ShaderIdMask = (1 << ShaderIdSize ) - 1;
constexpr uint32_t ShaderIdShift = 16;
```

src/graphic/Fast3D/gfx_cc.h
```diff
@@ -55,7 +55,6 @@ enum class ShaderOpts {
     TEXEL1_MASK,
     TEXEL0_BLEND,
     TEXEL1_BLEND,
-    USE_SHADER,
```
Rozelette, line 58: This should stay (but be renamed) to document that these bits are being used in `cc_options`. The width of the id (16 bits) also needs to be documented as being in use, so that the next bit to use is clear.

src/graphic/Fast3D/gfx_direct3d_common.cpp
```diff
@@ -210,8 +211,18 @@ std::string gfx_direct3d_common_build_shader(size_t& num_floats, const CCFeature
     init->Type = (uint32_t)Ship::ResourceType::Shader;
     init->ByteOrder = Ship::Endianness::Native;
     init->Format = RESOURCE_FORMAT_BINARY;
-    auto res = static_pointer_cast<Ship::Shader>(Ship::Context::GetInstance()->GetResourceManager()->LoadResource(
-        "shaders/directx/default.shader.hlsl", true, init));
+    auto shaderMod = gfx_get_shader(cc_features.shader_id);
```
Rozelette: Consider reducing the use of `auto` throughout this PR, especially in situations where it is easy to type out the type (`int`, etc.).

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -1558,6 +1557,9 @@ static void gfx_sp_tri1(uint8_t vtx1_idx, uint8_t vtx2_idx, uint8_t vtx3_idx, bo
     if (use_grayscale) {
         cc_options |= SHADER_OPT(GRAYSCALE);
     }
+
+    cc_options |= (g_rdp.current_shader << 16);
```
Rozelette: Should be updated to use a constant too.

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -2946,34 +2939,46 @@ bool gfx_movemem_handler_otr(F3DGfx** cmd0) {
...
-    initData->Path = path;
-    initData->IsCustom = false;
-    initData->ByteOrder = Ship::Endianness::Native;
-    auto shader = Ship::Context::GetInstance()->GetResourceManager()->GetArchiveManager()->LoadFile(path);
-    if (shader == nullptr || !shader->IsLoaded) {
-        return -1;
+int16_t gfx_search_existing_shader(const char* vertex, const char* fragment) {
+    ShaderMod shader = { vertex, fragment };
+    for (int i = 0; i < shader_ids.size(); i++) {
+        if (memcmp(&shader_ids[i], &shader, sizeof(ShaderMod)) == 0) {
```
Rozelette: Instead of doing a `memcmp`, implement `operator==` for `ShaderMod`. I don't think this `memcmp` did what you intended anyways, since it compared the pointer values rather than their contents.

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -2946,34 +2939,46 @@ bool gfx_movemem_handler_otr(F3DGfx** cmd0) {
...
+        g_rdp.current_shader = -1;
     }
 
-    const auto path = std::string(file);
-    const auto shaderId = gfx_create_shader(path);
-    g_rdp.current_shader = { true, shaderId, (uint8_t)C0(16, 1) };
     return false;
 }
 
+std::optional<ShaderMod> gfx_get_shader(int16_t id) {
```
Rozelette: I don't like this use of `std::optional`, since it hides error cases from us. If an id is >= the size, or a negative value other than -1, that should error. Either an assert or a warning and we fall back to the default.

Consider an implementation that uses `uint16_t` as the id type, 0 is the default, and shader_id is seeded with the default shader. Each implementation calls `gfx_get_shader` and just uses the return value, not needing to check if it has a value. The logic of defaulting is handled by the invariant that 0 means default. If there is an error, we can log it and default to `shader_ids[0]`. Thoughts?

src/graphic/Fast3D/gfx_pc.h
```diff
@@ -118,9 +119,22 @@ struct RawTexMetadata {
 };
 
 struct ShaderMod {
-    bool enabled = false;
-    int16_t id;
-    uint8_t type;
+    const char* vertex;
```
Rozelette: These can be strings since they're always returned as strings anyways. I think that means you can default `operator==`, as discussed earlier.

src/graphic/Fast3D/gfx_pc.h
```diff
@@ -118,9 +119,22 @@ struct RawTexMetadata {
...
+    const char* fragment;
+
+    std::optional<std::string> GetVertex() const {
```
Rozelette: Why use `std::optional` here, when we can use empty string to mean "no shader is set"?

src/graphic/Fast3D/gfx_pc.h
```diff
@@ -259,5 +273,8 @@ int32_t gfx_check_image_signature(const char* imgData);
 void gfx_register_blended_texture(const char* name, uint8_t* mask, uint8_t* replacement = nullptr);
 void gfx_unregister_blended_texture(const char* name);
 const char* GfxGetOpcodeName(int8_t opcode);
-
 #endif
+
+#ifdef __cplusplus
```
Rozelette: This header is already c++ only. This guard is not need.

Also, put this function inside the `#endif` above. That is the end of the header guard and should always be the last line of the file.

src/graphic/Fast3D/gfx_pc.cpp
```diff
@@ -2946,34 +2939,46 @@ bool gfx_movemem_handler_otr(F3DGfx** cmd0) {
...
-    std::shared_ptr<Ship::ResourceInitData> initData = std::make_shared<Ship::ResourceInitData>();
-    initData->Path = path;
-    initData->IsCustom = false;
-    initData->ByteOrder = Ship::Endianness::Native;
-    auto shader = Ship::Context::GetInstance()->GetResourceManager()->GetArchiveManager()->LoadFile(path);
-    if (shader == nullptr || !shader->IsLoaded) {
-        return -1;
+int16_t gfx_search_existing_shader(const char* vertex, const char* fragment) {
+    ShaderMod shader = { vertex, fragment };
+    for (int i = 0; i < shader_ids.size(); i++) {
```
Rozelette: I am weary of a linear search. This array will grow in size, especially for longer play sessions. Consider using a map. If we have a LRU cache in LUS, this may be a good place for it.

Rozelette (changes requested):
This is a good step towards custom shaders!

Some out-of-scope-for-this-PR future steps:

- The set/unset interface should be push/pop, since otherwise a DL can erase the context that it is running in. DLs are nested and commonly reused, so they lack the context to know how to restore the previous custom shader if they set a new one. push/pop fixes that problem. As we discussed, I will implement this later.

- While currently users only provide the name of the shader, this will likely be insufficient for many use-cases. Just like we have shader options that enable/disable portions of the default shader, users with want to pass shader options for their own shader. We will also need to know what inputs and outputs they use. We can likely extend the `SetShader` command to accept a pointer to a data structure with all the relevant info, rather than just being a pointer to a shader name.

- Currently, the shader id is 16 bits, giving us 64k shaders (32k if negatives aren't allowed) over the lifetime of the program. This may not be enough for long play sessions or heavy graphic mods. I don't think this is an immediate problem, but we should consider the code lift to split off the shader id from the rest of `cc_options`.

- We are bundling the vertex and fragment shader in the shader id. This may cause unnecessary shader pipeline changes. `cc_options` is intended for the color combiner (i.e. the fragment shader), so changing the vertex shader shouldn't have to affect it. We should consider what it would take to keep vertex and fragment shader state separate.

src/graphic/Fast3D/gfx_pc.cpp
MegaMech: How many shaders do you expect a game to implement? Wouldn't a linear search generally be faster than a map within a hundred to a few hundred elements? And if this is true, we probably want the more performant option since this is graphics/shader side.

Rozelette: This is an rather old comment, so I'm not too sure of the context now. I can easily see custom shaders getting to the several hundred mark, especially as permutations accumulate. It's impossible to fully know without usage data. The "more performant option" would be the LRU cache I mentioned, since I assume the set of shaders used on a frame-to-frame basis is largely unchanged.

PR #840, opened 2025-03-20
Improve compile times

NEstelami:
This PR improves compile times by making use of forward declarations, as well as moving header-only libraries includes from .h to .cpp files where applicable.

This has, in my testing, reduced compile times by several seconds per .cpp file.


KiritoDv:
It looks like its just missing some includes on gfx_metal_shader and it should be nice

louist103:
clang-tidy suggestions are invalid so I will not be applying them.

PR #841, opened 2025-03-20
Use *Buffered* RawInput

Spodi:
This should prevent performance issues with high polling rate mice.


Kenix3 (approved):
Merge pending conflict resolution

Spodi:
why was that even a conflict? should be fixed now

PR #844, opened 2025-03-22
try to see what's going on with windows ci

briaguya0:


briaguya0:
LUS windows CI is failing for everything since prism landed.

prism requires spdlog but the vcpkg install of it seems to not be happening until after we include prism in cmake (prism is in https://github.com/briaguya-ai/libultraship/blob/main/cmake/dependencies/common.cmake but the vcpkg stuff is in https://github.com/briaguya-ai/libultraship/blob/main/cmake/dependencies/windows.cmake and we include common before windows https://github.com/briaguya-ai/libultraship/blob/6636c2cf4d43c5e161ab04be73217c86de944f52/CMakeLists.txt#L33-L55 (for good reason, we find packages in windows that are depend on things from common)

PRd a fix https://github.com/Kenix3/libultraship/pull/845

PR #849, opened 2025-03-22
remove initdata in File struct

coco875:


briaguya0:
working in my local testing, soh testing PR here https://github.com/HarbourMasters/Shipwright/pull/5188

PR #850, opened 2025-03-23
fix tidy-result-publish workflow failing

Spodi:
apparently the API path changed


Spodi:
how do i even test this properly? as it still uses the old settings here...
and it doesn't run at all on my fork

Spodi:
This seems to work now. Or at least it isn't failing and sending emails because of that.
Tested here:
https://github.com/Spodi/libultraship/actions/runs/14020946513

What should it even do?

PR #851, opened 2025-03-23
Add Writefile and GetArchiveFromFile for Archives

MegaMech:
This allows writing files into the o2r file. The purpose is to add some user edited metadata.
GetArchiveFromFile allows retrieving the o2r file to write to using the file string.

The only thing currently missing is adding the new file into the file system. Suggestions?
For my purposes, I can just restart the game. Decent compromise.

Tested and it works

### How To Use
```cpp
auto archive = GameEngine::Instance->context->GetResourceManager()->GetArchiveManager()->GetArchiveFromFile(filePath);

if (archive) {
    GameEngine::Instance->context->GetResourceManager()->GetArchiveManager()->WriteFile(archive, "file_name_to_write", some_data)
}
```

### DIFF
--- a/src/resource/archive/Archive.h
+++ b/src/resource/archive/Archive.h
@@ -38,6 +38,7 @@ class Archive : public std::enable_shared_from_this<Archive> {
 
     virtual bool Open() = 0;
     virtual bool Close() = 0;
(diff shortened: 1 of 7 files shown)


Kenix3:
I'm wondering if it may be best to keep the Archive on the file because this is a map lookup. Either way, no need to block merge for it.

PR #855, opened 2025-03-28
Simplify Mac Fullscreen and improve fullscreen handling

Archez:
This aims to add a few fullscreen improvements:
* Drops windowed fullscreen as an option for mac in favor of only using the native fullscreen behavior.
* Supports booting into fullscreen on Mac when using Metal
* Move applying windowed size/position until after the fullscreen change is applied (necessary for mac native behavior)
* Fullscreening and mouse capture was not correctly handling the new modern menu (was checking for a CVar that wasn't used by SoH). This was updated to instead use the exposed visibility method that accounts for menubar or menu at the same time.
* `SaveWindowToConfig` during a fullscreen toggle was saving the fullscreen state before it was toggled, meaning that during runtime the config value would always be opposite. I've added an additional config set in the fullscreen callback to then set the new fullscreen enabled value.
* On Mac, it is possible to fullscreen using the Green traffic light button. To avoid having fullscreen state de-synced with Fast3D, I've added an Apple only check in HandleEvents that checks for the fullscreen state and re-executes the fullscreen callback.

---

Soh Testing PR https://github.com/HarbourMasters/Shipwright/pull/5242

### DIFF
--- a/src/graphic/Fast3D/Fast3dWindow.cpp
+++ b/src/graphic/Fast3D/Fast3dWindow.cpp
@@ -263,8 +263,7 @@ uint32_t Fast3dWindow::GetCurrentRefreshRate() {
 }
 
 bool Fast3dWindow::SupportsWindowedFullscreen() {
-    if (GetWindowBackend() == Ship::WindowBackend::FAST3D_SDL_OPENGL ||
-        GetWindowBackend() == Ship::WindowBackend::FAST3D_SDL_METAL) {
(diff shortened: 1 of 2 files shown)


Archez:
These changes have been tested on Mac, Windows and Linux for all three renderers.

larsy1995:
The comments on line 246 and 247 in gfx_sdl2.cpp are no longer relevant to the current implementation.

PR #863, opened 2025-04-21
Add `ListDirectories()` and `unordered_set<std::string>::mDirectories` to ArchiveManager

MegaMech:
Purpose:

A way to check if paths exists. For instance checking if a type of mod exists:
actors/some_name
actors/some_other_name
maps/mymapname

Then you can check if certain files exists in those directories like so:
```cpp
void FindMaps() {
    auto manager = GameEngine::Instance->context->GetResourceManager()->GetArchiveManager();

    auto ptr = manager->ListDirectories("maps/*");
    if (ptr) {
        auto dirs = *ptr;
    
        for (const std::string& dir : dirs) {
            if (manager->HasFile(dir+"/scene.json")) {
                printf("MAP: %s HAS SCENE FILE\n", dir.c_str());
            } else {
                printf("MAP: %s MISSING SCENE FILE\n", dir.c_str());
            }
        }
}
```

It is possible to do this without adding this function. But it's not very efficient and requires a lot of string modification.

This results in an easy way to check for valid maps and add them to the map list. And for the invalid maps to be added to the editor for the user to initialize the map.

### DIFF
--- a/src/resource/archive/ArchiveManager.h
+++ b/src/resource/archive/ArchiveManager.h
@@ -38,7 +38,7 @@ class ArchiveManager {
     std::shared_ptr<std::vector<std::string>> ListFiles(const std::string& searchMask = "");
     std::shared_ptr<std::vector<std::string>> ListFiles(const std::list<std::string>& includes,
                                                         const std::list<std::string>& excludes);
-    std::shared_ptr<std::vector<std::string>> ListDirectories(const std::string& pattern = "");
+    std::shared_ptr<std::vector<std::string>> ListDirectories(const std::string& searchMask = "");
     std::vector<uint32_t> GetGameVersions();
     const std::string* HashToString(uint64_t hash) const;
     bool IsGameVersionValid(uint32_t gameVersion);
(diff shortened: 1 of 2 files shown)


src/resource/archive/ArchiveManager.h
```diff
@@ -38,6 +38,7 @@ class ArchiveManager {
     std::shared_ptr<std::vector<std::string>> ListFiles(const std::string& searchMask = "");
     std::shared_ptr<std::vector<std::string>> ListFiles(const std::list<std::string>& includes,
                                                         const std::list<std::string>& excludes);
+    std::shared_ptr<std::vector<std::string>> ListDirectories(const std::string& pattern);
```
briaguya0:
```suggestion
    std::shared_ptr<std::vector<std::string>> ListDirectories(const std::string& pattern = "");
```

i figure we want to support someone calling `ListDirectories()` without a param

briaguya0: also the name "pattern" feels a little ambiguous. for `ListFiles` we use `searchMask` which seems like a good name here too

Kenix3 (approved):

PR #864, opened 2025-04-22
O2rArchive: simplify WriteFile

serprex:


src/resource/archive/O2rArchive.cpp
```diff
@@ -111,23 +111,12 @@ bool O2rArchive::WriteFile(const std::string& filename, const std::vector<uint8_
...
+    if (zip_file_add(mZipArchive, filename.c_str(), source, ZIP_FL_ENC_UTF_8 | ZIP_FL_OVERWRITE) < 0) {
+        SPDLOG_ERROR("Failed to add file \"{}\" to ZIP", filename);
+        zip_source_free(source);
+        return false;
     }
-    printf("Success wrote file\n");
+    SPDLOG_INFO("Success wrote file");
 
     // Save changes to disk
     if (zip_close(mZipArchive) < 0) {
```
serprex, line 121: I have someone hitting this a lot on Windows, would be nice if error message had more info

Other issue is that when this function fails without being able to check `mZipArchive` being null or not the caller can't know whether they need to close archive or not. In particular it isn't clear to me that when close fails (but might succeed later) whether the zip source is now invalid if the vector doesn't live beyond end of call to WriteFile

MegaMech: Which game is already making use of the writefile method? This function is new of like, last week?

MegaMech: Delete the first printf line

replace the second one with
SPDLOG_INFO("Successfully wrote file: {}", filename);

MegaMech: It's also worth mentioning, if other ports are using WriteFile note that it is not supported in the OTR format, only O2R.

MegaMech: So, here's the general use-case:

You have an o2r mod that the port needs to add some sort of config file to.

You write the file, and then the function closes the O2R to finalize the write. Then we open it again so the game can continue using the O2R file. This means that the internal game data will need to be refreshed.
You probably need to call `ResetVirtualFileSystem()` after.

The o2r file should already be open.

If you want to write to an o2r file that's not open, then presumably you would need to manually close it after.

MegaMech: I actually thought I did call `ResetVirtualFileSystem` but I probably need to add that.

MegaMech: Finally, you're correct that the WriteFile function in ArchiveManager does not return the correct boolean if the function fails. I'll PR a fix for that.

MegaMech (approved):

src/resource/archive/O2rArchive.cpp
serprex: https://github.com/HarbourMasters/Shipwright/pull/5435 I'm using this API while trying to update SOH's accessibility build to latest

Reset isn't needed since the workflow is `turn on extraction` -> `reset soh` -> `extraction runs` -> `reset soh` -> `it works`

It works for me on Linux, generating 250MB o2r. & it seemed to work for Malkierian on Windows. But for others they're failing to write many of the files & end up with an incomplete archive that's a tenth the size it should be

serprex: re archive manager return, that's not what I'm talking about

https://libzip.org/documentation/zip_close.html

> If successful, archive is freed. Otherwise archive is left unchanged and must still be freed.

When this function returns false the caller isn't able to tell whether that's from `zip_close` failing or `zip_open` failing. The state of `mZipArchive` is indeterminate. I would like to be able to apply some kind of retry logic in case this is being caused by Windows file locking issues

MegaMech: Sure, I'll look into that.

Why are you using this function during extraction? Shouldn't the extractor be generating the o2r's? Not the game itself? The use of this function isn't a requirement on any port for it to work?

MegaMech:
Is this sufficient?
```cpp
if (zip_close(mZipArchive) < 0) {
    zip_error_t* error = zip_get_error(mZipArchive);
    SPDLOG_ERROR("Failed to save changes to ZIP archive: {} ({})",
                zip_error_strerror(error),
                zip_error_code_zip(error));
    zip_discard(mZipArchive); // Optional, if not already discarded
    return false;
}
```
This will output the specific error to the console.

Or do you also want this?
```cpp
    enum class WriteError {
       SUCCESS,
       FAILED_TO_OPEN,
       FAILED_TO_WRITE,
    };
```
I could return this enum.

src/resource/archive/O2rArchive.cpp
serprex: not requirement. this design choice was made by original creators of this experiment, I'd like to move away from the current design

relevant discussion: https://github.com/HarbourMasters/Shipwright/pull/3207#discussion_r1328327428

serprex:
@MegaMech that would be sufficient

MegaMech:
Please see:
https://github.com/Kenix3/libultraship/pull/867

It includes the requested changes.

Note that I added ResetVirtualFileSystem for the reason in the PR description.

If your use-case differs we'll need to discuss a solution because for my use-case Reset is required.

Prior, if WriteFile failed, it was still returning true. This is now fixed.

serprex:
I get a sense that I should just use libzip to write the o2r file myself for my use case

MegaMech:
I could add a `WriteFile(RefreshFileSystem true/false)` argument.

If you need an error enum output for WriteFile I can also do that so you can do a more intensive "Retry" function.

This is a brand new function. So it makes sense that is has querks that need working out. I just didn't expect someone would actually need it.

MegaMech:
Another idea, is that ArchiveManager.cpp could have retry functionality build into it. (or o2rArchive.cpp)

serprex:
That'd be nice, but this class is in an odd spot since you want to keep read potential. If I were using libzip myself I could skip the close/open song & dance, since in the end I just want to open zip -> write a thousand files -> close zip

MegaMech:
We should ask Kenix. Maybe my version of writefile should be some sort of WriteFileInArchive or something.

Kenix3 (approved):

PR #867, opened 2025-04-23
O2rArchive::WriteFile() Better Error Checking

### LINKED ISSUE #882: add file to VFS on write instead of calling `ResetVirtualFileSystem`
based on this conversation https://github.com/Kenix3/libultraship/pull/867/files#r2074327506

MegaMech:
* Adds ResetVirtualFileSystem()

Without this, lus recollection of the files in the archive will be outdated and not include the files that were added to the archive.

The downside of this, is that multiple file writes in a row would result in a refresh that may not be desirable. Feel free to suggest an alternative solution to this. Note that the refresh function is protected (we can't call it outside of the class)

Requires and based on #864 

Resolves https://github.com/Kenix3/libultraship/issues/882

### DIFF
--- a/src/resource/archive/ArchiveManager.cpp
+++ b/src/resource/archive/ArchiveManager.cpp
@@ -140,11 +140,13 @@ void ArchiveManager::ResetVirtualFileSystem() {
     }
 }
 
-bool ArchiveManager::WriteFile(std::shared_ptr<Archive> archive, const std::string& filename,
+bool ArchiveManager::WriteFile(std::shared_ptr<Archive> archive, const std::string& filePath,
                                const std::vector<uint8_t>& data) {
     if (archive) {
-        if (archive->WriteFile(filename, data)) {
-            ResetVirtualFileSystem();
+        if (archive->WriteFile(filePath, data)) {
+            IndexFile(filePath);
+            mFileToArchive[CRC64(filePath.c_str())] = archive;
+            LoadFile(filePath);
             return true; // Successfully wrote file
         }
     }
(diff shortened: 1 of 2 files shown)


serprex (approved):

src/resource/archive/ArchiveManager.cpp
```diff
@@ -143,10 +143,12 @@ void ArchiveManager::ResetVirtualFileSystem() {
 bool ArchiveManager::WriteFile(std::shared_ptr<Archive> archive, const std::string& filename,
                                const std::vector<uint8_t>& data) {
     if (archive) {
-        archive->WriteFile(filename, data);
-        return true;
+        if (archive->WriteFile(filename, data)) {
+            ResetVirtualFileSystem();
```
Kenix3: Instead of resetting the VFS, you can probably just add the new file to the VFS.

MegaMech: What if the file already exists?

MegaMech: And how do I add the new file?

Kenix3: You can add the file by adding it to the VFS map. If it already exists in the VFS no need to do anything.

briaguya0 (approved):
i'm happy to land this as is

i created a follow up issue to replace the `ResetVirtualFileSystem` call https://github.com/Kenix3/libultraship/issues/882

once this is updated to include latest main i'll merge it

Kenix3 (approved):

PR #884, opened 2025-05-17
bump imgui to `v1.91.9b-docking`

briaguya0:
~i verified [the patch](https://github.com/Kenix3/libultraship/blob/main/cmake/dependencies/patches/imgui-fixes-and-config.patch) still works on the new version~

ran into an issue with the mac build, tracked it down to https://github.com/ocornut/imgui/commit/e6a7c7689f57038b2519fe43e55e3d57103ad0f7, added a revert to the patch for now

context on the mac build issue:

me on discord:
> looks like the error is
> ```
> imgui-src/backends/imgui_impl_metal.mm:175:134: error: 'autorelease' is unavailable: not available in automatic reference counting mode
> ```
> 
> which seems to have come from <https://github.com/ocornut/imgui/commit/e6a7c7689f57038b2519fe43e55e3d57103ad0f7>
> 
> i'm not sure what "automatic reference counting mode" is in this context, where it's being set, or if we should be using it

me responding to myself on discord:
> > where it's being set
>
> looks like <https://github.com/Kenix3/libultraship/blob/7135d1494e74c38aa7fca734b92ec50bba543b44/src/CMakeLists.txt#L247-L249> (has been around since <https://github.com/Kenix3/libultraship/pull/80>)
> 
> > if we should be using it
> 
> based on <https://github.com/Kenix3/libultraship/blob/7135d1494e74c38aa7fca734b92ec50bba543b44/src/graphic/Fast3D/backends/gfx_metal.cpp#L74>,  and <https://developer.apple.com/documentation/foundation/nsautoreleasepool> it looks like we're doing at least some things the non ARC way, but my gut feeling is turning it off may cause more harm than good

soh testing pr:
* https://github.com/HarbourMasters/Shipwright/pull/5501

### DIFF
--- a/cmake/dependencies/common.cmake
+++ b/cmake/dependencies/common.cmake
@@ -9,7 +9,7 @@ set(imgui_apply_patch_command ${CMAKE_COMMAND} -Dpatch_file=${imgui_fixes_and_co
 FetchContent_Declare(
     ImGui
     GIT_REPOSITORY https://github.com/ocornut/imgui.git
-    GIT_TAG v1.91.6-docking
+    GIT_TAG v1.91.9b-docking
     PATCH_COMMAND ${imgui_apply_patch_command}
 )
 FetchContent_MakeAvailable(ImGui)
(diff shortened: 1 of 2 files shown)


briaguya0:
image

PR #888, opened 2025-05-21
clean imgui memory

coco875:


src/resource/archive/O2rArchive.cpp
```diff
@@ -10,6 +10,7 @@ O2rArchive::O2rArchive(const std::string& archivePath) : Archive(archivePath) {
 
 O2rArchive::~O2rArchive() {
     SPDLOG_TRACE("destruct o2rarchive: {}", GetPath());
+    Close();
```
briaguya0: `Close` just has

```cpp
bool O2rArchive::Close() {
    if (zip_close(mZipArchive) == -1) {
        SPDLOG_ERROR("Failed to close zip file \"{}\"", GetPath());
        return false;
    }
    return true;
}
```

i think we should probably add a null check for `mZipArchive` into close so we don't get failed to close errors if we never opened a zip file

coco875:
to avoid ImGui to make a double free on font this:
```cpp
std::shared_ptr<Ship::Font> fontData = std::static_pointer_cast<Ship::Font>(
            Ship::Context::GetInstance()->GetResourceManager()->LoadResource(fontPath, false, initData));
```
need to be change to:
```cpp
auto file = Ship::Context::GetInstance()->GetResourceManager()->LoadFileProcess(fontPath);
std::shared_ptr<Ship::Font> fontData = std::static_pointer_cast<Ship::Font>(
            Ship::Context::GetInstance()->GetResourceManager()->GetResourceLoader()->LoadResource(fontPath, file, initData));
```
it allow to allocate one font per size and let imgui free it at the end

src/resource/archive/O2rArchive.cpp
coco875: hmm if mZipArchive are null then I guess zip_close will return -1, I will double check that

src/window/gui/Gui.cpp
```diff
@@ -93,6 +93,16 @@ Gui::Gui() : Gui(std::vector<std::shared_ptr<GuiWindow>>()) {
...
+    if (mImGuiIo != nullptr) {
+        if (mImGuiIo->IniFilename != nullptr) {
+            delete[] mImGuiIo->IniFilename;
+        }
+        if (mImGuiIo->LogFilename != nullptr) {
+            delete[] mImGuiIo->LogFilename;
+        }
+        mImGuiIo->IniFilename = nullptr;
+        mImGuiIo->LogFilename = nullptr;
+    }
```
briaguya0, line 105: do we need to do this manually, does [`ImGui::Shutdown`](https://github.com/ocornut/imgui/blob/5f0acadf7db1b6d469f0771284e2e3f7818443ce/imgui.cpp#L4185) not handle it?

src/resource/archive/O2rArchive.cpp
coco875: try it and yeah it's the case.

src/window/gui/Gui.cpp
coco875: not it's allocate with new here https://github.com/Kenix3/libultraship/blob/7ad23ad94ded59952cc586ad66bf05f4699cb94a/src/window/gui/Gui.cpp#L127-L128
so it's doesn't free if we don't say so.

src/resource/archive/O2rArchive.cpp
```diff
@@ -89,7 +90,7 @@ bool O2rArchive::Open() {
 }
 
 bool O2rArchive::Close() {
-    if (zip_close(mZipArchive) == -1) {
+    if (mZipArchive != nullptr && zip_close(mZipArchive) == -1) {
```
briaguya0: i was thinking something like

```suggestion
    if (mZipArchive == nullptr) {
        SPDLOG_ERROR("Cannot close zip file. Zip file not loaded. \"{}\"", GetPath());
        return false;
     }
     
     if (zip_close(mZipArchive) == -1) {
```

coco875: oh ok

briaguya0 (commented):
can you split the o2r fixes out into a separate PR so we can land those and then figure out the imgui stuff?

coco875:
create https://github.com/Kenix3/libultraship/pull/889

src/window/gui/Gui.cpp
KiritoDv: That looks awful, just do an strdup

Kenix3: Coco didn't change that code. They're just fixing a memory leak.

src/window/gui/resource/Font.cpp
```diff
@@ -5,9 +5,7 @@ Font::Font() : Resource(std::shared_ptr<ResourceInitData>()) {
 }
 
 Font::~Font() {
-    if (Data != nullptr) {
-        delete[] Data;
-    }
+    // let ImGui free the font data
```
Kenix3: Not sure I love the idea of a factory allocating memory for a resource, then the resource individually not being responsible for that memory. That's the pattern in all other cases.

Can you convince me, or maybe call the imgui free functions directly?

src/window/gui/Gui.cpp
Kenix3: Don't love the fact we aren't using strings here, but that's out of scope

src/window/gui/resource/Font.cpp
coco875: the imgui free are just free wrap in a function of free, and while the font are in imgui, it's imgui who try to free it, so we can't free in the class else it will be make a double free error, and don't really know how to remove a font in imgui.

coco875:
Note: port will probably need to load font for imgui like this:
```cpp
std::shared_ptr<Ship::Font> fontData = std::static_pointer_cast<Ship::Font>(
            Ship::Context::GetInstance()->GetResourceManager()->LoadResource(fontPath, false, initData));
        char* fontDataPtr = (char*)malloc(fontData->DataSize);
        memcpy(fontDataPtr, fontData->Data, fontData->DataSize);
        font = mImGuiIo->Fonts->AddFontFromMemoryTTF(fontDataPtr, fontData->DataSize, size);
```

PR #891, opened 2025-05-23
Fix macOS wait timing.

ReddestDream:
This uses the correct timing value for macOS's scheduler. Thanks to @larsy1995 for the fix!


ReddestDream:
Closed literally due to tidy format so I can figure out a new message that it doesn't want split over two lines. 🙄

Issue #879, opened 2025-05-14
Add a way to define joystick deadzone defaults for every port.


sonicdcer:
The joystick deadzone defaults are currently hardcoded to 20 % in the LUS library, forcing ports to use that value. Games such as SF64 have their own deadzone in code which is less than 20 %, losing some control range and requiring the users to manually change it.

briaguya-ai:
i do want to note that in-game deadzones will be entirely separate from LUS deadzones, as in-game deadzones interact with the n64 control stick values while LUS deadzones interact with the SDL values (before they are translated to n64 values)

that being said, this should be a reasonably straightforward feature to add

briaguya-ai:
i think this is actually a good candidate for a cmake variable

it's currently just a hardcoded define

https://github.com/Kenix3/libultraship/blob/7135d1494e74c38aa7fca734b92ec50bba543b44/src/controller/controldevice/controller/ControllerStick.h#L12

but we could follow the pattern from https://github.com/Kenix3/libultraship/blob/main/cmake/cvars.cmake for it

PR #893, opened 2025-05-23
Set Default DeadZone to 0

sonicdcer:
This should be 0 at least until the configuration for it is added.

https://github.com/Kenix3/libultraship/issues/879


briaguya-ai (changes requested):
as I said before, this is a good candidate for using cmake vars

PR #894, opened 2025-05-24
make EXCLUDE_MPQ_SUPPORT optional

coco875:


briaguya0:
i don't think we need this. we use set cache everywhere else

from https://cmake.org/cmake/help/latest/command/set.html#set-cache-entry:
>  Sets the given cache <variable> (cache entry). Since cache entries are meant to provide user-settable values this does not overwrite existing cache entries by default. Use the FORCE option to overwrite existing entries.

Kenix3:
I don't think we need this anymore with the change to make StormLib opt-in. Anyone have qualms with closing this?

briaguya0:
closing it seems reasonable to me

PR #896, opened 2025-05-29
fix dx11 memleak

coco875:


Malkierian (approved):
Tested on 2ship as well, while charging a Deku bubble. Increases RAM by about 5MB/s without this fix, fix doesn't break effects.

Issue #526, opened 2024-04-27
standardize header guards


briaguya0:
we have `#pragma once` and `#ifndef THING_H`

i think we should standardize on `#ifndef THING_H`

KiritoDv:
I vote to use #pragma once since its cleaner

PR #897, opened 2025-06-01
Use pragma once everywhere

louist103:
#526

### DIFF
--- a/include/libultraship/libultra/gbi.h
+++ b/include/libultraship/libultra/gbi.h
@@ -1,37 +1,4 @@
 #pragma once
-
-#ifndef u8
-#define u8 uint8_t
-#endif
-
-#ifndef u16
-#define u16 uint16_t
-#endif
-
-#ifndef u32
-#define u32 uint32_t
-#endif
-
-#ifndef u64
-#define u64 uint64_t
-#endif
-
-#ifndef s8
-#define s8 int8_t
-#endif
-
-#ifndef s16
-#define s16 int16_t
-#endif
-
-#ifndef s32
-#define s32 int32_t
-#endif
-
-#ifndef s64
-#define s64 int64_t
-#endif
-
 #include "mbi.h"
 
 /* To enable Fast3DEX grucode support, define F3DEX_GBI. */
--- a/include/libultraship/libultra/gbi.h
+++ b/include/libultraship/libultra/gbi.h
@@ -1,4 +1,39 @@
 #pragma once
+
+#ifdef _MSC_VER
+#ifndef u8
+#define u8 uint8_t
+#endif
+
+#ifndef u16
+#define u16 uint16_t
+#endif
+
+#ifndef u32
+#define u32 uint32_t
+#endif
+
+#ifndef u64
+#define u64 uint64_t
+#endif
+
+#ifndef s8
+#define s8 int8_t
+#endif
+
+#ifndef s16
+#define s16 int16_t
+#endif
+
+#ifndef s32
+#define s32 int32_t
+#endif
+
+#ifndef s64
+#define s64 int64_t
+#endif
+#endif
+
 #include "mbi.h"
 
 /* To enable Fast3DEX grucode support, define F3DEX_GBI. */
--- a/include/libultraship/libultra/gbi.h
+++ b/include/libultraship/libultra/gbi.h
@@ -1,5 +1,37 @@
 #pragma once
 
+#ifndef u8
+#define u8 uint8_t
+#endif
+
+#ifndef u16
+#define u16 uint16_t
+#endif
+
+#ifndef u32
+#define u32 uint32_t
+#endif
+
+#ifndef u64
+#define u64 uint64_t
+#endif
+
+#ifndef s8
+#define s8 int8_t
+#endif
+
+#ifndef s16
+#define s16 int16_t
+#endif
+
+#ifndef s32
+#define s32 int32_t
+#endif
+
+#ifndef s64
+#define s64 int64_t
+#endif
+
 #include "mbi.h"
 
 /* To enable Fast3DEX grucode support, define F3DEX_GBI. */
--- a/include/libultraship/libultra/gbi.h
+++ b/include/libultraship/libultra/gbi.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "mbi.h"
+
 #ifdef _MSC_VER
 #ifndef u8
 #define u8 uint8_t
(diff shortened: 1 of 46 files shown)


Malkierian:
I swear I'd been told by someone on one of my PRs on either SoH or 2ship (I think Archez?) that `#pragma once` shouldn't be used because it's not part of a standard, and thus not guaranteed to be supported on various compilers, but I've had PRs accepted to both that used pragma rather than ifdefs, which means anything we've supported so far hasn't had an issue with it, so I don't know what to make of it XD.

briaguya0:
people still argue about it https://stackoverflow.com/questions/5776910/what-does-pragma-once-mean-in-c

in general people think all of these are valid options:
* only use pragma
* only use ifndef
* use pragma with ifndef as a fallback

i don't really care and think standardizing would be good instead of having it be a file-by-file thing

as for why soh and 2ship use ifndef, i'm pretty sure a lot of it comes from wanting to match the patterns used in the decomps (which are ifndef only, no pragma)

include/libultraship/libultra/gbi.h
```diff
@@ -1,41 +1,6 @@
...
-#endif
-
-#ifndef s32
-#define s32 int32_t
-#endif
-
-#ifndef s64
-#define s64 int64_t
-#endif
-#endif
```
briaguya0: was this intentional? if so it's not mentioned in the PR description and this goes beyond header guard stuff

louist103:
> people still argue about it https://stackoverflow.com/questions/5776910/what-does-pragma-once-mean-in-c
> 
> in general people think all of these are valid options:
> 
>     * only use pragma
> 
>     * only use ifndef
> 
>     * use pragma with ifndef as a fallback
> 
> 
> i don't really care and think standardizing would be good instead of having it be a file-by-file thing
> 
> as for why soh and 2ship use ifndef, i'm pretty sure a lot of it comes from wanting to match the patterns used in the decomps (which are ifndef only, no pragma)

Decomp uses guards because IDO doesn't support the pragma. I tried it and it just crashes.
At this point  think its safe to assume any compiler someone would use supports the pragma.  GCC supports it going all the way back to 3.4 and MSVC since 4.2

include/libultraship/libultra/gbi.h
louist103: They were pointless since the other headers defined these types already. I can revert if though.

briaguya0: 2 thoughts:
* something somewhere in some port might be relying on those defines being there
* it's not part of standardizing on pragma, so if we want to remove those it should happen in a separate PR

Malkierian:
> I swear I'd been told by someone on one of my PRs on either SoH or 2ship (I think Archez?) that `#pragma once` shouldn't be used because it's not part of a standard, and thus not guaranteed to be supported on various compilers, but I've had PRs accepted to both that used pragma rather than ifdefs, which means anything we've supported so far hasn't had an issue with it, so I don't know what to make of it XD.

Found it: https://github.com/HarbourMasters/Shipwright/pull/5428#discussion_r2051357491
Was actually pretty recent.

briaguya0:
> > I swear I'd been told by someone on one of my PRs on either SoH or 2ship (I think Archez?) that `#pragma once` shouldn't be used because it's not part of a standard, and thus not guaranteed to be supported on various compilers, but I've had PRs accepted to both that used pragma rather than ifdefs, which means anything we've supported so far hasn't had an issue with it, so I don't know what to make of it XD.
> 
> Found it: [HarbourMasters/Shipwright#5428 (comment)](https://github.com/HarbourMasters/Shipwright/pull/5428#discussion_r2051357491) Was actually pretty recent.

it's come up on multiple PRs over the years. it's true that it's not part of the standard (which is mentioned in the stackoverflow post i linked), but it's also true that it's supported by every compiler we support (https://github.com/HarbourMasters/Shipwright/pull/3132, https://github.com/HarbourMasters/Shipwright/pull/3445#discussion_r1405224157 (i know there are more but those showed up with a quick search for "pragma")

include/libultraship/libultra/interrupt.h
```diff
@@ -1,6 +1,5 @@
-#ifndef ULTRA64_INTERRUPT_H
-#define ULTRA64_INTERRUPT_H
+#pragma once
 
-typedef u32 OSIntMask;
+#include <stdint.h>
```
briaguya0, line 3: why is this being added?

src/graphic/Fast3D/backends/gfx_dxgi.h
```diff
@@ -1,7 +1,5 @@
-#ifndef GFX_DXGI_H
-#define GFX_DXGI_H
 #if defined(ENABLE_DX11) || defined(ENABLE_DX12)
-
+#pragma once
```
briaguya0, line 1: this is happening after `#if defined(ENABLE_DX11) || defined(ENABLE_DX12)` when the header guards were before it?

src/public/bridge/gfxdebuggerbridge.h
```diff
@@ -1,11 +1,5 @@
 #pragma once
 
-#ifndef GFXDEBUGGER_H
-#define GFXDEBUGGER_H
-
-#include "stdint.h"
-#include "stddef.h"
```
briaguya0, line 7: why are these being removed?

briaguya0 (commented):
most of this looks great, couple comments in there where it's not clear why changes are being made

include/libultraship/libultra/interrupt.h
louist103: This was also causing clang-tidy to error

src/public/bridge/gfxdebuggerbridge.h
louist103: There was nothing in the file using them.

include/libultraship/libultra/interrupt.h
briaguya0: had you changed it from `u32` to `uint32_t`? if you keep it as `u32` does clang-tidy complain?

src/public/bridge/gfxdebuggerbridge.h
```diff
@@ -1,15 +1,12 @@
...
-
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+#include <stddef.h>
+#include <stdint.h>
```
briaguya0: these are in the extern C now when they weren't before?

include/libultraship/libultra/interrupt.h
louist103: u32 is what it was complaining about. tidy said it was an undefined type.

briaguya0 (approved):

PR #902, opened 2025-06-20
set `FontDataOwnedByAtlas` in `ImFontConfig` to `false` in `GameOverlay::LoadFont()`

Malkierian:
Following <https://github.com/Kenix3/libultraship/pull/888#issuecomment-2899476493> when trying to update SoH, I was still seeing crashes on close that looked like the same issue, and realized that the process hadn't been applied to `GameOverlay::LoadFont()`, but after some investigation, I found out that the memcpy method coco outlined was unnecessary. All we needed was to pass an `ImFontConfig` with `FontDataOwnedByAtlas` set to false, which prevents ImGui from trying to free the data in its own shutdown.

### DIFF
--- a/src/window/gui/GameOverlay.cpp
+++ b/src/window/gui/GameOverlay.cpp
@@ -33,7 +33,13 @@ void GameOverlay::LoadFont(const std::string& name, float fontSize, const Resour
         return;
     }
 
-    mFonts[name] = io.Fonts->AddFontFromMemoryTTF(font->Data, font->DataSize, fontSize);
+    char* fontDataPtr = (char*)malloc(font->DataSize);
+    if (fontDataPtr != NULL) {
+        memcpy(fontDataPtr, font->Data, font->DataSize);
+        mFonts[name] = io.Fonts->AddFontFromMemoryTTF(fontDataPtr, font->DataSize, fontSize);
+    } else {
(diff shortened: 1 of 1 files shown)


briaguya0 (approved):

briaguya0:
context on the fix

from [the imgui docs](https://github.com/ocornut/imgui/blob/ed7d965818f9a7d04de0d5ec76ea9809b49a78e7/docs/FONTS.md#loading-font-data-from-memory):
> ## Loading Font Data from Memory
> 
> ```cpp
> ImFont* font = io.Fonts->AddFontFromMemoryTTF(data, data_size, size_pixels, ...);
> ```
> 
> IMPORTANT: `AddFontFromMemoryTTF()` by default transfer ownership of the data buffer to the font atlas, which will attempt to free it on destruction.
> This was to avoid an unnecessary copy, and is perhaps not a good API (a future version will redesign it).
> If you want to keep ownership of the data and free it yourself, you need to clear the `FontDataOwnedByAtlas` field:
> 
> ```cpp
> ImFontConfig font_cfg;
> font_cfg.FontDataOwnedByAtlas = false;
> ImFont* font = io.Fonts->AddFontFromMemoryTTF(data, data_size, size_pixels, &font_cfg);
> ```

PR #909, opened 2025-08-02
NullAudioEngine: replace SDL as fallback

serprex:
Avoids infinite recursion when SDL init fails


Kenix3:
Make a quick and dirty null audio player. Just extend from the same class as SDLAudioPlayer with methods that "do nothing." After that, if failing to load the requested engine, load the null engine.

serprex:
Done & tested

Kenix3 (approved):

PR #911, opened 2025-08-09
Add basic GameCube controller adapter support.

hal609:
I wanted to get my GC controller working with ship on my Mac and there aren't really any good methods to set them up as HID devices so I had a go at hacking in native support. 

I pulled GCAdapter.cpp directly out of Dolphin, which I think is OK to do under their licence but I'm not totally sure, and then went about trying to map that into the existing structure. I tried to touch as little of the existing code as possible and I copied the formatting from what was already there as much as I could. That said, it proved a lot harder than I thought it would and I ended up needing to ask AI about a tonne of stuff, so the code quality might be kinda iffy. I can say though that it _works on my machine_™ and meshes with the existing deadzone and sensitivity stuff just fine.

Also, this is my first time contributing to an open source project, so I imagine there will be plenty of issues but I'd really appreciate any feedback and I'm happy for anyone else to take it further or to have a go at make any necessary changes myself.


Malkierian:
I'm not sure if this will be added, though I'm not in charge of that, just thought I'd mention that we were waiting to be able to upgrade to SDL3 before adding native GC adapter support because there are tools built in to make that much more feasible.

That being said, though, if it works with SDL2, and we don't know when SDL3 will be cleared for us (we're waiting on SDL_net for SDL3), then maybe it'll be allowed as a stopgap. The Dolphin code is something Kenix or briaguya will have to weigh in on.

Kenix3:
I'm good with it landing until SDL3 is ready. Can you link to the license that applies to the code you copied? Finally we should credit the people we copied from in the main readme of the project. Not required, just something I like to do.

hal609:
OK, nice. Yeah sure, it looks like each file in Dolphin has its own licence listed at the top of the file. The files I used are these: 

https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/InputCommon/GCAdapter.cpp
https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/InputCommon/GCAdapter.h 

which both use "GPL-2.0-or-later" -> https://spdx.org/licenses/GPL-2.0-or-later.html.

Kenix3:
GPL-2 or later is unfortunately not compatible with our license and goals. Due to this issue, the PR cannot land.

hal609:
😭 I wish I knew that before I spent so long yesterday getting it to build on Windows 💀 Ah well, it is what it is. Tbh, it actually hardly needs the Dolphin code at all cause it just needs to set up a libusb device and get the data, but if SDL3's gonna supersede it anyway it's doesn't seem worth it to replace it.

I guess I can licence my fork under GPL and put the compiled binaries on there in case anyone is super desperate to use the adapter right now.

Kinda sad to have wasted my time on it but I suppose it's a lesson to check this stuff before I get too invested.

Should I just close the PR now ?

KiritoDv:
Actually, another thing you could do is ask for special permissions to relicense the code but it means asking multiple people that were involved on the development of that impl if they accept

PR #912, opened 2025-08-10
Fix special character paths on Windows

Eblo:
#869 causes [2ship to not boot if the path contains special characters](https://github.com/HarbourMasters/2ship2harkinian/issues/1220). It seems like `GetModuleFileNameA` does not play well with special characters. This PR changes that to `GetModuleFileNameW` and treats the path as a wide string, then converts it back to a regular string for the return.

This maintains the behavior of locating the O2R when run from a different directory while also allowing for special characters in the path. I'd like to know if @lepideble has any thoughts on this as the author of the aforementioned PR.

https://stackoverflow.com/questions/4804298/how-to-convert-wstring-into-string
https://stackoverflow.com/questions/215963/how-do-you-properly-use-widechartomultibyte
https://learn.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-widechartomultibyte

### DIFF
--- a/src/Context.cpp
+++ b/src/Context.cpp
@@ -15,6 +15,7 @@
 #include <libloaderapi.h>
 #include <tchar.h>
 #include <windows.h>
+#include <stringapiset.h>
 #endif
 
 #ifdef __APPLE__
--- a/src/Context.cpp
+++ b/src/Context.cpp
@@ -435,19 +435,24 @@ std::string Context::GetAppBundlePath() {
 #endif
 
 #ifdef _WIN32
-    std::string progpath(MAX_PATH, '\0');
+    std::wstring progpath(MAX_PATH, '\0');
 
-    int len = GetModuleFileNameA(NULL, &progpath[0], progpath.size());
+    int len = GetModuleFileNameW(NULL, &progpath[0], progpath.size());
     if (len != 0 && len < progpath.size()) {
         progpath.resize(len);
 
         // Find the last '\' and remove everything after it
-        long unsigned int lastSlash = progpath.find_last_of("\\");
+        long unsigned int lastSlash = progpath.find_last_of('\\');
         if (lastSlash != std::string::npos) {
             progpath.erase(lastSlash);
         }
 
-        return progpath;
+        // Convert wstring to string
+        len = WideCharToMultiByte(CP_UTF8, 0, progpath.data(), (int)progpath.size(), nullptr, 0, nullptr, nullptr);
+        std::string newProgpath(len, 0);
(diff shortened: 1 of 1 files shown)


lepideble:
My first try of https://github.com/Kenix3/libultraship/pull/869 actually used `GetModuleFileNameW` but for some reason that I don't remember the conversion from `wstring` to `string` didn't work.

I don't know the Windows API well enough to comment on the code since but if your change work it's good for me.

lepideble:
Oh actually I did test my changes with a special character in the path (my username has accents) so either something else is causing the problem or it's only some special characters that cause the problem and not all of them.

Eblo:
> either something else is causing the problem or it's only some special characters that cause the problem and not all of them

For my testing, I tried putting 2ship in a folder named `blahé`. It would fail to boot entirely with no fanfare until my changes here.

I just tried the same with SoH. It boots, only to fail to find its OTR and freeze. I tried it again with this change, but the result is the same. [2ship had a change to call `setlocale` at boot](https://github.com/HarbourMasters/2ship2harkinian/pull/743), while SoH didn't. SoH might be failing before this change comes into effect.

lepideble:
When doing my tests for my PR I was testing with SoH in a folder that had 2 accents (an `é` and an `ô`) and it did work correctly, maybe this is something related to windows internal encoding that can change base of the version and configuration of the system.

githubshizzler:
I can confirm it does not launch on my end. I have a ü in my username. 
Workaround for me is to move the folder to the root or to C:\Users\Public
Runs fine after that.

src/Context.cpp
```diff
@@ -435,19 +435,24 @@ std::string Context::GetAppBundlePath() {
...
         // Find the last '\' and remove everything after it
-        long unsigned int lastSlash = progpath.find_last_of("\\");
+        long unsigned int lastSlash = progpath.find_last_of('\\');
         if (lastSlash != std::string::npos) {
             progpath.erase(lastSlash);
         }
 
-        return progpath;
+        // Convert wstring to string
+        len = WideCharToMultiByte(CP_UTF8, 0, progpath.data(), (int)progpath.size(), nullptr, 0, nullptr, nullptr);
```
briaguya0, line 452: it looks like `WideCharToMultiByte` comes from `stringapiset.h` (https://learn.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-widechartomultibyte)

i assume that header is making it in to this file somehow, but knowing how annoying dealing with windows headers can be i think it'd be good to explicitly add it to the includes in this file

Kenix3 (approved):

Issue #874, opened 2025-05-11
Frametime spikes on KeyUp


Spodi:
I identified some performance impact when releasing keyboard keys and tracked it down to `GetConfig` when checking scan codes.

https://github.com/Kenix3/libultraship/blob/ce38cb6883ca030e61749eae584b6d9a7cb7bca5/src/graphic/Fast3D/Fast3dWindow.cpp#L321
https://github.com/Kenix3/libultraship/blob/ce38cb6883ca030e61749eae584b6d9a7cb7bca5/src/graphic/Fast3D/Fast3dWindow.cpp#L326

Mashing multiple keys at once and in quick succession can bring down FPS and I guess it is more severe on low end systems.

Spodi:
Probably related: https://github.com/Kenix3/libultraship/pull/918

PR #919, opened 2025-09-23
KeyUp Variables Quick Fix

Malkierian:
So apparently, static class members in the header are illegal. Somehow, this wasn't caught in local testing or LUS Actions. This removes static from the variables, privates them, and adds getters/setters for them for use with the `Ship::Context::GetInstance()->GetWindow()` chain.


briaguya0:
if the original plan was to have the static vars in protected wouldn't you want the getters and setters to be in protected instead of public?

Malkierian:
No, the getters and setters need to be accessible port-side, and they always eventually needed to be added, I was just trying to take a shortcut with the internal referencing (talking in reference to being able to customize them later).

Spodi:
Does this potentially fix https://github.com/Kenix3/libultraship/issues/874?

Malkierian:
This is a followup to #918, which is what was intended to address that.

Issue #926, opened 2025-10-09
fix includes not following heirarchy


briaguya0:
### `libultraship` can include:
* `ship`
* `fast`
* `libultraship`

### `fast` can include:
* `ship`
* `fast`

### `ship` can include:
* `ship`

### `fast` including `libultraship`
_____

https://github.com/Kenix3/libultraship/blob/2782e554d961be13db8d862c6ab48839b82f30ee/src/fast/backends/gfx_metal.cpp#L37

https://github.com/Kenix3/libultraship/blob/2782e554d961be13db8d862c6ab48839b82f30ee/src/fast/resource/factory/DisplayListFactory.cpp#L4

https://github.com/Kenix3/libultraship/blob/2782e554d961be13db8d862c6ab48839b82f30ee/src/fast/resource/factory/VertexFactory.cpp#L4

https://github.com/Kenix3/libultraship/blob/2782e554d961be13db8d862c6ab48839b82f30ee/src/fast/resource/type/Vertex.cpp#L2

https://github.com/Kenix3/libultraship/blob/2782e554d961be13db8d862c6ab48839b82f30ee/include/fast/resource/type/DisplayList.h#L6

https://github.com/Kenix3/libultraship/blob/5f4be9b6f5f74917c303ab8b66a0b2f4ef91613d/src/fast/interpreter.cpp#L38
_____

<details>

<summary>Addressed in #932</summary>

https://github.com/Kenix3/libultraship/blob/2782e554d961be13db8d862c6ab48839b82f30ee/include/fast/interpreter.h#L14

https://github.com/Kenix3/libultraship/blob/2782e554d961be13db8d862c6ab48839b82f30ee/include/fast/resource/type/Matrix.h#L4

https://github.com/Kenix3/libultraship/blob/2782e554d961be13db8d862c6ab48839b82f30ee/src/fast/interpreter.cpp#L24

</details>

<details>

<summary>Addressed in #937</summary>

https://github.com/Kenix3/libultraship/blob/2782e554d961be13db8d862c6ab48839b82f30ee/src/fast/interpreter.cpp#L36

see https://github.com/Kenix3/libultraship/issues/933

</details>

_____

### `ship` including `fast`

https://github.com/Kenix3/libultraship/blob/2782e554d961be13db8d862c6ab48839b82f30ee/include/ship/window/gui/Gui.h#L22

[...]

PR #928, opened 2025-10-09
clean up includes (part 2)

briaguya0:

these ones can just be removed


briaguya0:
from @Kenix3 on discord:
> I can't green check right now, but all four of those PRs can land.

PR #937, opened 2025-10-10
clean up resourcebridge

### LINKED ISSUE #933: clean up includes (bridge used for resources)
`fast/interpreter.cpp` has

https://github.com/Kenix3/libultraship/blob/2782e554d961be13db8d862c6ab48839b82f30ee/src/fast/interpreter.cpp#L36

without it things like

https://github.com/Kenix3/libultraship/blob/2782e554d961be13db8d862c6ab48839b82f30ee/src/fast/interpreter.cpp#L2712

break

there's quite a bit of logic that lives in the bridge such as

https://github.com/Kenix3/libultraship/blob/2782e554d961be13db8d862c6ab48839b82f30ee/src/libultraship/bridge/resourcebridge.cpp#L29-L33

____

in the original header PR, i replaced a ton of `CVarGet...` etc. with `Ship::Context::GetInstance()->GetConsoleVariables()->Get...`, as `consolevariablebridge.cpp` truly is just a bridge (externing class methods so they can be used in c code).

i think we should probably move the logic from `resourcebridge.cpp` into classes (archive manager/archive/resource manager/resource), make it a true bridge, and directly call the class methods from the `fast` code

briaguya0:
resolves https://github.com/Kenix3/libultraship/issues/933

### DIFF
--- a/src/fast/interpreter.cpp
+++ b/src/fast/interpreter.cpp
@@ -2763,7 +2763,8 @@ bool gfx_mtx_otr_filepath_handler_custom_f3dex2(F3DGfx** cmd0) {
     Interpreter* gfx = mInstance.lock().get();
     F3DGfx* cmd = *cmd0;
     const char* fileName = (const char*)cmd->words.w1;
-    const int32_t* mtx = (const int32_t*)Ship::Context::GetInstance()->GetResourceManager()->GetResourceRawPointer((const char*)fileName);
+    const int32_t* mtx = (const int32_t*)Ship::Context::GetInstance()->GetResourceManager()->GetResourceRawPointer(
+        (const char*)fileName);
 
     if (mtx != NULL) {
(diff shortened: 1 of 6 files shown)


src/fast/interpreter.cpp
```diff
@@ -2708,8 +2708,6 @@ bool gfx_marker_handler_otr(F3DGfx** cmd0) {
     Interpreter* gfx = mInstance.lock().get();
     (*cmd0)++;
     F3DGfx* cmd = (*cmd0);
-    const uint64_t hash = ((uint64_t)(cmd)->words.w0 << 32) + (cmd)->words.w1;
-    std::string dlName = ResourceGetNameByCrc(hash);
```
briaguya0, line 2712: this just isn't being used, not sure why it was here so i just removed it

PR #941, opened 2025-10-14
clean up includes (fast from Gui)

briaguya0:

not everything but these parts can be removed cleanly

soh test run https://github.com/briaguya0/Shipwright/actions/runs/18482421270

### DIFF
--- a/include/fast/Fast3dWindow.h
+++ b/include/fast/Fast3dWindow.h
@@ -1,8 +1,6 @@
 #pragma once
 #include "ship/window/Window.h"
 #include "ship/window/gui/Gui.h"
-#include "fast/backends/gfx_window_manager_api.h"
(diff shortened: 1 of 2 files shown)


include/fast/Fast3dWindow.h
```diff
@@ -1,8 +1,6 @@
 #pragma once
 #include "ship/window/Window.h"
 #include "ship/window/gui/Gui.h"
-#include "fast/backends/gfx_window_manager_api.h"
-#include "fast/backends/gfx_rendering_api.h"
```
briaguya0, line 5: `Fast3dWindow` is still being included in `Gui`. removing these to prove the direct includes in `Gui` can *actually* be removed and things aren't just working because they're still being included here.

briaguya0:
image

PR #943, opened 2025-10-15
move button names from ship to LUS

### LINKED ISSUE #930: clean up includes (`BUTTON_BITMASKS`)
https://github.com/Kenix3/libultraship/blob/2782e554d961be13db8d862c6ab48839b82f30ee/include/ship/controller/controldevice/controller/ControllerButton.h#L12-L14

relies on the button macros from

https://github.com/Kenix3/libultraship/blob/2782e554d961be13db8d862c6ab48839b82f30ee/include/libultraship/libultra/controller.h#L63-L85

removing

https://github.com/Kenix3/libultraship/blob/2782e554d961be13db8d862c6ab48839b82f30ee/include/ship/controller/controldevice/controller/ControllerButton.h#L7

and trying to hack it away with something like

```cpp
#ifndef BUTTON_BITMASKS
#define BUTTON_BITMASKS 0x1, 0x2
#endif
```

leads to issues in the cpp file

https://github.com/Kenix3/libultraship/blob/2782e554d961be13db8d862c6ab48839b82f30ee/src/ship/controller/controldevice/controller/ControllerButton.cpp#L25-L61

briaguya0:
edit: moved to draft because i want to do the inputeditorwindow stuff as part of this PR to inform where things should live

resolves https://github.com/Kenix3/libultraship/issues/930

### DIFF
--- a/include/ship/controller/controldevice/controller/ControllerButton.h
+++ b/include/ship/controller/controldevice/controller/ControllerButton.h
@@ -10,8 +10,7 @@ namespace Ship {
 
 class ControllerButton {
   public:
-    ControllerButton(uint8_t portIndex, CONTROLLERBUTTONS_T bitmask,
-                     std::unordered_map<CONTROLLERBUTTONS_T, std::string> buttonNames);
+    ControllerButton(uint8_t portIndex, CONTROLLERBUTTONS_T bitmask);
     ~ControllerButton();
 
     std::shared_ptr<ControllerButtonMapping> GetButtonMappingById(std::string id);
@@ -41,8 +40,6 @@ class ControllerButton {
     uint8_t mPortIndex;
     CONTROLLERBUTTONS_T mBitmask;
     std::unordered_map<std::string, std::shared_ptr<ControllerButtonMapping>> mButtonMappings;
-
-    std::unordered_map<CONTROLLERBUTTONS_T, std::string> mButtonNames;
     std::string GetConfigNameFromBitmask(CONTROLLERBUTTONS_T bitmask);
 
     bool mUseEventInputToCreateNewMapping;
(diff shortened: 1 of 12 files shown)


src/fast/interpreter.cpp
```diff
@@ -35,6 +35,8 @@
 #include "ship/Context.h"
 #include "ship/config/ConsoleVariable.h"
 
+#include "libultraship/libultra/os.h"
```
briaguya0, line 38: this was coming in through controllers somehow, i'll update #926 to add this when this PR lands

src/ship/window/gui/InputEditorWindow.cpp
```diff
@@ -5,6 +5,7 @@
 #include "ship/config/ConsoleVariable.h"
 #include "ship/controller/controldevice/controller/mapping/sdl/SDLAxisDirectionToButtonMapping.h"
 #include "ship/controller/controldeck/ControlDeck.h"
+#include "libultraship/libultra/controller.h"
```
briaguya0, line 8: bitmasks were coming in from somewhere else, will update #926 to include this after this lands

include/ship/controller/controldevice/controller/ControllerButton.h
```diff
@@ -45,6 +41,8 @@ class ControllerButton {
     uint8_t mPortIndex;
     CONTROLLERBUTTONS_T mBitmask;
     std::unordered_map<std::string, std::shared_ptr<ControllerButtonMapping>> mButtonMappings;
+
+    std::unordered_map<CONTROLLERBUTTONS_T, std::string> mButtonNames;
```
briaguya0: maybe this could live on controldeck or controller instead, but having controllerbutton "reach up" to get that feels less than ideal to me (but so does passing it down through a ton of constructors and keeping it on every button)

i guess the other possible refactor here would be to only keep the one string associated with the button on each, but my goal was to limit changes to logic as much as possible and focus purely on the separating lus/fast/ship aspects here

the other thing to consider is how this choice impacts the inputeditorwindow stuff, so it might be worth just trying to pull that work into this PR and letting that inform the final decision

Kenix3 (approved):

PR #944, opened 2025-10-15
Cursor Visibility Tweaks

Malkierian:
It seems the longer the cursor was not visible (and calling `wnd->SetCursorVisibility(false)` each frame), the longer it took for mouse movement to register as far as making the cursor visible again. This tweaks the cursor visibility tick to only call SetCursorVisibility(false) once per visibility cycle (by setting `mCursorVisibleTicks` to -1 after calling it, where it remains until mouse movement registers again).

Also fixes a related typo (ForceCursorVisability -> ForceCursorVisibility).

### DIFF
--- a/src/ship/window/gui/Gui.cpp
+++ b/src/ship/window/gui/Gui.cpp
@@ -619,8 +619,9 @@ void Gui::CursorTimeoutTick() {
         }
         if (mCursorVisibleTicks > 0) {
             mCursorVisibleTicks--;
-        } else {
+        } else if (mCursorVisibleTicks == 0) {
             wnd->SetCursorVisibility(false);
+            mCursorVisibleTicks = -1;
         }
         mPrevMousePos = mousePos;
     }
--- a/src/ship/window/gui/Gui.cpp
+++ b/src/ship/window/gui/Gui.cpp
@@ -610,20 +610,28 @@ void Gui::HandleMouseCapture() {
 
 void Gui::CursorTimeoutTick() {
     auto wnd = std::dynamic_pointer_cast<Fast::Fast3dWindow>(Context::GetInstance()->GetWindow());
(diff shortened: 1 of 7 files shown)


src/ship/window/gui/Gui.cpp
```diff
@@ -616,8 +619,9 @@ void Gui::CursorTimeoutTick() {
         }
         if (mCursorVisibleTicks > 0) {
             mCursorVisibleTicks--;
-        } else {
+        } else if (mCursorVisibleTicks == 0) {
             wnd->SetCursorVisibility(false);
+            mCursorVisibleTicks = -1;
```
briaguya0: how do you feel about changing this method to

```cpp
void Gui::CursorTimeoutTick() {
    auto wnd = std::dynamic_pointer_cast<Fast::Fast3dWindow>(Context::GetInstance()->GetWindow());
    if (wnd->ShouldForceCursorVisibility()) {
        return;
    }

    Ship::Coords mousePos = wnd->GetMousePos();
    bool mouseMoved = abs(mousePos.x - mPrevMousePos.x) > 0 || abs(mousePos.y - mPrevMousePos.y) > 0;
    mPrevMousePos = mousePos;

    if (!wnd->IsMouseCaptured() && mouseMoved) {
        wnd->SetCursorVisibility(true);
        mCursorVisibleTicks = mCursorVisibleSeconds * wnd->GetTargetFps();
        return;
    }

    if (mCursorVisibleTicks == 0) {
        wnd->SetCursorVisibility(false);
        mCursorVisibleTicks = -1;
        return;
    }

    if (mCursorVisibleTicks > 0) {
        mCursorVisibleTicks--;
    }
}
```

it's *almost* logically identical, the only difference is the early return when setting ticks if the mouse moves (so that run we start with the full count instead of decrementing once

briaguya0 (approved):
:shipit:

PR #946, opened 2025-10-19
Mouse Capture Clip Adjustment

Malkierian:
The default SDL capture clip rect appeared to just be the size of the window. DXGI's was set to be just inside the size of the window. This causes issues with auto-hidden taskbars (being able to show them when the hidden cursor still goes to the edge of the window) and docked windows (which would block mouse button clicks from getting to the main game, blocking their input processing) in the ports.
This adds a 2x2 clip rect at the center of the window in SDL when mouse capture is active, and adjusts DXGI's to match.


briaguya0 (approved):
LGTM, holding off on merging for a little so @Kenix3 has a chance to look at it

PR #948, opened 2025-10-25
Check compute shaders compat

Spodi:
Still a draft because I'm unsure about the ThrowWithMessage addition. Also everything feels hacked together somehow... 

Loose follow up to https://github.com/Kenix3/libultraship/pull/497

- fix DEVICE_DRAW_RESOURCE_SAMPLE_COUNT_MISMATCH
 we are already using shader model 4.1 for that, so we can just use the unsized version of Texture2DMS (probably not even needed, but the debug layer shuts up now)

- add ThrowWithMessage
 similar to ThrowIfFaled, but not bound to an actual thing failing (should this be done different?)

- check GPU for ComputeShader support and throw a slightly more useful message if not supported (maybe this solves our feature level mystery?).
EDIT: It does! The error is triggered on Intel HD graphics 3000. 

SoH build: https://github.com/HarbourMasters/Shipwright/pull/5888


Spodi:
Closes in favor of https://github.com/Kenix3/libultraship/pull/967

PR #949, opened 2025-10-27
Gui Only Render

Malkierian:
Add `RunGuiOnly` to `interpreter` to run ImGui interface without game commands.

### DIFF
--- a/include/fast/Fast3dWindow.h
+++ b/include/fast/Fast3dWindow.h
@@ -16,7 +16,7 @@ class Fast3dWindow : public Ship::Window {
 
     void Init() override;
     void Close() override;
-    void RunGuiOnly();
+    void RunGuiOnly() override;
     void StartFrame() override;
     void EndFrame() override;
     bool IsFrameReady() override;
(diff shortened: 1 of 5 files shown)


include/fast/Fast3dWindow.h
```diff
@@ -16,6 +16,7 @@ class Fast3dWindow : public Ship::Window {
 
     void Init() override;
     void Close() override;
+    void RunGuiOnly();
```
briaguya0:
```suggestion
    void RunGuiOnly() override;
```

briaguya0 (commented):
i think this makes sense, the one thing i'd want to change is adding a virtual `RunGuiOnly` to `Ship::Window`instead of only having it on `Fast3dWindow`.

i see 2 ways to accomplish what https://github.com/HarbourMasters/Shipwright/pull/5892/ needs from this PR:
1. do what this PR is doing (make the `RunGuiOnly` method)
2. create a new "gui only" class that inherits from `Ship::Window`

i think it probably makes sense to have *both*. supporting a `RunGuiOnly` "mode" for different backends is probably a good thing, and won't get in the way of adding a "gui only" class in the future.

i'll be happy with this PR landing once `Ship::Window` includes a virtual `RunGuiOnly`method

include/fast/Fast3dWindow.h
Malkierian: Done.

briaguya0 (approved):
:shipit:

PR #954, opened 2025-11-06
Logging Setup Improvements

Malkierian:
With the true nature of SPDLOG cmake defines being revealed, it was determined that, while the definition should still be configurable, it should just default to trace for both debug and release, and the actual printing control should be done primarily in code instead. This changes the  cmake logger defines to set SPDLOG_MIN_CUTOFF with SPDLOG_LEVEL_TRACE as the active level for both debug and release.
In addition to that, this adds debug and release build level parameters to `InitLogging()` with defaults of debug and warn, respectively, uses those in `set_level()` called during initialization, and hardcodes the level at which flush happens to trace for debug builds and info for release builds.


briaguya0 (approved):
:shipit:

PR #956, opened 2025-11-20
Allow writing new blocks to config

Eblo:
Fixes a bug where trying to write a block that does not already exist in the config will fail. Credit @garrettjoecox

### DIFF
--- a/src/ship/config/Config.cpp
+++ b/src/ship/config/Config.cpp
@@ -156,6 +156,8 @@ void Config::SetBlock(const std::string& key, nlohmann::json block) {
     } else {
         if (gjson.contains(key)) {
             gjson[key] = block;
+        } else {
+            gjson.emplace(key, block);
         }
     }
     mFlattenedJson = gjson.flatten();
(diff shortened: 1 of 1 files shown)


Malkierian (approved):

src/ship/config/Config.cpp
```diff
@@ -156,6 +156,8 @@ void Config::SetBlock(const std::string& key, nlohmann::json block) {
     } else {
         if (gjson.contains(key)) {
             gjson[key] = block;
+        } else {
+            gjson.emplace(key, block);
```
briaguya0: why does this need to be in an else block? shouldn't `emplace` work either way? i'd think we could completely remove the `if (gjson.contains(key))` check and just always call emplace here.

Malkierian: emplace only works when the key doesn't exist first.

Malkierian: To be fair, I'm not sure why I checked to see if it contained the key first. We may be able to get away with just using the `gjson[key] = block;` line.

Eblo: Always using the direct assignment seems to work whether there's a key already or not.

PR #957, opened 2025-11-23
Fix ImGui Controller Opts Id Conflict

MegaMech:
Also fixed some mispellings.

### DIFF
--- a/include/ship/window/gui/InputEditorWindow.h
+++ b/include/ship/window/gui/InputEditorWindow.h
@@ -17,7 +17,7 @@ class InputEditorWindow : public GuiWindow {
     virtual ~InputEditorWindow();
 
     void DrawInputChip(const char* buttonName, ImVec4 color);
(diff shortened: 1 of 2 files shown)


include/ship/window/gui/InputEditorWindow.h
```diff
@@ -17,7 +17,7 @@ class InputEditorWindow : public GuiWindow {
     virtual ~InputEditorWindow();
 
     void DrawInputChip(const char* buttonName, ImVec4 color);
-    void DrawAnalogPreview(const char* label, ImVec2 stick, float deadzone = 0, bool gyro = false);
+    void DrawAnaloguePreview(const char* label, ImVec2 stick, float deadzone = 0, bool gyro = false);
```
briaguya0: everything else uses US english ("color" instead of "colour" etc.), so this shouldn't change

PR #959, opened 2025-12-02
remove shared in ArchiveManager getter and setter

coco875:
Remove shared because it's useless to have it in this context

### DIFF
--- a/include/ship/resource/archive/ArchiveManager.h
+++ b/include/ship/resource/archive/ArchiveManager.h
@@ -22,8 +22,8 @@ class ArchiveManager {
 
     std::shared_ptr<Archive> AddArchive(const std::string& archivePath);
     std::shared_ptr<Archive> AddArchive(std::shared_ptr<Archive> archive);
-    std::shared_ptr<std::vector<std::shared_ptr<Archive>>> GetArchives();
-    void SetArchives(std::shared_ptr<std::vector<std::shared_ptr<Archive>>> archives);
+    std::vector<std::shared_ptr<Archive>> GetArchives();
+    void SetArchives(std::vector<std::shared_ptr<Archive>> archives);
     size_t RemoveArchive(std::shared_ptr<Archive> archive);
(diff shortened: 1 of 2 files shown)


Kenix3:
This would do a copy on return. Why is this the preferred behaviour?

briaguya0:
*Investigation assisted by, response written by Claude (claude.ai) on behalf of the commenter.*

The return value won't actually be copied in practice due to two layers of C++ optimization:

**1. NRVO (Named Return Value Optimization)**

`archives` is a named local variable returned directly, which qualifies for NRVO. From [cppreference — Copy elision](https://en.cppreference.com/w/cpp/language/copy_elision):

> In a return statement in a function with a class return type, when the operand is the name of a non-volatile object with automatic storage duration... the copy-initialization of the result object can be omitted by constructing obj directly into the function call's result object. This variant of copy elision is known as *named return value optimization* (NRVO).

The vector is constructed directly in the caller's storage — no copy occurs.

**2. Implicit move as a fallback**

Even if NRVO is not applied, the same page guarantees:

> If the operand of a return statement is a (possibly parenthesized) id-expression that names an implicitly movable entity... overload resolution to select the constructor for the copy is first performed as if the operand were an rvalue.

So at worst the vector is *moved* (O(1) pointer swap), not copied.

As for the `shared_ptr` wrapper — is there a use case where callers need shared ownership of the archives list? It's worth noting that even the original implementation builds a **new vector** on every call rather than returning a `shared_ptr` to the internal `mArchives`, so any caller holding onto the result already has a snapshot, not a live view. The `shared_ptr` wrapper isn't providing shared ownership of the real data either way. The individual `shared_ptr<Archive>` elements inside remain shared regardless.

Looking at both libultraship and consuming repos (Shipwright, 2ship2harkinian, Starship), `GetArchives()` doesn't appear to be called anywhere — so it's hard to identify a scenario where the `shared_ptr` wrapping is or would be load-bearing.

PR #960, opened 2025-12-02
configure other controller automatically

### LINKED ISSUE #436: Assign gamepads to different players by default
Edit: in the discussion below it's been pointed out that there is a way to configure multiple player inputs manually. This worked for me in the end. However, I would like to amend this issue: the _default behaviour_ should be to distribute the gamepads among the players. See "expected behaviour" below.

Edit 2: This issue originally had "(Steam Deck)" in the title. I have now removed it since apparently it's a problem on other OS's as well. However, I'd like to specifically mention that Steam Deck has the additional problem that there's no way to input an Esc key press using the (manually-activated) on-screen keyboard, so a Steam Deck user is especially unlikely to discover the escape menu. This could be improved by letting the backtick/tilde key open the menu, since that is available via the on-screen keyboard, and is commonly used in other games as a secondary escape key (e.g. to open a developer console.)

Original issue report:

## Context

I'm using the AppImage Steam Deck release, `Alfredo Alfa 0.9.9.1`, and I'm on a steam deck. I've tried both desktop mode and big picture mode. And I've also tried enabling and disabling steam input in big picture mode. In all cases, the result is the same.

## Observed behaviour

All gamepads control player 1, and players 2-4 are unable to select their characters.

## Expected behaviour

From the player select menu onward, if there are ≤ 4 gamepads available, then each gamepad should be mapped to a different player's input, not all to the same player. If there are ≥ 4 gamepads available, then all players should have at least one corresponding gamepad.

**Edit:** Note that the Steam Deck already has an OS-level method to remap gamepads to different virtual slots which are made available to the game. So it's perfectly okay to assign the gamepads to players in an arbitrary order, as the player can remap these easily at the steam level. (`(Steam Button) -> Controllers`)

coco875:
fix https://github.com/HarbourMasters/SpaghettiKart/issues/436


briaguya0 (changes requested):
this doesn't resolve the linked issue. this makes it so all controllers are configured for all ports from the start (meaning pressing "A" on any controller will lead to all 4 ports registering an A press)

coco875:
ok was thinking that mPortIndex give to Controller also influence sdl my bad

PR #961, opened 2025-12-02
remove cmake code who have no effect

coco875:


briaguya0 (commented):
instead of removing these lines shouldn't we just fix the paths? so instead of the broken `graphic/Fast3D/backends/gfx_dxgi*` we'd have `backends/gfx_dxgi*` (since it's relative to `src/fast` now)

coco875:
It's also useless because we have put #ifdef in file to avoid adding their code, or you would like to remove the ifdef and instead remove file from compilation ?

briaguya0:
> It's also useless because we have put #ifdef in file to avoid adding their code, or you would like to remove the ifdef and instead remove file from compilation ?

the ifdef is an extra safety measure (some ports end up including files from lus in their cmakelists in messy ways), but we should definitely remove the files from compilation by default on platforms they aren't used on

PR #964, opened 2025-12-04
Configure other controller

coco875:
finally find the way to make each controller assign to one port controller


briaguya0 (changes requested):
this needs to be configurable instead of just changing the behavior. for single player games being able to pick up any controller and use it is important (especially considering we don't have control over device order in SDL, and that people often end up having strange things showing up as controllers on their systems)

i'd also like to dig into all the scenarios in which `RefreshConnectedSDLGamepads` gets called. have you tested disconnecting/reconnecting controllers with this change? my guess is that if someone manually changes their ignore list (by unchecking devices) this will lead to that getting reset

coco875:
yes add and remove controller reset with controller get ignored or not, hmm can probably add an option to ConnectedPhysicalDeviceManager but for 4 players, if sdl detect weird controller then in 4 player it will always make problem, don't know how to handle that

briaguya0:
before https://github.com/Kenix3/libultraship/pull/793 we had the controller reordering window. i think a modified version of that might work, but it wouldn't be simple.

SDL also has a concept of player index that i haven't done any digging into
* https://wiki.libsdl.org/SDL2/SDL_GameControllerFromPlayerIndex, https://wiki.libsdl.org/SDL2/SDL_GameControllerGetPlayerIndex, https://wiki.libsdl.org/SDL2/SDL_GameControllerSetPlayerIndex
* https://wiki.libsdl.org/SDL3/SDL_GetGamepadFromPlayerIndex, https://wiki.libsdl.org/SDL3/SDL_GetGamepadPlayerIndex, https://wiki.libsdl.org/SDL3/SDL_SetGamepadPlayerIndex

overall the UX of "one of the connected controllers is player 1, we have no way of knowing which it will be or ensuring it'll stay the same next time you open the game" is bad

most modern local multiplayer games i've played have it so in the menu all controllers work, then when you start multiplayer people join in by pressing a button (brotato comes to mind as a game that does it well)

a reasonable multiplayer flow imo would be to:
* on boot - all physical controllers work on the first controller port, all physical controllers ignored on all other controller ports
* have a way for the game to tell LUS "hey, we need controllers for other controller ports now" - lus shows a reordering window that sets the ignore state based on the ordering
* have a way for the game to tell LUS "we're back in the main menu, make all the controllers work on port 1 again"

but things get tricky when you start dealing with controllers disconnecting/reconnecting etc.

this isn't a simple feature to add, and it's extremely difficult to test considering how many different controllers people have and how they report their connected status (4 port adapters tend to show up as 4 controllers regardless of if 4 are plugged in or not for example)

coco875:
Hmm for now, there is no way for a temporal in between because for now each time you start the game all controller are on the port 1, not the best for spaghettikart local multiplayer experience, if at least one controller are to one port for an avg controller it will work fine when you start the game, So make the experience less worst for now

MegaMech:
> make the experience less worst for now  

I second this

src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
```diff
@@ -66,13 +75,33 @@ void ConnectedPhysicalDeviceManager::RefreshConnectedSDLGamepads() {
         mConnectedSDLGamepads[instanceId] = gamepad;
         mConnectedSDLGamepadNames[instanceId] = name;
 
+#ifdef ENABLE_EXP_AUTO_CONFIGURE_CONTROLLERS
+        // Check if auto-configure controllers is enabled (default: enabled/1)
+        bool autoConfigureEnabled = Context::GetInstance()->GetConsoleVariables()->GetInteger(CVAR_AUTO_CONFIGURE_CONTROLLERS, 0);
```
briaguya0: this says "`default: enabled/1`" but `GetInteger` is using `0` as the default/fallback value

src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
```diff
@@ -65,9 +75,31 @@ void ConnectedPhysicalDeviceManager::RefreshConnectedSDLGamepads() {
...
+        } else {
+            for (uint8_t p = 1; p < 4; p++) {
+                mIgnoredInstanceIds[p].insert(instanceId);
+            }
+        }
+#else
+        for (uint8_t j = 1; j < 4; j++) {
+            mIgnoredInstanceIds[j].insert(instanceId);
         }
+        port++;
```
briaguya0:
```suggestion
```

this isn't needed here

src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
```diff
@@ -52,6 +55,13 @@ void ConnectedPhysicalDeviceManager::RefreshConnectedSDLGamepads() {
     mConnectedSDLGamepads.clear();
     mConnectedSDLGamepadNames.clear();
 
+#ifdef ENABLE_EXP_AUTO_CONFIGURE_CONTROLLERS
+    for (uint8_t i = 0; i < 4; i++) {
+        mIgnoredInstanceIds[i].clear();
+    }
+#endif
+
+    uint8_t port = 0;
```
briaguya0:
```suggestion
#ifdef ENABLE_EXP_AUTO_CONFIGURE_CONTROLLERS
    for (uint8_t i = 0; i < 4; i++) {
        mIgnoredInstanceIds[i].clear();
    }
    
    uint8_t port = 0;
#endif
```

switching from using `port` to `j` lower down makes the existing logic harder to understand. this makes it so we can keep using `port` there

src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
```diff
@@ -65,9 +75,31 @@ void ConnectedPhysicalDeviceManager::RefreshConnectedSDLGamepads() {
...
+            }
+            port++;
+        } else {
+            for (uint8_t p = 1; p < 4; p++) {
+                mIgnoredInstanceIds[p].insert(instanceId);
+            }
+        }
+#else
+        for (uint8_t j = 1; j < 4; j++) {
+            mIgnoredInstanceIds[j].insert(instanceId);
```
briaguya0: with the above suggestion we can keep this as it was

```suggestion
        for (uint8_t port = 1; port < 4; port++) {
            mIgnoredInstanceIds[port].insert(instanceId);
```

src/ship/window/gui/InputEditorWindow.cpp
```diff
@@ -1393,6 +1394,25 @@ void InputEditorWindow::DrawSetDefaultsButton(uint8_t portIndex) {
...
+                        &autoConfigureControllers)) {
+        Context::GetInstance()->GetConsoleVariables()->SetInteger(CVAR_AUTO_CONFIGURE_CONTROLLERS,
+                                                                  autoConfigureControllers ? 1 : 0);
+        Context::GetInstance()->GetConsoleVariables()->Save();
+        // Refresh connected gamepads to apply the new setting
+        Context::GetInstance()->GetControlDeck()->GetConnectedPhysicalDeviceManager()->RefreshConnectedSDLGamepads();
+    }
+    if (ImGui::IsItemHovered()) {
+        ImGui::SetTooltip("When enabled, each connected controller is automatically assigned to a specific port.\nWhen "
+                          "disabled, all controllers are available on all ports.");
```
briaguya0: this isn't how it works. when it's disabled, all controllers are set to work on the first port (and ignored on the others by default).

src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
```diff
@@ -65,9 +75,31 @@ void ConnectedPhysicalDeviceManager::RefreshConnectedSDLGamepads() {
...
+
+        // Only auto-assign controllers to specific ports if enabled
+        if (autoConfigureEnabled) {
+            for (uint8_t j = 0; j < 4; j++) {
+                if (port == j) {
+                    continue;
+                }
+                mIgnoredInstanceIds[j].insert(instanceId);
+            }
+            port++;
```
briaguya0, line 91: this means when more than 4 controllers are connected any controllers past the first 4 will be ignored on all ports. is that intentional?

CMakeLists.txt
```diff
@@ -26,6 +26,11 @@ set(ADDITIONAL_LIB_INCLUDES "")
 # ========= Configuration Options =========
 option(INCLUDE_MPQ_SUPPORT "Enable StormLib and MPQ archive support" OFF)
 option(GBI_UCODE "Specify the GBI ucode version" F3DEX_GBI_2)
+option(ENABLE_EXP_AUTO_CONFIGURE_CONTROLLERS "Enable experimental auto-configure controllers feature (assign each controller to a specific port)" OFF)
```
briaguya0: i'm not sure what the best wording for this would be, but "assign each controller to a specific port" implies functionality that this doesn't have (such as "setting" a controller to always be on a certain port).

briaguya0 (changes requested):
overall this seems reasonable

i left a couple comments about the code itself, but those should be small changes.


i *do* think we need a better name than "auto configure controllers"

with this _disabled_ controllers _are_ auto-configured, they're just all configured to work on port 1.

this should have a name that properly implies what this does, and descriptions that make it very clear.

src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
coco875: yes because else what do you want to do ? set it to port 1 ? or reroll again so port 1, port 2, port 3, port 4 again ?

CMakeLists.txt
coco875: hmm "Have the possibility to no longer initialise all controller on port 1" ?

src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
coco875: manually doing it

src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
coco875: same as above

src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
coco875: same as above

src/ship/window/gui/InputEditorWindow.cpp
coco875: have change the comment by yours

src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
briaguya0: it's fine, just wanted to make sure it's intentional.

CMakeLists.txt
briaguya0: maybe something like

```suggestion
option(ENABLE_EXP_AUTO_MULTIPLAYER_CONTROLLERS "Experimental: add support for an alternate default controller-to-port routing where the first four SDL controllers use distinct player ports" OFF)
```

coco875: I have made the modification manually

cmake/cvars.cmake
```diff
@@ -19,6 +19,7 @@ set(CVAR_MENU_BAR_OPEN "gOpenMenuBar" CACHE STRING "")
 set(CVAR_PREFIX_CONTROLLERS "gControllers" CACHE STRING "")
 set(CVAR_PREFIX_ADVANCED_RESOLUTION "gAdvancedResolution" CACHE STRING "")
 set(CVAR_AUDIO_CHANNELS_SETTING "gAudioChannelsSetting" CACHE STRING "")
+set(CVAR_AUTO_CONFIGURE_CONTROLLERS "gAutoConfigureControllers" CACHE STRING "")
```
briaguya0: this should also be updated to match the new cmake naming

src/ship/window/gui/InputEditorWindow.cpp
```diff
@@ -1393,6 +1394,25 @@ void InputEditorWindow::DrawSetDefaultsButton(uint8_t portIndex) {
 }
 
 void InputEditorWindow::DrawElement() {
+#ifdef ENABLE_EXP_AUTO_MULTIPLAYER_CONTROLLERS
+    // Checkbox for auto-configure controllers
+    bool autoConfigureControllers =
```
briaguya0: the comment and local var should be updated to match the new cmake naming

CMakeLists.txt
```diff
@@ -26,6 +26,11 @@ set(ADDITIONAL_LIB_INCLUDES "")
 # ========= Configuration Options =========
 option(INCLUDE_MPQ_SUPPORT "Enable StormLib and MPQ archive support" OFF)
 option(GBI_UCODE "Specify the GBI ucode version" F3DEX_GBI_2)
+option(ENABLE_EXP_AUTO_MULTIPLAYER_CONTROLLERS "Experimental: add support for an alternate default controller-to-port routing where the first four SDL controllers use distinct player ports" OFF)
+
+if(ENABLE_EXP_AUTO_MULTIPLAYER_CONTROLLERS)
+    add_compile_definitions(ENABLE_EXP_AUTO_MULTIPLAYER_CONTROLLERS)
+endif()
```
briaguya0: i'm not sure if we should have this here or follow the pattern `GBI_UCODE` uses
https://github.com/Kenix3/libultraship/blob/3ef24c3d1645fe1fc2a7065d1c2e132973ebc151/src/CMakeLists.txt#L109

my main concern is with cmake files getting wild (we have over 1k lines of cmake code in SoH), so keeping the "Configuration Options" section limited to just `option` lines seems nice

coco875: so put this three line before or after ?

briaguya0: i think you can just do
```cmake
 target_compile_definitions(libultraship PRIVATE ${GBI_UCODE} ${ENABLE_EXP_AUTO_MULTIPLAYER_CONTROLLERS}) 
```
but i haven't checked

coco875: have try but didn't so will just put the if below

briaguya0: now i'm curious as to how the `GBI_UCODE` part is working

coco875: by memory it's the port who define it so it will be equivalent to target_compile_definitions(libultraship PRIVATE F3DEX) for example (so we will define F3DEX as a define)

PR #966, opened 2025-12-14
Mobile: Update deps for Android/iOS and add Android CI

SternXD:
iOS builds were failing due to the SDL2 version being too old, which caused a CMake compatibility error with newer CMake versions. This PR updates all of the deps, fixes the affected include paths, re-enables iOS CI, and adds Android CI.


briaguya0 (approved):
LGTM but i'd like to get a 👍 from @Kenix3 before landing this one since it's adding android to CI

Kenix3 (approved):
I'm good with Android being added to CI

PR #967, opened 2025-12-15
Add fallback to WARP (software renderer) for DirectX

Spodi:
This actually changes multiple things for the Directx renderer:
- Graphic adapters (more or less "GPUs") aren't enumerated and checked until one can do at least feature level 10_0. Instead the first one (preferred) is used. This makes the checks easier (only one needs to be tried) and also always respects the "preferred GPU" setting from Windows10 (or driver control panel on older Windows versions).
- Fallback to WARP (software renderer), if the primary adapter can't at least do feature level 10_0, can't do compute shaders / DirectCompute or fails for some other reason.
- More granular error messages why the selected adapter isn't suitable/capable. Those will pop up each start, even when the fallback is successful.
- More logging. Including adapter / GPU name.
- Log when MSAA setting is ignored on old feature level 10_0 hardware. This was the case since some attempts for more compatibility where made, but was never logged. Also apparently there are not many GPUs out there that only support feature level 10_0, but with compute shaders. So this was and will be rarely noticed. But better save than sorry.

I feel like it's ready. Feel free to suggest better messages or anything else that can be improved. 

### Builds:
- Ship of Harkinian: implemented since official release of version 9.2.0
- 2 Ship 2 Harkinian: https://github.com/HarbourMasters/2ship2harkinian/pull/1408
- Ghostship: https://github.com/HarbourMasters/Ghostship/pull/148

### DIFF
--- a/include/fast/backends/gfx_dxgi.h
+++ b/include/fast/backends/gfx_dxgi.h
@@ -51,7 +51,7 @@ class GfxWindowBackendDXGI final : public GfxWindowBackend {
     // These need to be public to be accessible in the window callback
     void CreateSwapChain(IUnknown* mDevice, std::function<void()>&& before_destroy_fn);
     void CreateFactoryAndDevice(bool debug, int d3d_version, class GfxRenderingAPIDX11* self,
-                                bool (*createFunc)(class GfxRenderingAPIDX11* self));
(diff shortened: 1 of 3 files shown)


Ammar-Sadaoui:
out of curiosity is it possible to add DirectX 8 or 9 support and access depth buffer too ?

Kenix3 (approved):

PR #970, opened 2025-12-27
Fix window resize crash on macOS with Metal

koplas:
Previously, the code manually released the texture associated with the current drawable during SetupScreenFramebuffer. However, textures obtained via CAMetalDrawable->texture() are managed by the Core Animation layer. Manually calling release() caused a race condition where the system compositor (CA::Context) would attempt to access a deallocated texture, leading to a "Corrupt NSInvocation" EXC_BREAKPOINT crash.

This fix relies on the CAMetalLayer's internal pool to manage the lifecycle of these textures.

Relevant: https://github.com/HarbourMasters/Shipwright/issues/5764

### DIFF
--- a/src/fast/backends/gfx_metal.cpp
+++ b/src/fast/backends/gfx_metal.cpp
@@ -606,9 +606,6 @@ void GfxRenderingAPIMetal::SetupScreenFramebuffer(uint32_t width, uint32_t heigh
 
     NS::AutoreleasePool* autorelease_pool = NS::AutoreleasePool::alloc()->init();
 
-    if (tex.texture != nullptr)
-        tex.texture->release();
-
     tex.texture = mCurrentDrawable->texture();
 
     MTL::RenderPassDescriptor* render_pass_descriptor = MTL::RenderPassDescriptor::renderPassDescriptor();


ShishuTheDragon:
+1 this fixes the crash for me

serprex:
Does this release need to be conditional? ie, we don't want to run if it came from https://github.com/koplas/libultraship/blob/c2c5e974cc2a80090f48993fc33f484261e1cfd2/src/fast/backends/gfx_metal.cpp#L609 but do if it came from https://github.com/koplas/libultraship/blob/c2c5e974cc2a80090f48993fc33f484261e1cfd2/src/fast/backends/gfx_metal.cpp#L695 ?

in that case a bool could be set on `tex` dictating whether texture needs to be freed or not

koplas:
`UpdateFramebufferParameters` already has an early return, I don't think it is necessary. But adding a check could be good defensive practice, that could avoid issues with future code modifications. I would like to keep this fix simple and I am open for further extensions in other pull requests.

https://github.com/koplas/libultraship/blob/c2c5e974cc2a80090f48993fc33f484261e1cfd2/src/fast/backends/gfx_metal.cpp#L659-L667

koplas:
Would it help to attach screenshots of the Metal debugger?

PR #972, opened 2025-12-28
Add PrismProcessor for Dynamic Shader Loading

MegaMech:
Somehow https://github.com/Kenix3/libultraship/pull/838 broke. So this an updated duplicate of that PR.
Credit to @KiritoDv and Prism team for their work.

## Changes
`RSP { shader_id }` member removed because it is unused

## Review
- [x] Constexpr
- [x] Split vertex/pixel shader
  * Combined into a single shader asset
  * Metal does not support multi-file shaders.
- [x] Add enum member back and rename
image  

- [x] LRU cache
- [x] Remove std::optional
- [x] Remove some use of `auto`
- [x] Decrease Gfx size
- [x] Fix `gfx_get_shader()` to return default of 0 if shader is not found.
image 

- [x] Re-adding the same shader again is ignored. Thus fitting shaders into int16_t limit.


include/fast/lus_gbi.h
```diff
@@ -67,9 +67,10 @@ constexpr int8_t OTR_G_READFB = OPCODE(0x3e);
 constexpr int8_t OTR_G_REGBLENDEDTEX = OPCODE(0x3f);
 constexpr int8_t OTR_G_SETINTENSITY = OPCODE(0x40);
 constexpr int8_t OTR_G_MOVEMEM_HASH = OPCODE(0x42);
-constexpr int8_t OTR_G_LOAD_SHADER = OPCODE(0x43);
-constexpr int8_t RDP_G_SETTILESIZE_INTERP = OPCODE(0x44);
-constexpr int8_t RDP_G_SETTARGETINTERPINDEX = OPCODE(0x45);
+constexpr int8_t OTR_G_PUSH_SHADER = OPCODE(0x43);
```
Rozelette, line 70: What do you think about combining push and pop to be a single command? Where `Push(NULL)` means `Pop`? Alternatively, to use a bit in the first word to mean if it's push or pop, like in `G_MTX`?

include/fast/interpreter.h
```diff
@@ -129,6 +130,10 @@ class GfxRenderingAPI;
 class GfxWindowBackend;
 
 constexpr size_t MAX_SEGMENT_POINTERS = 16;
+constexpr size_t SHADER_ID_SHIFT = 16;
+constexpr int16_t ShaderIdUnmask(int id) {
+    return (id >> SHADER_ID_SHIFT) & 0xFFFF;
```
Rozelette, line 135: `0xFFFF` should be a constant, such as `SHADER_ID_MASK`

include/fast/interpreter.h
```diff
@@ -514,7 +510,11 @@ class Interpreter {
 
     const std::unordered_map<Mtx*, MtxF>* mCurMtxReplacements;
     bool mMarkerOn; // This was originally a debug feature. Now it seems to control s2dex?
-    std::vector<std::string> shader_ids;
+    std::unordered_map<size_t, const char*> mShaders;
```
Rozelette, line 513: Does this need to be a map? Since the key (the shader id) is a linearly increasing int, can't this be a vector instead?

src/fast/interpreter.cpp
```diff
@@ -2898,36 +2895,39 @@ bool gfx_movemem_handler_otr(F3DGfx** cmd0) {
...
-        return -1;
-    }
-    shader_ids.push_back(std::string(shader->Buffer->data()));
-    return shader_ids.size() - 1;
+bool gfx_push_shader(F3DGfx** cmd0) {
+    Interpreter* gfx = mInstance.lock().get();
+    F3DGfx* cmd = *cmd0;
+    const char* shader = (char*)cmd->words.w1;
+
+    gfx->mShaders[gfx->mShadersIndex] = shader;
```
Rozelette, line 2903: Every time a "push shader" command is in a DL, it will add a new entry to `mShaders`. I don't see where it is cleared, or where `mShadersIndex` is reset to 0. Isn't this going to constantly grow for the whole play session? I think we would want to reset it for every frame.

src/fast/interpreter.cpp
```diff
@@ -2898,36 +2895,39 @@ bool gfx_movemem_handler_otr(F3DGfx** cmd0) {
...
-    const auto path = std::string(file);
-    const auto shaderId = gfx->CreateShader(path);
-    gfx->mRdp->current_shader = { true, shaderId, (uint8_t)C0(16, 1) };
     return false;
 }
 
+const char* gfx_get_shader(int16_t id) {
+    Interpreter* gfx = mInstance.lock().get();
+
+    for (const std::pair<size_t, const char*>& shader : gfx->mShaders) {
```
Rozelette, line 2922: This should use `find`

src/fast/interpreter.cpp
```diff
@@ -1483,6 +1482,13 @@ void Interpreter::GfxSpTri1(uint8_t vtx1_idx, uint8_t vtx2_idx, uint8_t vtx3_idx
     if (use_grayscale) {
         cc_options |= SHADER_OPT(GRAYSCALE);
     }
+
+    if (!mShaderStack.empty()) {
```
Rozelette, line 1486: These should use that `SHADER_ID_MASK` to mask the id so we don't affect the upper bits. Especially that `-1`.

src/fast/interpreter.cpp
```diff
@@ -1483,6 +1482,13 @@ void Interpreter::GfxSpTri1(uint8_t vtx1_idx, uint8_t vtx2_idx, uint8_t vtx3_idx
...
+        cc_options |= (mShaderStack.top() << SHADER_ID_SHIFT);
+    } else {
+        cc_options |= -1 << SHADER_ID_SHIFT;
```
Rozelette, line 1489: `-1` should be a constant. Consider `SHADER_USE_DEFAULT`.

src/fast/interpreter.cpp
```diff
@@ -2898,36 +2895,39 @@ bool gfx_movemem_handler_otr(F3DGfx** cmd0) {
...
-    shader_ids.push_back(std::string(shader->Buffer->data()));
-    return shader_ids.size() - 1;
+bool gfx_push_shader(F3DGfx** cmd0) {
+    Interpreter* gfx = mInstance.lock().get();
+    F3DGfx* cmd = *cmd0;
+    const char* shader = (char*)cmd->words.w1;
+
+    gfx->mShaders[gfx->mShadersIndex] = shader;
+    gfx->mShaderStack.push(gfx->mShadersIndex);
+    gfx->mShadersIndex++;
```
Rozelette, line 2905: Consider checking for overflow here. A `constexpr size_t SHADER_ID_MAX = 0xFFFF;` could help.

src/fast/resource/factory/DisplayListFactory.cpp
```diff
@@ -1133,6 +1140,9 @@ ResourceFactoryXMLDisplayListV0::ReadResource(std::shared_ptr<Ship::File> file,
             std::string rawMode1 = child->Attribute("Mode1");
             std::string rawMode2 = child->Attribute("Mode2");
             g = gsDPSetRenderMode(renderModes[rawMode1], renderModes[rawMode2]);
+        } else if (childName == "SetShader") {
```
Rozelette, line 1143: This needs to be updated to the new opcodes

Kenix3:
@Rozelette It seems as though the author has no intention to update based on the feedback. The PR can stand as documentation for future improvements.

PR #973, opened 2025-12-31
add error handling to `ConnectedPhysicalDeviceManager::RefreshConnectedSDLGamepads`

### LINKED ISSUE #976: crash when `SDL_GameControllerName` returns `NULL`
```text
  strlen in minkernel\crts\ucrt\src\appcrt\string\amd64\strlen.asm Line: 76
  Ship::ConnectedPhysicalDeviceManager::RefreshConnectedSDLGamepads in D:\a\2ship2harkinian\2ship2harkinian\libultraship\src\controller\physicaldevice\ConnectedPhysicalDeviceManager.cpp Line: 66
  Ship::SDLAddRemoveDeviceEventHandler::UpdateElement in D:\a\2ship2harkinian\2ship2harkinian\libultraship\src\ship\controller\physicaldevice\SDLAddRemoveDeviceEventHandler.cpp Line: 23
  [...]
  ```

### PR DESCRIPTION
**Pull request by Spodi**
- Fixes https://github.com/Kenix3/libultraship/issues/976
- Uses the GUID as name instead, if that is the case:
  531769358-fa077d06-701a-4cce-895c-48ce356790d9
- Adds error checking at multiple points of failure and skips the device if any of them occur, because it's most likely unusable in those cases.
- Add logging for the errors catched above
### LUS bump PRs for ports:
- SoH: https://github.com/HarbourMasters/Shipwright/pull/6121

### DIFF
--- a/src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
+++ b/src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
@@ -59,6 +59,10 @@ void ConnectedPhysicalDeviceManager::RefreshConnectedSDLGamepads() {
         }
 
         auto gamepad = SDL_GameControllerOpen(i);
+        // skip if it could not be opened (isn't usable)
+        if (!gamepad) {
+            continue;
+        }
         auto instanceId = SDL_JoystickInstanceID(SDL_GameControllerGetJoystick(gamepad));
         auto name = SDL_GameControllerName(gamepad);
 
--- a/src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
+++ b/src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
@@ -1,4 +1,5 @@
 #include "ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.h"
+#include <spdlog/spdlog.h>
 
 namespace Ship {
 ConnectedPhysicalDeviceManager::ConnectedPhysicalDeviceManager() {
@@ -53,22 +54,41 @@ void ConnectedPhysicalDeviceManager::RefreshConnectedSDLGamepads() {
     mConnectedSDLGamepadNames.clear();
 
     for (int32_t i = 0; i < SDL_NumJoysticks(); i++) {
-        // skip if this SDL joystick isn't a Gamepad
+        char deviceGUID[33] = "";
+        SDL_JoystickGetGUIDString(SDL_JoystickGetDeviceGUID(i), deviceGUID, sizeof(deviceGUID));
+
         if (!SDL_IsGameController(i)) {
+            SPDLOG_WARN("SDL Joystick (GUID: {}) not recognized as gamepad."
+                        "This is likely due to a missing mapping string in gamecontrollerdb.txt."
+                        "See https://github.com/mdqinc/SDL_GameControllerDB for more info.",
+                        i, deviceGUID);
             continue;
         }
 
         auto gamepad = SDL_GameControllerOpen(i);
-        // skip if it could not be opened (isn't usable)
-        if (!gamepad) {
+        if (gamepad == nullptr) {
+            SPDLOG_ERROR("SDL GameControllerOpen error (GUID: {}): {}", deviceGUID, SDL_GetError());
             continue;
         }
+
         auto instanceId = SDL_JoystickInstanceID(SDL_GameControllerGetJoystick(gamepad));
+        if (instanceId < 0) {
+            SPDLOG_ERROR("SDL JoystickInstanceID error (GUID: {}): {}", deviceGUID, SDL_GetError());
+            continue;
+        }
+
+        std::string gamepadName;
         auto name = SDL_GameControllerName(gamepad);
+        if (name == nullptr) {
+            gamepadName = "Gamepad " + std::to_string(instanceId);
+            SPDLOG_WARN("SDL_GameControllerName returned null (GUID: {}). Setting name to \"{}\"", deviceGUID,
+                        gamepadName);
+        } else {
+            gamepadName = name;
+        }
 
         mConnectedSDLGamepads[instanceId] = gamepad;
-        // "name" can be NULL and the map only likes strings
-        mConnectedSDLGamepadNames[instanceId] = name ? name : "Unknown " + std::to_string(instanceId); 
+        mConnectedSDLGamepadNames[instanceId] = gamepadName;
 
         for (uint8_t port = 1; port < 4; port++) {
             mIgnoredInstanceIds[port].insert(instanceId);
--- a/src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
+++ b/src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
@@ -55,12 +55,13 @@ void ConnectedPhysicalDeviceManager::RefreshConnectedSDLGamepads() {
 
     for (int32_t i = 0; i < SDL_NumJoysticks(); i++) {
         char deviceGUID[33] = "";
+        // Sould return 00000000000000000000000000000000 if the GUID or controller index is invalid
         SDL_JoystickGetGUIDString(SDL_JoystickGetDeviceGUID(i), deviceGUID, sizeof(deviceGUID));
 
         if (!SDL_IsGameController(i)) {
             SPDLOG_WARN("SDL Joystick (GUID: {}) not recognized as gamepad."
                         "This is likely due to a missing mapping string in gamecontrollerdb.txt."
-                        "See https://github.com/mdqinc/SDL_GameControllerDB for more info.",
+                        "Refer to https://github.com/mdqinc/SDL_GameControllerDB for more information.",
                         i, deviceGUID);
             continue;
         }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp [hunk 1]
```diff
@@ -59,11 +59,16 @@ void ConnectedPhysicalDeviceManager::RefreshConnectedSDLGamepads() {
         }
 
         auto gamepad = SDL_GameControllerOpen(i);
+        // skip if it could not be opened (isn't usable)
+        if (!gamepad) {
+            continue;
+        }
         auto instanceId = SDL_JoystickInstanceID(SDL_GameControllerGetJoystick(gamepad));
```
briaguya0, line 85: for completeness sake it'd be good to check this too

from https://wiki.libsdl.org/SDL2/SDL_JoystickInstanceID, https://wiki.libsdl.org/SDL2/SDL_JoystickID
> The value -1 is an invalid ID.

and we should also log errors instead of just silently continuing

so that'd make this whole thing

```cpp
void ConnectedPhysicalDeviceManager::RefreshConnectedSDLGamepads() {
    mConnectedSDLGamepads.clear();

    for (int32_t i = 0; i < SDL_NumJoysticks(); i++) {
        if (!SDL_IsGameController(i)) {
            // todo: add useful logging here to tell people
            // their joystick needs to be added to gamecontrollerdb 
            continue;
        }

        auto gamepad = SDL_GameControllerOpen(i);
        if (gamepad == nullptr) {
            SPDLOG_ERROR("SDL GameControllerOpen error: %s\n", SDL_GetError());
            continue;
        }

        auto instanceId = SDL_JoystickInstanceID(SDL_GameControllerGetJoystick(gamepad));
        if (instanceId == -1) {
            SPDLOG_ERROR("SDL JoystickInstanceID error: %s\n", SDL_GetError());
            continue;
        }

        std::string gamepadName;
        auto name = SDL_GameControllerName(gamepad);
        if (name == nullptr) {
            // todo: log something about lack of name for gamepad
            // and the fact that we're setting it to "Gamepad ({instanceId})"
            gamepadName = "Gamepad " + std::to_string(instanceId);
        } else {
            gamepadName = name;
        }

        mConnectedSDLGamepadNames[instanceId] = gamepadName;

        for (uint8_t port = 1; port < 4; port++) {
            mIgnoredInstanceIds[port].insert(instanceId);
        }
    }
}
```

and need to add

```diff
--- a/src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
+++ b/src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
@@ -1,4 +1,5 @@
 #include "ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.h"
+#include <spdlog/spdlog.h>
```

Spodi: > ```cpp
> // todo: log something about lack of name for gamepad
> // and the fact that we're setting it to "Gamepad ({instanceId})"
> gamepadName = "Gamepad " + std::to_string(instanceId);
> ```
 So you're for `Gamepad <ID>` instead of of `Unknown <ID>`? :D
 As for the log. How about:
 ```cpp
gamepadName = "Gamepad " + std::to_string(instanceId);
SPDLOG_WARN("Could not get a name for Gamepad. Setting it to \"{}\"", gamepad);  
 ```
&NewLine;
&NewLine;
&NewLine;
> ```cpp
> if (!SDL_IsGameController(i)) {
>    // todo: add useful logging here to tell people
>    // their joystick needs to be added to gamecontrollerdb 
>    continue;
> }
> ```
  
I'm currently not sure what the standard procedure for adding stuff to gamecontrollerdb is. It sais the mapping tool is outdated and I've no clue how to use the other thing - so i can't explain it in a message. https://github.com/mdqinc/SDL_GameControllerDB?tab=readme-ov-file#mapping-tools

briaguya0: > So you're for `Gamepad <ID>` instead of of `Unknown <ID>`? :D

i think so yeah, i'm ok with "unknown" too though

_____

> As for the log. How about:
> 
> ```c++
> gamepadName = "Gamepad " + std::to_string(instanceId);
> SPDLOG_WARN("Could not get a name for Gamepad. Setting it to \"{}\"", gamepad);  
> ```

i think for the logging it'd be good to make it clear that `SDL_GameControllerName` gave us null instead of a name.

also, https://wiki.libsdl.org/SDL2/SDL_GameControllerName#return-value mentions
> Returns the implementation dependent name for the game controller, or NULL if there is no name or the identifier passed is invalid. 

which means we're actually looking at 2 possible states here:
1. controller just doesn't have a name for some reason (perfectly reasonable to just give it a generic name)
2. somehow the instance id is invalid now despite being valid when we called `SDL_JoystickInstanceID` right above this

so maybe

```cpp
gamepadName = "Gamepad " + std::to_string(instanceId);
SPDLOG_WARN("SDL GameControllerName returned null. Set name to \"{}\"", gamepadName);  
```

_____

> I'm currently not sure what the standard procedure for adding stuff to gamecontrollerdb is. It sais the mapping tool is outdated and I've no clue how to use the other thing - so i can't explain it in a message. https://github.com/mdqinc/SDL_GameControllerDB?tab=readme-ov-file#mapping-tools

i don't think we should try to explain *how* to make a mapping string in the log message, just that we don't have one and we need one.

the docs for the db repo used to mention using steam as an option (https://github.com/mdqinc/SDL_GameControllerDB/commit/3266af403894a437fbee282ffd0defbfdd767f25), but linking out to more accessible (aka, don't need to build the tool from SDL source) ways to make mappings is definitely something that should be happening at the db repo level instead of in this log message.

with https://wiki.libsdl.org/SDL2/SDL_JoystickGetDeviceGUID in mind, i'm thinking something like:

```cpp
if (!SDL_IsGameController(i)) {
    SPDLOG_WARN("SDL Joystick ({}) not recognized as gamepad. This is likely due to a missing mapping string in gamecontrollerdb.txt. See https://github.com/mdqinc/SDL_GameControllerDB for more info.", SDL_JoystickGetDeviceGUID(i));
    continue;
}
```

Spodi:
Urg, that's what I get for trying to fix formatting on my phone. Lol.

src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
```diff
@@ -53,17 +54,41 @@ void ConnectedPhysicalDeviceManager::RefreshConnectedSDLGamepads() {
     mConnectedSDLGamepadNames.clear();
 
     for (int32_t i = 0; i < SDL_NumJoysticks(); i++) {
-        // skip if this SDL joystick isn't a Gamepad
+        char deviceGUID[33] = "";
+        SDL_JoystickGetGUIDString(SDL_JoystickGetDeviceGUID(i), deviceGUID, sizeof(deviceGUID));
```
briaguya0: realizing now we should probably error check *these* too

_____

https://wiki.libsdl.org/SDL2/SDL_JoystickGetDeviceGUID

> If called on an invalid index, this function returns a zero GUID.

_____

https://wiki.libsdl.org/SDL2/SDL_JoystickGetGUIDString

this one probably doesn't need error checking assuming we have a nonzero value from `SDL_JoystickGetDeviceGUID`

Spodi: > If called on an invalid index, this function returns a zero GUID.

I don't know how i would check this before it is an actual string. Would it hurt if the GUID still gets passed and shows up as all zeros?

Also maybe instead of Gamepad (ID), we can name a controller without a name just with the GUID (no "gamepad", just the number)? It's a long number, but unique (at least as unique as the names are) and don't change on reconnect. And if it shows as all zeros in the menu, we know something is terribly wrong :D
grafik

src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
Spodi: I think we got that covered now

src/ship/controller/physicaldevice/ConnectedPhysicalDeviceManager.cpp
briaguya0: > I don't know how i would check this before it is an actual string. Would it hurt if the GUID still gets passed and shows up as all zeros?

looks like SDL just uses a static var internally to do this. from https://github.com/libsdl-org/SDL/blob/f432937f9ca8b5a66dda8ad011ffdd605dbe9d02/src/joystick/SDL_gamecontroller.c#L126 and https://github.com/libsdl-org/SDL/blob/f432937f9ca8b5a66dda8ad011ffdd605dbe9d02/src/joystick/SDL_gamecontroller.c#L714-L716

```cpp
static SDL_JoystickGUID s_zeroGUID;
[...]

        if (SDL_memcmp(&mapping->guid, &s_zeroGUID, sizeof(mapping->guid)) == 0) {
            continue;
        }
```

so i think we can probably do

right above the method
```cpp
static SDL_JoystickGUID sZeroGUID;
```

then for the looping part
```cpp
for (int32_t i = 0; i < SDL_NumJoysticks(); i++) {
    SDL_JoystickGUID deviceGUID = SDL_JoystickGetDeviceGUID(i);
    if (SDL_memcmp(&deviceGUID, &s_zeroGUID, sizeof(deviceGUID)) == 0) {
        SPDLOG_WARN("Calling SDL JoystickGetDeviceGUID with index ({}) returned zero GUID. This is likely due to an invalid index. See https://wiki.libsdl.org/SDL2/SDL_JoystickGetDeviceGUID for more info.", i);
        continue;
    }

    char deviceGUIDString[33];
    SDL_JoystickGetGUIDString(deviceGUID, deviceGUIDString, sizeof(deviceGUIDString));
```

Spodi: So skip the device altogether if it's a zero GUID? Well probably doesn't hurt to skip it, cause that should only be happening with an invalid device (index doesn't exist anymore or something like that).

Spodi: implemented the check and skipping devices with a zeroed out GUID now

briaguya0 (approved):

PR #974, opened 2026-01-01
Some rendering fixes

KiritoDv:
The tidy-format is a false positive, it should be ignored since it breaks the gbi file

- It adds back Vtx being as float when GBI_FLOATS is enabled, taken from the original Fast3D Repo
- Fixes G_MOVEMEM_OTR not being declared as an expanded command when it is
- Fixed a bug on F3D that reads invalid mtxs

### DIFF
--- a/src/fast/interpreter.cpp
+++ b/src/fast/interpreter.cpp
@@ -2825,7 +2825,7 @@ bool gfx_mtx_otr_handler_custom_f3d(F3DGfx** cmd0) {
         (const int32_t*)Ship::Context::GetInstance()->GetResourceManager()->GetResourceRawPointer(hash);
     if (mtx != nullptr) {
         cmd--;
-        gfx->GfxSpMatrix(C0(16, 8), (const int32_t*)gfx->SegAddr(cmd->words.w1));
+        gfx->GfxSpMatrix(C0(16, 8), mtx);
         cmd++;
     }
     return false;
--- a/src/fast/resource/factory/DisplayListFactory.cpp
+++ b/src/fast/resource/factory/DisplayListFactory.cpp
@@ -177,7 +177,7 @@ ResourceFactoryBinaryDisplayListV0::ReadResource(std::shared_ptr<Ship::File> fil
 
         int8_t opcode = (int8_t)(command.words.w0 >> 24);
         bool isExpanded = opcode == G_SETTIMG_OTR_HASH || opcode == G_DL_OTR_HASH || opcode == G_VTX_OTR_HASH ||
-                          opcode == G_BRANCH_Z_OTR || opcode == G_MARKER || opcode == G_MTX_OTR;
+                          opcode == G_BRANCH_Z_OTR || opcode == G_MARKER || opcode == G_MTX_OTR || opcode == G_MOVEMEM_OTR;
 
         // These are 128-bit commands, so read an extra 64 bits...
         if (isExpanded) {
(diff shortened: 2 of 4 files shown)


NEstelami (approved):

briaguya0:
i made an soh testing pr for this https://github.com/HarbourMasters/Shipwright/pull/6122

any idea what things should be tested?

KiritoDv:
Nothing really, it only affects games with GBI_FLOATS or that use f3d

briaguya0:
> Nothing really, it only affects games with GBI_FLOATS

that makes sense for the changes in `gbi.h` and `lus_gbi.h`, they're all behind `#ifndef`s

> or that use f3d

could you elaborate on this? don't they *all* use f3d?

KiritoDv:
> > Nothing really, it only affects games with GBI_FLOATS
> 
> that makes sense for the changes in `gbi.h` and `lus_gbi.h`, they're all behind `#ifndef`s
> 
> > or that use f3d
> 
> could you elaborate on this? don't they _all_ use f3d?

Sure, but no, actually sm64 is the only port that uses the 'f3d' microcode, soh and 2ship use 'f3dex2' and spaghetti and starship use 'f3dex'

src/fast/resource/factory/DisplayListFactory.cpp
```diff
@@ -177,7 +177,8 @@ ResourceFactoryBinaryDisplayListV0::ReadResource(std::shared_ptr<Ship::File> fil
 
         int8_t opcode = (int8_t)(command.words.w0 >> 24);
         bool isExpanded = opcode == G_SETTIMG_OTR_HASH || opcode == G_DL_OTR_HASH || opcode == G_VTX_OTR_HASH ||
-                          opcode == G_BRANCH_Z_OTR || opcode == G_MARKER || opcode == G_MTX_OTR;
+                          opcode == G_BRANCH_Z_OTR || opcode == G_MARKER || opcode == G_MTX_OTR ||
+                          opcode == G_MOVEMEM_OTR;
```
briaguya0, line 181: to expand on my comments, this change doesn't only impact ports using `GBI_FLOATS`

it's not clear to me how it's specific to f3d (as opposed to f3dex or f3dex2)

src/fast/interpreter.cpp
```diff
@@ -2825,7 +2825,7 @@ bool gfx_mtx_otr_handler_custom_f3d(F3DGfx** cmd0) {
         (const int32_t*)Ship::Context::GetInstance()->GetResourceManager()->GetResourceRawPointer(hash);
     if (mtx != nullptr) {
         cmd--;
-        gfx->GfxSpMatrix(C0(16, 8), (const int32_t*)gfx->SegAddr(cmd->words.w1));
+        gfx->GfxSpMatrix(C0(16, 8), mtx);
```
briaguya0, line 2828: to expand on my comments, this change doesn't only impact ports using GBI_FLOATS, but it's clear that it's only for `f3d` as right above this we have `gfx_mtx_otr_handler_custom_f3dex2`

briaguya0 (commented):
> Sure, but no, actually sm64 is the only port that uses the 'f3d' microcode, soh and 2ship use 'f3dex2' and spaghetti and starship use 'f3dex'

gotcha, it wasn't fully clear to me that you didn't just mean "fast3d" in general. `f3d` specifically (as opposed to `ex` or `ex2`) makes sense.

it's clear to me how the change in `interpreter.cpp` is `f3d` specific, but the `DisplayListFactory.cpp` change isn't clear.

src/fast/resource/factory/DisplayListFactory.cpp
KiritoDv: G_MOVEMEM_OTR Is a new opcode that i introduced on torch used by only one port rn

briaguya0 (approved):

PR #975, opened 2026-01-04
Initialize WheelHandler::mBufferedCoords to fix undefined behavior

koplas:
The mBufferedCoords member was never initialized, causing UpdateAxisBuffer to read uninitialized memory when checking if the buffer value is non-zero.

Related: https://github.com/HarbourMasters/Shipwright/issues/5976

### DIFF
--- a/src/ship/controller/controldevice/controller/mapping/mouse/WheelHandler.cpp
+++ b/src/ship/controller/controldevice/controller/mapping/mouse/WheelHandler.cpp
@@ -5,6 +5,7 @@
 namespace Ship {
 WheelHandler::WheelHandler() {
     mDirections = { LUS_WHEEL_NONE, LUS_WHEEL_NONE };
+    mBufferedCoords = { 0.0f, 0.0f };
 }
 
 WheelHandler::~WheelHandler() {


src/ship/controller/controldevice/controller/mapping/mouse/WheelHandler.cpp
```diff
@@ -5,6 +5,7 @@
 namespace Ship {
 WheelHandler::WheelHandler() {
     mDirections = { LUS_WHEEL_NONE, LUS_WHEEL_NONE };
+    mBufferedCoords = { 0.0f, 0.0f };
```
briaguya0, line 8: for completeness sake should we also initialize `mCoords` here?

koplas: Valgrind only flagged `mBufferedCoords` for uninitialized reads, not `mCoords`. While `mCoords` is always initialized before use in the current code, explicitly initializing it would make the code more resilient to future changes. However, I'm keeping this fix minimal and only addressing current issue.

briaguya0 (approved):

PR #977, opened 2026-01-09
Fix memory leaks in Gui and ConsoleWindow

koplas:
- Store ImGui ini/log paths as std::string members instead of raw char* allocations, eliminating manual memory management
- Free mInputBuffer and mFilterBuffer in ConsoleWindow destructor

### DIFF
--- a/include/ship/window/gui/Gui.h
+++ b/include/ship/window/gui/Gui.h
@@ -136,6 +136,8 @@ class Gui {
   private:
     GuiWindowInitData mImpl;
     bool mNeedsConsoleVariableSave;
+    std::string mImGuiIniPath;
+    std::string mImGuiLogPath;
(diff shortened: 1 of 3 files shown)


briaguya0 (approved):
this looks good code-wise. has it been tested against a port? just want to make sure there isn't something i'm missing.

briaguya0:
clang-format should fix itself if you rebase this on or merge in latest main

koplas:
> this looks good code-wise. has it been tested against a port? just want to make sure there isn't something i'm missing.

I found this memory leak while running Valgrind on https://github.com/HarbourMasters/Shipwright. I did not test this patch on other ports.

koplas:
> clang-format should fix itself if you rebase this on or merge in latest main

Is now rebased.

PR #978, opened 2026-01-10
Add option for empty paths list when initializing resource manager.

Malkierian:
This is desired because my ImGui Extraction Flow is setup to handle the port archive not existing on its own, which makes the system popup, and even the iOS exit, redundant, so I don't want that part to process if the port archive happens to not exist during tests.

Defaults to false to maintain current functionality.


Malkierian:
I don't get why tidy failed. I didn't even touch gbi.h

PR #982, opened 2026-01-12
Fix OpenGL api violations and reduce state switching

koplas:
This fixes some issues that I noticed when using the NSight graphics frame debugger. This can slightly improve performance depending on the GPU driver.

### DIFF
--- a/src/fast/backends/gfx_opengl.cpp
+++ b/src/fast/backends/gfx_opengl.cpp
@@ -47,9 +47,11 @@ static void VertexArraySetAttribs(ShaderProgram* prg) {
     size_t pos = 0;
 
     for (int i = 0; i < prg->numAttribs; i++) {
-        glEnableVertexAttribArray(prg->attribLocations[i]);
-        glVertexAttribPointer(prg->attribLocations[i], prg->attribSizes[i], GL_FLOAT, GL_FALSE,
-                              numFloats * sizeof(float), (void*)(pos * sizeof(float)));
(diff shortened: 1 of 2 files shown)


src/fast/backends/gfx_opengl.cpp
```diff
@@ -73,18 +75,24 @@ void GfxRenderingAPIOGL::SetPerDrawUniforms() {
 }
 
 void GfxRenderingAPIOGL::UnloadShader(ShaderProgram* old_prg) {
-    if (old_prg != nullptr) {
+    if (old_prg != nullptr && old_prg == mLastLoadedShader) {
```
briaguya0, line 78: curious about this logic. it seems if:
* `LoadShader` is called twice
  * first on `ShaderProgram` "A" (`mLastLoadedShader` is set to "A")
  * then on `ShaderProgram` "B" (`mLastLoadedShader` is set to "B")
* `UnloadShader` is then called on `ShaderProgram` "A". `mLastLoadedShader` is "B" at this point, so the conditional fails and this block doesn't run.

is that intentional?

koplas: The interpreter always calls `UnloadShader` before `LoadShader`. This is more like a safety check and avoids potential redundant OpenGL calls. I currently don't know what happens, if `LoadShader` is called twice. This is more a question of API contract and taste. If this check is confusing, I am open to remove it.

PR #983, opened 2026-01-14
Fixed meta files failing to load

KiritoDv:
This PR fixes the bug that caused the removal of the .meta when calling IndexFile, which ends telling the resource manager to never find it


briaguya0 (changes requested):
see https://github.com/Kenix3/libultraship/issues/984 for details

PR #986, opened 2026-01-18
Sound Matrix Decoder

tortugaveloz:
This PR adds support for Sound Matrix Decoder, which can be used in supported games. The way it works is by passively converting the stereo output in compatible games to 5.1 output, by using a passive sound matrix decoder, included in this PR. It should work with any 5.1 capable system.

It also allows to transition from stereo to 5.1 dynamically, not requiring a restart of the game. This means this PR does not require adding any new options to the F1/Esc menu of the games, but it rather works using the built-in Surround option the game ships. The audio APIs (SDL / WASAPI) are automatically switched upon request of Surround to 6-channel output.

PRs that depend on this to be merged:

* Ocarina of Time Surround support https://github.com/HarbourMasters/Shipwright/pull/6171
* Majora's Mask Surround support https://github.com/HarbourMasters/2ship2harkinian/pull/1470.


include/ship/audio/DolbyProLogicII.h
```diff
@@ -0,0 +1,119 @@
+#pragma once
+
+#include <cstdint>
+#include <cmath>
+#include <algorithm>
+
+#ifndef M_PI
+#define M_PI 3.14159265358979323846
```
serprex: when does `#include <cmath>` lack this standard define?

tortugaveloz: Thanks, you're right of course. Done.

tortugaveloz: It looks like Windows and ARM don't have M_PI as part of cmath. Reverting to include this change.

serprex: https://stackoverflow.com/questions/6563810/m-pi-works-with-math-h-but-not-with-cmath-in-visual-studio

ugh

https://cplusplus.com/forum/beginner/105176 looks like M_PI is somehow not standard? TIL. https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/math.h.html just POSIX things

include/libultraship/bridge/audiobridge.h
```diff
@@ -14,6 +14,10 @@ AudioChannelsSetting GetAudioChannels();
 int32_t GetNumAudioChannels();
 void AudioPlayerPlayFrame(const uint8_t* buf, size_t len);
 
+// Set audio channels configuration at runtime (stereo or 5.1 surround)
+// This will reinitialize the audio backend without requiring a game restart
+void SetAudioChannelsSetting(AudioChannelsSetting channels);
```
briaguya0: nit: `SetAudioChannelsSetting` feels like an odd name, we just have `GetAudioChannels` for the getter, maybe just `SetAudioChannels`?

include/ship/audio/AudioPlayer.h
```diff
@@ -22,7 +25,11 @@ class AudioPlayer {
 
     bool Init();
     virtual int32_t Buffered() = 0;
-    virtual void Play(const uint8_t* buf, size_t len) = 0;
+
+    // Play stereo audio - will decode to surround if surround mode is enabled
+    // buf: interleaved stereo samples (L, R, L, R, ...)
+    // len: length in bytes
+    void Play(const uint8_t* buf, size_t len);
```
briaguya0, line 31: my main concern here is that this couples using the decoder and surround mode. ideally we'd be able to
* have 2 channel stereo
* have 6 channels raw
* have 6 channels (using the matrix decoder)

include/ship/audio/AudioPlayer.h
briaguya0: i know this isn't part of your changes, but the name `AudioSurround` feels quite odd here.

briaguya0 (commented):
overall this seems to be heading in a good direction. i have a few concerns about coupling the decoding to the channels. i'm not sure what the best way to decouple that would be, maybe editing `AudioChannelsSetting` to have `audioStereo`, `audioMatrix51`, `audioRaw51` or something like that to differentiate.

looking forward to hearing your thoughts!

include/libultraship/bridge/audiobridge.h
tortugaveloz: Done.

include/ship/audio/AudioPlayer.h
tortugaveloz: I agree, in fact this will be beneficial for e.g. Starfox 64 support. Done.

include/ship/audio/AudioPlayer.h
tortugaveloz: Changed to ChannelSetting.

tortugaveloz:
Hi @briaguya0, went through your comments. Please let me know if there are any further comments, thanks!

src/ship/audio/AudioPlayer.cpp
```diff
@@ -43,7 +63,60 @@ void AudioPlayer::SetDesiredBuffered(int32_t size) {
...
+
+        // Decode stereo to surround using sound matrix decoder
+        mSoundMatrixDecoder->Process(stereoIn, mSurroundBuffer.data(), numStereoSamples);
+
+        // Play the surround audio
+        DoPlay(reinterpret_cast<const uint8_t*>(mSurroundBuffer.data()), numStereoSamples * 6 * sizeof(int16_t));
+    } else {
+        // Stereo or Raw 5.1 passthrough
+        DoPlay(buf, len);
+    }
```
briaguya0, line 95: i think it'd be good to add some error logging in here and possibly rework this to use early returns, something like


```suggestion
void AudioPlayer::Play(const uint8_t* buf, size_t len) {
    if (mAudioSettings.ChannelSetting != AudioChannelsSetting::audioMatrix51) {
        // Stereo or Raw 5.1 passthrough
        DoPlay(buf, len);
        return;
    }
    
    if (!mSoundMatrixDecoder) {
        // todo: log error about not having the decoder
    }
    
    // Input is stereo, decode to surround using matrix decoder
    const int16_t* stereoIn = reinterpret_cast<const int16_t*>(buf);
    int numStereoSamples = len / (2 * sizeof(int16_t)); // Number of stereo sample pairs

    // Resize surround buffer if needed
    size_t surroundSamplesNeeded = numStereoSamples * 6;
    if (mSurroundBuffer.size() < surroundSamplesNeeded) {
        mSurroundBuffer.resize(surroundSamplesNeeded);
    }

    // Decode stereo to surround using sound matrix decoder
    mSoundMatrixDecoder->Process(stereoIn, mSurroundBuffer.data(), numStereoSamples);

    // Play the surround audio
    DoPlay(reinterpret_cast<const uint8_t*>(mSurroundBuffer.data()), numStereoSamples * 6 * sizeof(int16_t));
```

another possible option would be to move the matrix stuff out into a new method, and then this could be something like

```cpp
void AudioPlayer::Play(const uint8_t* buf, size_t len) {
    if (mAudioSettings.ChannelSetting == AudioChannelsSetting::audioMatrix51) {
        DoPlayMatrix(buf, len);
        return;
    }

    DoPlay(buf, len);
}
```

include/ship/audio/AudioPlayer.h
```diff
@@ -22,7 +25,11 @@ class AudioPlayer {
 
     bool Init();
     virtual int32_t Buffered() = 0;
-    virtual void Play(const uint8_t* buf, size_t len) = 0;
+
+    // Play stereo audio - will decode to surround if surround mode is enabled
```
briaguya0: this isn't just stereo right? if we're using 5.1 raw then it'll go into `DoPlay` with the raw 6 channels

src/ship/config/Config.cpp
```diff
@@ -264,8 +264,10 @@ AudioChannelsSetting Config::GetCurrentAudioChannelsSetting() {
     int32_t surround =
```
briaguya0: nit: worth a rename

include/ship/audio/AudioPlayer.h
tortugaveloz: Done.

src/ship/config/Config.cpp
tortugaveloz: Done.

tortugaveloz:
Went through the additional comments @briaguya0

include/ship/audio/AudioPlayer.h
```diff
@@ -40,14 +47,35 @@ class AudioPlayer {
...
+    virtual void DoClose() = 0;
+
+    // Internal play method - receives audio in the output format (stereo or surround)
+    virtual void DoPlay(const uint8_t* buf, size_t len) = 0;
+
+    // Sound matrix decoder for surround mode
+    std::unique_ptr<SoundMatrixDecoder> mSoundMatrixDecoder;
+
+    // Surround output buffer
+    std::vector<int16_t> mSurroundBuffer;
```
briaguya0: high level question: would it be possible for this to live in the decoder?

include/ship/audio/AudioPlayer.h
```diff
@@ -40,14 +47,35 @@ class AudioPlayer {
...
+
+    AudioSettings mAudioSettings;
+
+    // Whether matrix decoder needs priming (to fill delay buffers)
+    bool mNeedsPriming = false;
```
briaguya0: high level question: would it be possible for this to live in the decoder?

include/ship/audio/AudioPlayer.h
```diff
@@ -40,14 +47,35 @@ class AudioPlayer {
...
+    // Internal play method - receives audio in the output format (stereo or surround)
+    virtual void DoPlay(const uint8_t* buf, size_t len) = 0;
+
+    // Sound matrix decoder for surround mode
+    std::unique_ptr<SoundMatrixDecoder> mSoundMatrixDecoder;
+
+    // Surround output buffer
+    std::vector<int16_t> mSurroundBuffer;
+
+    AudioSettings mAudioSettings;
```
briaguya0, line 70: does this need to move to protected from private? i don't see it being used outside of `AudioPlayer.cpp`

include/ship/audio/SoundMatrixDecoder.h
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Passive matrix decoder for stereo to 5.1 surround upmixing.
+ * Implements standard audio matrix decoding techniques using:
+ * - Linkwitz-Riley crossover filters for frequency band separation
+ * - All-pass filters for phase manipulation
+ * - Delay lines for surround channel timing
+ */
+class SoundMatrixDecoder {
+  public:
+    explicit SoundMatrixDecoder(int32_t sampleRate = 32000);
```
briaguya0: curious as to why the `explicit` keyword is needed here. do you foresee issues with people passing other types of int for sample rate or something?

include/ship/audio/SoundMatrixDecoder.h
```diff
@@ -0,0 +1,112 @@
...
+     * @param surroundOutput Interleaved 5.1 samples [FL, FR, C, LFE, SL, SR, ...]
+     * @param samplePairs Number of stereo sample pairs to process
+     */
+    void Process(const int16_t* stereoInput, int16_t* surroundOutput, int samplePairs);
+
+    void Reset();
+    void SetSampleRate(int32_t sampleRate);
+    int32_t GetSampleRate() const {
+        return mSampleRate;
+    }
```
briaguya0: more of a high-level question, why does `SoundMatrixDecoder` need to hold on to the sample rate? can we not just grab it from `AudioPlayer` (maybe add a public getter in there for sample rate)

include/ship/audio/WasapiAudioPlayer.h
```diff
@@ -15,10 +15,13 @@ class WasapiAudioPlayer : public AudioPlayer, public IMMNotificationClient {
     WasapiAudioPlayer(AudioSettings settings) : AudioPlayer(settings) {
     }
 
-    int Buffered();
-    void Play(const uint8_t* buf, size_t len);
+    int Buffered() override;
```
briaguya0, line 18: good catch on the missing `override` here. i know it works anyways without it, but this definitely makes it more clear/explicit.

include/ship/audio/NullAudioPlayer.h
```diff
@@ -8,10 +8,11 @@ class NullAudioPlayer final : public AudioPlayer {
     }
     ~NullAudioPlayer();
 
-    int Buffered();
-    void Play(const uint8_t* buf, size_t len);
+    int Buffered() override;
```
briaguya0, line 11: good catch on the missing `override` here. i know it works anyways without it, but this definitely makes it more clear/explicit.

include/ship/audio/SDLAudioPlayer.h
```diff
@@ -9,14 +9,15 @@ class SDLAudioPlayer final : public AudioPlayer {
     }
     ~SDLAudioPlayer();
 
-    int Buffered();
-    void Play(const uint8_t* buf, size_t len);
+    int Buffered() override;
```
briaguya0, line 12: good catch on the missing `override` here. i know it works anyways without it, but this definitely makes it more clear/explicit.

src/libultraship/bridge/audiobridge.cpp
briaguya0, line 51: <details>
<summary><del>just for completeness sake it might be good to have this be</del></summary>

```cpp
int32_t GetNumAudioChannels() {
    switch (GetAudioChannels()) {
        case audioMatrix51:
        case audioRaw51:
            return 6;
        case audioStereo:
        default:
            return 2;
    }
}
```
</details>

now that `AudioPlayer` has `GetNumOutputChannels` this should probably just call that

src/ship/audio/AudioPlayer.cpp
```diff
@@ -1,12 +1,32 @@
...
+        case AudioChannelsSetting::audioStereo:
+            return "Stereo";
+        case AudioChannelsSetting::audioMatrix51:
+            return "5.1 Matrix";
+        case AudioChannelsSetting::audioRaw51:
+            return "5.1 Raw";
+        default:
+            return "Unknown";
+    }
+}
```
briaguya0: i don't think this should be a global thing declared in this cpp file.

if the definition is going to stay here then it should be a private method on `AudioPlayer`.

if we want to keep it out of a class it might be ok to move it to `AudioChannelsSetting.h`.

i don't *love* either of those options, but i prefer them both to the current state of it.

src/ship/audio/AudioPlayer.cpp
```diff
@@ -1,12 +1,32 @@
 #include "ship/audio/AudioPlayer.h"
 #include "spdlog/spdlog.h"
+#include <algorithm>
```
briaguya0: i don't see where this is needed in this file.

src/ship/audio/AudioPlayer.cpp
```diff
@@ -43,7 +63,66 @@ void AudioPlayer::SetDesiredBuffered(int32_t size) {
...
+
+    // Update channel setting
+    mAudioSettings.ChannelSetting = channels;
+
+    // Setup or teardown sound matrix decoder
+    if (channels == AudioChannelsSetting::audioMatrix51 && !mSoundMatrixDecoder) {
+        mSoundMatrixDecoder = std::make_unique<SoundMatrixDecoder>(mAudioSettings.SampleRate);
+    } else if (channels != AudioChannelsSetting::audioMatrix51 && mSoundMatrixDecoder) {
+        mSoundMatrixDecoder.reset();
+    }
```
briaguya0: nit for readability ( know nesting ifs generally feels bad but this was kind of hard to parse for me as-is)

```suggestion
    if (channels == AudioChannelsSetting::audioMatrix51) {
        if (!mSoundMatrixDecoder) {
            mSoundMatrixDecoder = std::make_unique<SoundMatrixDecoder>(mAudioSettings.SampleRate);
        }
        
        mSoundMatrixDecoder.reset();
    }
```

if calling `reset()` right after making the decoder is problematic then that could go into an `else`

src/ship/audio/AudioPlayer.cpp
```diff
@@ -43,7 +63,66 @@ void AudioPlayer::SetDesiredBuffered(int32_t size) {
...
+
+int32_t AudioPlayer::GetNumOutputChannels() const {
+    switch (mAudioSettings.ChannelSetting) {
+        case AudioChannelsSetting::audioMatrix51:
+        case AudioChannelsSetting::audioRaw51:
+            return 6;
+        default:
+            return 2;
+    }
+}
```
briaguya0, line 88: for completeness sake

```suggestion
int32_t AudioPlayer::GetNumOutputChannels() const {
    switch (mAudioSettings.ChannelSetting) {
        case AudioChannelsSetting::audioMatrix51:
        case AudioChannelsSetting::audioRaw51:
            return 6;
        case AudioChannelsSetting::audioStereo:
        default:
            return 2;
    }
}
```

src/ship/audio/AudioPlayer.cpp
```diff
@@ -43,7 +63,66 @@ void AudioPlayer::SetDesiredBuffered(int32_t size) {
...
+    int numStereoSamples = len / (2 * sizeof(int16_t)); // Number of stereo sample pairs
+
+    // Resize surround buffer if needed
+    size_t surroundSamplesNeeded = numStereoSamples * 6;
+    if (mSurroundBuffer.size() < surroundSamplesNeeded) {
+        mSurroundBuffer.resize(surroundSamplesNeeded);
+    }
+
+    // Decode stereo to surround using sound matrix decoder
+    mSoundMatrixDecoder->Process(stereoIn, mSurroundBuffer.data(), numStereoSamples);
```
briaguya0: same high-level question as https://github.com/Kenix3/libultraship/pull/986/changes#r2723318546, would it be possible for this entire block to live in the decoder?

src/ship/audio/SDLAudioPlayer.cpp
```diff
@@ -5,27 +5,46 @@ namespace Ship {
...
+        SDL_ClearQueuedAudio(mDevice);
+        SDL_CloseAudioDevice(mDevice);
+        mDevice = 0;
+    }
+}
+
 bool SDLAudioPlayer::DoInit() {
     if (SDL_Init(SDL_INIT_AUDIO) != 0) {
-        SPDLOG_ERROR("SDL init error: %s\n", SDL_GetError());
+        SPDLOG_ERROR("SDL init error: {}", SDL_GetError());
```
briaguya0: were we getting double newlines here before?

src/ship/audio/WasapiAudioPlayer.cpp
```diff
@@ -28,31 +28,31 @@ bool WasapiAudioPlayer::SetupStream() {
         ThrowIfFailed(mDeviceEnumerator->GetDefaultAudioEndpoint(eRender, eConsole, &mDevice));
         ThrowIfFailed(mDevice->Activate(IID_IAudioClient, CLSCTX_ALL, nullptr, IID_PPV_ARGS_Helper(&mClient)));
 
-        auto audioSurround = this->GetAudioChannels();
-        if (audioSurround == AudioChannelsSetting::audioStereo) {
-            mNumChannels = 2;
+        // Use GetNumOutputChannels() to determine stereo vs surround
+        mNumChannels = this->GetNumOutputChannels();
```
briaguya0, line 32: possible future cleanup (not part of this PR, just making a note to make an issue to dig into this), it feels like there's a bit of "juggling" when managing/storing the number of channels in so many different places. moving to a more solid "single source of truth" would be good

src/ship/audio/WasapiAudioPlayer.cpp
```diff
@@ -28,31 +28,31 @@ bool WasapiAudioPlayer::SetupStream() {
...
-            desired.nAvgBytesPerSec = desired.nSamplesPerSec * desired.nBlockAlign; // 2 bytes per sample (16-bit audio)
+            desired.nAvgBytesPerSec = desired.nSamplesPerSec * desired.nBlockAlign;
             desired.cbSize = 0;
 
             ThrowIfFailed(mClient->Initialize(
                 AUDCLNT_SHAREMODE_SHARED, AUDCLNT_STREAMFLAGS_AUTOCONVERTPCM | AUDCLNT_STREAMFLAGS_SRC_DEFAULT_QUALITY,
                 2000000, 0, &desired, nullptr));
-        } else if (audioSurround == AudioChannelsSetting::audioSurround51) {
-            mNumChannels = 6;
+        } else {
```
briaguya0: the previous logic was explicit here, i think at least having

```suggestion
        } else if (mNumChannels == 6) {
```

would be good.

src/ship/audio/WasapiAudioPlayer.cpp
```diff
@@ -75,15 +75,27 @@ bool WasapiAudioPlayer::SetupStream() {
 
 bool WasapiAudioPlayer::DoInit() {
     try {
-        ThrowIfFailed(
-            CoCreateInstance(CLSID_MMDeviceEnumerator, nullptr, CLSCTX_ALL, IID_PPV_ARGS(&mDeviceEnumerator)));
+        if (!mDeviceEnumerator) {
+            ThrowIfFailed(
+                CoCreateInstance(CLSID_MMDeviceEnumerator, nullptr, CLSCTX_ALL, IID_PPV_ARGS(&mDeviceEnumerator)));
+            ThrowIfFailed(mDeviceEnumerator->RegisterEndpointNotificationCallback(this));
+        }
```
briaguya0, line 82: just want to make sure i'm understanding this change, this is needed for the re-init without restart?

include/ship/audio/AudioPlayer.h
```diff
@@ -40,14 +47,35 @@ class AudioPlayer {
...
+    bool SetAudioChannels(AudioChannelsSetting channels);
+
+    // Get the number of output channels (2 for stereo, 6 for surround)
+    int32_t GetNumOutputChannels() const;
 
   protected:
     virtual bool DoInit() = 0;
 
+    // Reinitialize the audio device with current settings (called after channel change)
+    virtual void DoClose() = 0;
```
briaguya0: this feels incorrect. `SetAudioChannels` calls `DoClose`, does some stuff, then calls `DoInit`. i assume the `DoInit` call does the "reinitialize" part, and `DoClose` just closes the device.

briaguya0 (commented):
thank you so much for your patience throughout this review process.

i finally got a chance to give this a full in-depth review

overall this is looking really great, i left quite a few comments this time but they should all be pretty small

include/ship/audio/AudioPlayer.h
tortugaveloz: Yes, done.

include/ship/audio/AudioPlayer.h
tortugaveloz: We don't need this, it's a remnant from a previous version where I was debugging an issue that is now solved. Removed.

include/ship/audio/AudioPlayer.h
tortugaveloz: Fair enough, moved these two to private.

include/ship/audio/SoundMatrixDecoder.h
tortugaveloz: You're right, we don't need to set it explicit. Changed it.

include/ship/audio/SoundMatrixDecoder.h
tortugaveloz: Changed it to holding calculated values instead.

src/libultraship/bridge/audiobridge.cpp
tortugaveloz: Done.

src/ship/audio/AudioPlayer.cpp
tortugaveloz: Removed.

src/ship/audio/AudioPlayer.cpp
tortugaveloz: Done.

src/ship/audio/AudioPlayer.cpp
tortugaveloz: I went with the second option, declaring it solely in `AudioChannelsSetting.h`.

src/ship/audio/SDLAudioPlayer.cpp
briaguya0: oh, looks like it was worse than that https://github.com/Kenix3/libultraship/pull/963

src/ship/audio/AudioPlayer.cpp
tortugaveloz: Nesting ifs is common practice, I agree with your approach. Fixed it.

src/ship/audio/SDLAudioPlayer.cpp
tortugaveloz: Reverted this.

src/ship/audio/WasapiAudioPlayer.cpp
tortugaveloz: Done.

src/ship/audio/WasapiAudioPlayer.cpp
tortugaveloz: Yes, the if statement is definitely needed, and I think the `RegisterEndpointNotificationCallback` as well.

include/ship/audio/AudioPlayer.h
tortugaveloz: Yes, the comment was indeed wrong. I have updated it.

src/ship/audio/AudioPlayer.cpp
tortugaveloz: Done.

src/ship/audio/AudioPlayer.cpp
```diff
@@ -43,7 +49,67 @@ void AudioPlayer::SetDesiredBuffered(int32_t size) {
...
+    }
+
+    if (!mSoundMatrixDecoder) {
+        SPDLOG_ERROR("AudioPlayer: Matrix 5.1 mode enabled but SoundMatrixDecoder is not initialized");
+        return;
+    }
+
+    // Input is stereo, decode to surround using matrix decoder
+    const int16_t* stereoIn = reinterpret_cast<const int16_t*>(buf);
+    int numStereoSamples = len / (2 * sizeof(int16_t)); // Number of stereo sample pairs
```
briaguya0: could these go into `SoundMatrixDecoder::Process` too?

src/ship/audio/AudioPlayer.cpp
```diff
@@ -43,7 +49,67 @@ void AudioPlayer::SetDesiredBuffered(int32_t size) {
...
+
+    // Decode stereo to surround using sound matrix decoder
+    const int16_t* surroundOut = mSoundMatrixDecoder->Process(stereoIn, numStereoSamples);
+    if (!surroundOut) {
+        SPDLOG_ERROR("AudioPlayer: SoundMatrixDecoder::Process failed - decoder not initialized");
```
briaguya0: the wording here feels a little odd, earlier in this method we already use "not initialized" when checking if the decoder is not null.

i think just moving the error logging into `SoundMatrixDecoder::Process` where the `mReady` check is might be clearer.

include/ship/audio/AudioPlayer.h
```diff
@@ -40,14 +46,29 @@ class AudioPlayer {
...
+    bool SetAudioChannels(AudioChannelsSetting channels);
+
+    // Get the number of output channels (2 for stereo, 6 for surround)
+    int32_t GetNumOutputChannels() const;
 
   protected:
+    // Initialize the audio device.
     virtual bool DoInit() = 0;
 
+    // Close the current audio device and reset the audio player.
```
briaguya0: this still feels off. specifically the

>  and reset the audio player.

part

src/ship/audio/AudioPlayer.cpp
```diff
@@ -43,7 +49,68 @@ void AudioPlayer::SetDesiredBuffered(int32_t size) {
...
+
+    // Setup or teardown sound matrix decoder
+    if (channels == AudioChannelsSetting::audioMatrix51) {
+        if (!mSoundMatrixDecoder) {
+            mSoundMatrixDecoder = std::make_unique<SoundMatrixDecoder>(mAudioSettings.SampleRate);
+        }
+    } else {
+        // When switching away from matrix mode, release the decoder
+        mSoundMatrixDecoder.reset();
+    }
```
briaguya0, line 74: my mistake on the earlier suggestion. i was getting smart pointer `.reset()` and `SoundMatrixDecoder::ResetState` mixed up in my mind.

briaguya0 (commented):
this looks super close to ready to land, could small extra comments

include/ship/audio/AudioPlayer.h
tortugaveloz: Fixed.

src/ship/audio/AudioPlayer.cpp
tortugaveloz: No, since we need `numStereoSamples` to invoke `DoPlay` afterwards.

src/ship/audio/AudioPlayer.cpp
tortugaveloz: That's a remnant from when we had the `Init()` function in the surround matrix decoder. I have removed that check and any reference to that function now.

src/ship/audio/AudioPlayer.cpp
briaguya0: what if instead of having `SoundMatrixDecoder::Process` return just the pointer to the data we had it return a struct? something like

```cpp
struct DecodedAudio {
    const uint8_t* data;
    size_t len;
};
```

tortugaveloz: Used a tuple for this, it's a simple enough case that I don't think we need the struct.

briaguya0 (approved):
beautiful!

thanks for working through the review process with me! i'm super happy with how this turned out!

SirYodaJedi:
> It also allows to transition from stereo to 5.1 dynamically, not requiring a restart of the game. This means this PR does not require adding any new options to the F1/Esc menu of the games, but it rather works using the built-in Surround option the game ships. The audio APIs (SDL / WASAPI) are automatically switched upon request of Surround to 6-channel output.

Does this mean it always outputs 5.1 when the surround option is enabled in-game, instead of the matrixed 2.0 that the original console did? I always select the surround option in-game, even on stereo speakers and headphones, so feel like this might change how the games sound by having the API downmix to stereo instead of the game itself (as well as loosing the ability to use my AV receiver's upmixing options, but that shouldn't be too different from this PR).

PR #987, opened 2026-01-20
L3DEX: Support gSPLine3D and gSPLineW3D (G_LINE3D)

inspectredc:
I mostly just followed the implementation for GfxSpTri1 but removed the texture sections and geometry state sections as those do not apply to the lines. I wasn't sure if only certain color combiner modes can apply to lines or not so left them unchanged compared to tris.

I found that when implementing using the line primitive, I was not able to set line width on backends such as metal so I ended up implementing this as quads with the new vertices calculated in the cpu side. The alternative would have been to write shaders on the line primitives but that seemed more extensive that it needed to be.

I was not sure if I should increase the max_tri_buffer for this as I'm not sure what that value originates from either, but obviously as I am implementing as quads here this halves the amount of lines you can draw when compared to tris which may be relevant.

I also was not sure if an implementation to check if the L3DEX ucode was loaded to disallow tris and allow lines was desirable or not.

To test this out I found replacing the gSP1Triangle call in Collider_DrawPoly in SoH with 3 gSPLineW3D calls was a nice easy way to check the results.


MegaMech (approved):

include/libultraship/libultra/gbi.h
```diff
@@ -1978,7 +1978,7 @@ typedef union Gfx {
     (((flag) == 0)   ? __gsSP1Triangle_w1(v0, v1, v2) \
      : ((flag) == 1) ? __gsSP1Triangle_w1(v1, v2, v0) \
                      : __gsSP1Triangle_w1(v2, v0, v1))
-#define __gsSPLine3D_w1(v0, v1, wd) (_SHIFTL((v0)*2, 16, 8) | _SHIFT((v1)*2, 8, 8) | _SHIFT((wd), 0, 8))
+#define __gsSPLine3D_w1(v0, v1, wd) (_SHIFTL((v0)*2, 16, 8) | _SHIFTL((v1)*2, 8, 8) | _SHIFTL((wd), 0, 8))
```
briaguya0, line 1981: curious about this change. i assume it's just a "this was never implemented/used before so it was never switched from `_SHIFT` to `_SHIFTL`" thing, but it'd be good to get confirmation on that

briaguya0 (approved):
i haven't looked at the interpreter code too closely, but overall this seems reasonable. i left one small comment with a question, i'll be happy to land this once that's answered

Kenix3:
I'm gonna PR to your branch soon. My implementation is a little fleshed out on the ucode side. Specifically it supports tris as lines like the official ucode did.

My implementation covers everything but the flagged vertex.

include/libultraship/libultra/gbi.h
inspectredc: It was the only thing that was using the _SHIFT macro in gbi.h still, which in mbi.h is normally just defined to _SHIFTL for compatibility purposes. I felt it made more sense just changing it here in case the _SHIFT macro had been removed from mbi.h for some OS reasons, but happy to have it either way if it doesn't have any issues related to it

src/fast/interpreter.cpp
```diff
@@ -4019,6 +4307,7 @@ static constexpr UcodeHandler f3dexHandlers = {
     { F3DEX_G_SPNOOP, { "G_SPNOOP", gfx_spnoop_command_handler_f3dex2 } },
     { F3DEX_G_RDPHALF_1, { "mRdpHALF_1", gfx_stubbed_command_handler } },
     { F3DEX_G_QUAD, { "G_QUAD", gfx_quad_handler_f3dex } },
+    { F3DEX_G_LINE3D, { "G_LINE3D", gfx_line3d_handler_f3dex } },
```
MegaMech: I just remebered, for f3dex 0.95 QUADRANGLE and line3D share the same op code.

Just keep that in-mind. And quadrangle is used. But line3D is not for 0.95 games.

Kenix3: I'm pretty sure the ucode for regular f3d doesn't implement the line3d opcode at all. I think it only works on l3d ucode.

PR #990, opened 2026-01-24
Improve interpreter performance

koplas:
- Cache culling constants and other_mode_l to reduce repeated lookups
- Precompute inv_w and texture reciprocals to replace divisions with multiplications
- Avoid double map lookups in gfx_step()
- Use hash-based resource loading to skip string allocation

### DIFF
--- a/src/fast/interpreter.cpp
+++ b/src/fast/interpreter.cpp
@@ -68,6 +68,11 @@ namespace Fast {
 
 static UcodeHandlers ucode_handler_index = ucode_f3dex2;
 
+// Cached culling constants (updated when ucode changes)
+static uint32_t cached_cull_both;
+static uint32_t cached_cull_front;
+static uint32_t cached_cull_back;
+
 const static uint32_t f3dex2AttrHandler[] = {
     F3DEX2_G_MTX_PROJECTION, F3DEX2_G_MTX_LOAD,  F3DEX2_G_MTX_PUSH,  F3DEX_G_MTX_NOPUSH,
     F3DEX2_G_CULL_FRONT,     F3DEX2_G_CULL_BACK, F3DEX2_G_CULL_BOTH,
@@ -92,6 +97,12 @@ static uint32_t get_attr(Attribute attr) {
     return (*ucode_map)[attr];
 }
 
+static void update_cached_cull_attrs() {
+    cached_cull_both = get_attr(CULL_BOTH);
+    cached_cull_front = get_attr(CULL_FRONT);
+    cached_cull_back = get_attr(CULL_BACK);
+}
+
 static std::string GetPathWithoutFileName(char* filePath) {
     size_t len = strlen(filePath);
 
@@ -105,6 +116,8 @@ static std::string GetPathWithoutFileName(char* filePath) {
 }
 
 constexpr size_t MAX_TRI_BUFFER = 256;
+static constexpr float INV_32 = 1.0f / 32.0f;
+static constexpr float INV_255 = 1.0f / 255.0f;
 
 Interpreter::Interpreter() {
     mRsp = new RSP();
@@ -1321,6 +1334,7 @@ void Interpreter::GfxSpVertex(size_t n_vertices, size_t dest_index, const F3DVtx
         d->y = y;
         d->z = z;
         d->w = w;
+        d->inv_w = (fabsf(w) > 0.0001f) ? (1.0f / w) : 0.0f;
 
         if (mRsp->geometry_mode & G_FOG) {
             if (fabsf(w) < 0.001f) {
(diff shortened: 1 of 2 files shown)


src/fast/interpreter.cpp
```diff
@@ -1459,45 +1475,17 @@ void Interpreter::GfxSpTri1(uint8_t vtx1_idx, uint8_t vtx2_idx, uint8_t vtx3_idx
...
-    }
+    // Branchless shader options computation
+    cc_options |= (use_alpha ? SHADER_OPT(ALPHA) : 0) | (use_fog ? SHADER_OPT(FOG) : 0) |
+                  (texture_edge ? SHADER_OPT(TEXTURE_EDGE) : 0) | (use_noise ? SHADER_OPT(NOISE) : 0) |
+                  (use_2cyc ? SHADER_OPT(_2CYC) : 0) | (alpha_threshold ? SHADER_OPT(ALPHA_THRESHOLD) : 0) |
+                  (invisible ? SHADER_OPT(INVISIBLE) : 0) | (use_grayscale ? SHADER_OPT(GRAYSCALE) : 0) |
+                  (mRdp->loaded_texture[0].masked ? SHADER_OPT(TEXEL0_MASK) : 0) |
+                  (mRdp->loaded_texture[1].masked ? SHADER_OPT(TEXEL1_MASK) : 0) |
+                  (mRdp->loaded_texture[0].blended ? SHADER_OPT(TEXEL0_BLEND) : 0) |
+                  (mRdp->loaded_texture[1].blended ? SHADER_OPT(TEXEL1_BLEND) : 0);
```
briaguya0: i don't think this actually makes a difference, i did a small test on godbolt.org (gcc -O2) with

```cpp

int with_ifs(int a, int b) {
    int blarg = 0;
    if (a) {
        blarg |= 1;
    }
    
    if (b) {
        blarg |= 2;
    }

    return blarg;
}

int no_ifs(int a, int b) {
    int blarg = 0;
    blarg |= (a ? 1 : 0) | (b ? 2 : 0);

    return blarg;
}
```

and got

```
"with_ifs(int, int)":
        xor     eax, eax
        test    edi, edi
        setne   al
        mov     edx, eax
        or      edx, 2
        test    esi, esi
        cmovne  eax, edx
        ret
"no_ifs(int, int)":
        xor     eax, eax
        test    esi, esi
        setne   al
        xor     edx, edx
        add     eax, eax
        test    edi, edi
        setne   dl
        or      eax, edx
        ret
```

it's very possible i'm missing something here, but at first glance this appears to just make it harder to read/debug without improving performance

briaguya0:
i don't mean to imply that there aren't meaningful performance improvements in this PR. there are absolutely great changes in here. thanks for looking into this!

src/fast/interpreter.cpp
koplas: I will look into the generated assembly. Maybe some changes can be omitted.

koplas:
It looks like I profiled a debug build by accident. I removed all optimizations that are already done by the compiler. The remaining changes should have a positive impact.

src/fast/interpreter.cpp
koplas: I removed this, as it is already done by the compiler.

src/fast/interpreter.cpp
```diff
@@ -3371,18 +3389,12 @@ bool gfx_set_timg_otr_hash_handler_custom(F3DGfx** cmd0) {
...
-    if (fileName == nullptr) {
-        (*cmd0)++;
-        return false;
-    }
-
+    // Use hash-based LoadResource to avoid string allocation from const char* -> std::string conversion
     std::shared_ptr<Fast::Texture> texture =
-        std::static_pointer_cast<Fast::Texture>(Ship::Context::GetInstance()->GetResourceManager()->LoadResourceProcess(
-            Ship::Context::GetInstance()->GetResourceManager()->GetArchiveManager()->HashToCString(hash)));
+        std::static_pointer_cast<Fast::Texture>(Ship::Context::GetInstance()->GetResourceManager()->LoadResource(hash));
```
briaguya0: `LoadResource` is async (it just queues the resource load on the thread pool), `LoadResourceProcess` is what gets queued, calling it directly bypasses the pool

i'd likely need to dig into the blame on this to understand why `LoadResourceProcess` is being used here instead of `LoadResource`

briaguya0: update, this PR is blocked until we know this change won't break things. `LoadResourceProcess` -> `LoadResource` is an *extremely* scary change, and should not be taken lightly

briaguya0 (changes requested):
i haven't found time to dig into the part of this i want to investigate yet. i'm not comfortable with it landing before resolving that.

src/fast/interpreter.cpp
briaguya0: ok, so i looked into this and i think we can accomplish what you're trying to accomplish with
> `// Use hash-based LoadResource to avoid string allocation from const char* -> std::string conversion`

by (diff against latest main's version of the function: 956f1082122f219d536c20483df89469f4b807dc as of writing)

```diff
diff --git a/src/fast/interpreter.cpp b/src/fast/interpreter.cpp
index 73a29336..7955bbaf 100644
--- a/src/fast/interpreter.cpp
+++ b/src/fast/interpreter.cpp
@@ -3440,7 +3440,7 @@ bool gfx_set_timg_otr_hash_handler_custom(F3DGfx** cmd0) {
     (*cmd0)++;
     uint64_t hash = ((uint64_t)(*cmd0)->words.w0 << 32) + (uint64_t)(*cmd0)->words.w1;
 
-    const char* fileName = Ship::Context::GetInstance()->GetResourceManager()->GetArchiveManager()->HashToCString(hash);
+    const std::string* fileName = Ship::Context::GetInstance()->GetResourceManager()->GetArchiveManager()->HashToString(hash);
     uint32_t texFlags = 0;
     RawTexMetadata rawTexMetadata = {};
 
@@ -3450,8 +3450,7 @@ bool gfx_set_timg_otr_hash_handler_custom(F3DGfx** cmd0) {
     }
 
     std::shared_ptr<Fast::Texture> texture =
-        std::static_pointer_cast<Fast::Texture>(Ship::Context::GetInstance()->GetResourceManager()->LoadResourceProcess(
-            Ship::Context::GetInstance()->GetResourceManager()->GetArchiveManager()->HashToCString(hash)));
+        std::static_pointer_cast<Fast::Texture>(Ship::Context::GetInstance()->GetResourceManager()->LoadResourceProcess(*fileName));
     if (texture != nullptr) {
         texFlags = texture->Flags;
         rawTexMetadata.width = texture->Width;
@@ -3488,7 +3487,7 @@ bool gfx_set_timg_otr_hash_handler_custom(F3DGfx** cmd0) {
 
         if (tex != NULL) {
             Interpreter* gfx = mInstance.lock().get();
-            gfx->GfxDpSetTextureImage(fmt, size, width, fileName, texFlags, rawTexMetadata, tex);
+            gfx->GfxDpSetTextureImage(fmt, size, width, fileName->c_str(), texFlags, rawTexMetadata, tex);
         }
     } else {
         SPDLOG_ERROR("G_SETTIMG_OTR_HASH: Texture is null");
```

koplas: I reverted the change. We should remove the string allocation in another pull request.

PR #993, opened 2026-01-29
Better surround sound mix

tortugaveloz:
This PR improves the surround sound mix according to ITU‑R BS.775 and general good practices about surround matrix sound.

### DIFF
--- a/src/ship/audio/SoundMatrixDecoder.cpp
+++ b/src/ship/audio/SoundMatrixDecoder.cpp
@@ -4,10 +4,10 @@ namespace Ship {
 
 // Standard matrix decoding gains derived from psychoacoustic principles
 namespace Gains {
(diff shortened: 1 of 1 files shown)


briaguya0:
curious as to where the values from before came from

tortugaveloz:
> curious as to where the values from before came from

The original implementation numbers came from https://stackoverflow.com/questions/23713626/correctly-implementing-dolby-pro-logic-ii. Upon more reading, now I think it's better to stick to the current numbers, sorry for the back and forth.

I have also encountered a few open source implementations of DPL decoders, such as the one used in Dolphin (https://foobar.hyv.fi/?view=foo_dsp_fsurround). It is far more powerful, however it is GPL, so as I understand it we'd rather keep the one we have around. I may look into it for improvements. For now, the current is a good enough first implementation.

PR #994, opened 2026-01-30
Add ability to toggle allowing background inputs

aMannus:
This adds some handling for allowing ports to enable or disable background inputs.

Both OpenGL and Metal support the SDL Hint, sadly DX11 does not seem to. Potentially this changes with SDL 3? I couldn't find it by googling. Either way, this handles OpenGL and Metal through the hint, and has different one using the already existing `ControllerBlockGameInput` in LUS for DX11.

I can't test this on a Mac or Linux, but I did test both OpenGL and DX11 on Windows and it worked as expected.

Example of this working: https://github.com/HarbourMasters/Shipwright/pull/6214


src/fast/backends/gfx_dxgi.cpp
```diff
@@ -52,6 +54,8 @@
 #endif
 using QWORD = uint64_t; // For NEXTRAWINPUTBLOCK
 
+#define ALLOW_BACKGROUND_INPUTS_BLOCK_ID 95237930
```
briaguya0, line 57: a comment providing context on why this is `95237930` would be nice

aMannus: I'll be honest, the only other instance of something using these functions was the input editor, and it used this:
`#define INPUT_EDITOR_WINDOW_GAME_INPUT_BLOCK_ID 95237929`

So I just uhh.. added one.

briaguya0: image

i'll do some digging into old branches to see if past me left any info about the one in input editor

briaguya0: and i also left no info... it first showed up in https://github.com/briaguya0/libultraship/commit/4a475344f341584ca7673c4b11d52107f01088f7 which just says "back to allowing multiple input blockers"

aMannus: Uh oh 😂

briaguya0: looks like the block id concept was from https://github.com/Kenix3/libultraship/pull/249, and there we just had the input editor window using `-1` which i assume caused issues?

my guess is i picked the number randomly and made it big so it wouldn't conflict with other block ids, but i really have no clue.

i think the best course of action would be to make an issue to dig into it and come up with a better strategy for ensuring we don't accidentally end up with duplicate block ids, but for this PR it'll be fine.

aMannus: Sounds good to me. I'll make an issue when I'm in front of my PC again if you didn't beat me to it by then.

Malkierian:
Linux has been confirmed working with the SoH PR.

Malkierian:
Mac also works.

src/fast/backends/gfx_dxgi.cpp
aMannus: issue made here: https://github.com/Kenix3/libultraship/issues/995

PR #1001, opened 2026-02-04
Fix race condition in WASAPI when changing between 2 and 6 channel configuration.

tortugaveloz:
The WASAPI backend has a race condition when changing the channel configuration in audio, that resulted in a crash when changing between e.g. stereo and surround. Tested in Ghostship and soh.

### DIFF
--- a/include/ship/audio/WasapiAudioPlayer.h
+++ b/include/ship/audio/WasapiAudioPlayer.h
@@ -6,6 +6,7 @@
 #include <wrl/client.h>
 #include <mmdeviceapi.h>
 #include <audioclient.h>
+#include <mutex>
 
 using namespace Microsoft::WRL;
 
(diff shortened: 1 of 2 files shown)


briaguya0 (commented):
couple questions:
* is this exclusively a WASAPI issue?
* are we locking more than we need to? the mutex `lock()`s are currently scoped to entire methods

tortugaveloz:
> couple questions:
> 
>     * is this exclusively a WASAPI issue?

Yes, it is exclusively a WASAPI issue. DoPlay in the SDLAudioPlayer only enqueues, and the playing is handled by the same thread that pauses when changing number of channels.

> 
>     * are we locking more than we need to? the mutex `lock()`s are currently scoped to entire methods

Anything that could result in a race condition is handled by the same mutex. That includes:

* DoClose() - which alters mClient, mInitialized and mRenderClient.
* DoPlay() - which checks and alters mInitialized by invoking SetupStream. The rest of the function relies on mClient being valid, so it also needs the lock.
* Buffered() - same reason as DoPlay.
* OnDefaultDeviceChanged - For this last one, there is a comment in the code specifying that:

        // This callback runs on a separate thread,
        // but it's not important how fast this write takes effect.
        
However, as mInitialized is not atomic, it can happen that one thread writes to it and another thread writes to it at the same time, which is UB.

Kenix3 (approved):

PR #1003, opened 2026-02-07
Add option to use nearest neighbor for framebuffer

coavins:
This branch enhances the interpreter and rendering api to allow specifying the texture filter on the framebuffers. This lets you use "nearest neighbor" when upscaling from N64 resolution to the viewport. When you call `SetUpscaleFilter` on the interpreter, the filter changes at runtime without requiring a restart.

I only made changes to the opengl backend, the others probably do not compile.

Resolves HarbourMasters/Shipwright#6234


coavins:
Screenshot from 2026-02-07 14-41-51
Screenshot from 2026-02-07 14-41-07

Malkierian:
The last time I looked at this, I didn't seem to remember needing the modifications to UpdateFramebufferParameters, but I think I had to restart to get the changes to register. Is that to allow changing without restart?

coavins:
Yes, changing that function is what made it work without restarting.

Being unfamiliar with the project, I wasn't sure exactly what the best design would be here. I tried a couple of different things, not all of them involved changes to UpdateFramebufferParameters.

I chose to settle for these changes because like you said they allow changing the setting without restarting, and it also hopefully doesn't violate any existing design philosophy too much.

Malkierian:
Are you on Linux, btw? Just wondering why the DX side didn't get changed.

coavins:
Sorry, yes I'm on Linux so I just changed what I was able to test.

Malkierian:
Does the way you have it now still work on startup? Just wondering because I think `SetSamplerParameters` also does things with the filtering, and I think that's supposed to be the startup.

coavins:
From the player's perspective it does remember my choice and it seems to use the right filter from startup with no problem. I don't think I looked at SetSamplerParameters, I could have missed something there.

Malkierian:
As long as it works, I don't think we necessarily need to change anything in there. My only question now is, does Update get called once per frame? If so, that by itself would recommend a separate function to update the filtering instead.

Malkierian:
Oh, nevermind, you didn't set it up with a CVar lookup on each frame, so that shouldn't matter as much.

coavins:
Yeah I was concerned about that; the interpreter holds on to the filter and just passes it into the graphics api calls. We update the flag from the gui if the setting is changed, so there shouldn't be any needless lookups.

Malkierian:
Anyway, you should be able to mirror the changes you made to OGL's file in the same functions in DX11 and Metal's files, as both have the desired variables for their upscaling methods listed in `SetSamplerParameters()`. We can use the SoH PR to test them all. I highly doubt this would get in with only one renderer being changed, anyway.

coavins:
Per your suggestion, I tried applying the same changes in the metal and d3d backends. I can't test or even compile these myself, so please be careful.

I looked at `SetSamplerParameters` and you were right, it seems it had what I needed.

coavins:
I don't know about the metal implementation here, it seems like I can't get the sampler parameters back out of the sampler, and I don't have all of the values needed to recreate it in `UpdateFramebufferParameters`.

coavins:
Metal sampler should be fine now.

Malkierian:
Tried this on Windows, OpenGL works fine, DX seems to still not work. If you can't figure it out, I'll try to see if I can, since I'm on Windows.

PR #1004, opened 2026-02-10
fix(fast): resolve 64-bit pointer truncation in SegAddr for ARM64

JeodC:
Refactor SegAddr to prevent 32-bit truncation of native 64-bit pointers, addressing an immediate SIGSEGV (SIG11) crash on arm64 platforms using C++20 and a "roulette" crash on other 64-bit platforms.

Changes:
- Implemented a high-address guard (> 0x0FFFFFFF) to identify and return raw 64-bit pointers immediately.
- Transitioned internal segmentation variables (segNum, offset) to uintptr_t to prevent C++20 compilers from performing aggressive truncation or "optimal" register narrowing.
- Removed legacy bit-stripping (0x...FE) that was corrupting potentially unaligned OTR resource handles.

This fix ensures that native system pointers passed via the display list are preserved in their entirety, while maintaining backward compatibility for legacy N64 segmented addresses.

### DIFF
--- a/include/fast/backends/gfx_opengl.h
+++ b/include/fast/backends/gfx_opengl.h
@@ -89,6 +89,7 @@ class GfxRenderingAPIOGL final : public GfxRenderingAPI {
     void CopyFramebuffer(int fbDstId, int fbSrcId, int srcX0, int srcY0, int srcX1, int srcY1, int dstX0, int dstY0,
                          int dstX1, int dstY1) override;
     void ClearFramebuffer(bool color, bool depth) override;
+    void ClearDepthRegion(int x, int y, int w, int h) override;
     void ReadFramebufferToCPU(int fbId, uint32_t width, uint32_t height, uint16_t* rgba16Buf) override;
     void ResolveMSAAColorBuffer(int fbIdTarger, int fbIdSrc) override;
     std::unordered_map<std::pair<float, float>, uint16_t, hash_pair_ff>
--- a/include/fast/backends/gfx_rendering_api.h
+++ b/include/fast/backends/gfx_rendering_api.h
@@ -61,6 +61,11 @@ class GfxRenderingAPI {
     virtual void CopyFramebuffer(int fbDstId, int fbSrcId, int srcX0, int srcY0, int srcX1, int srcY1, int dstX0,
                                  int dstY0, int dstX1, int dstY1) = 0;
     virtual void ClearFramebuffer(bool color, bool depth) = 0;
+    virtual void ClearDepthRegion(int x, int y, int w, int h) {
+        // Default: full depth clear. Backends that support scissored depth clears
+        // (e.g. OpenGL) should override for a more precise partial clear.
+        ClearFramebuffer(false, true);
+    }
     virtual void ReadFramebufferToCPU(int fbId, uint32_t width, uint32_t height, uint16_t* rgba16Buf) = 0;
     virtual void ResolveMSAAColorBuffer(int fbIdTarger, int fbIdSrc) = 0;
     virtual std::unordered_map<std::pair<float, float>, uint16_t, hash_pair_ff>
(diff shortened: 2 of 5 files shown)


JeodC:
If this pull request is merged then projects relying on LUS that do not use 64-bit memory alignment will break. If you see the linked SpaghettiKart pull request, you will find the accompanying macros `ALIGNED8` and `ALIGNED16` were added to work in tandem with the interpreter.

Kenix3:
This is a pretty big change to the way that segments were originally working. I might not be understanding the thought process. The `w1 & 1` check was there to be able to tell if we were loading a string asset or not iirc.

JeodC:
> This is a pretty big change to the way that segments were originally working. I might not be understanding the thought process. The `w1 & 1` check was there to be able to tell if we were loading a string asset or not iirc.

You’re remembering the original intent correctly: `w1 & 1` was historically used as a sentinel bit to distinguish a segmented address from a raw value (including string / asset handles). That behavior made sense when everything in this path was 32-bit and N64-style.

What changed is not the *concept* of segmentation, but the *nature of the values now flowing through this function*. Today, SegAddr can receive native host pointers and OTR resource handles, which are:
- 64-bit
- Not guaranteed to be aligned
- Not guaranteed to have LSB = 1

Once non-N64 addresses started flowing through the display list, `w1 & 1` stopped being a reliable discriminator. On ARM64 in particular, this leads to misclassification and pointer corruption, which is what caused the SIGSEGVs.

Range discrimination (w1 > 0x0FFFFFFF) is stable because:
- No valid N64 address can exceed that range
- It does not rely on alignment or sentinel bits
- It cleanly separates legacy segmented addresses from native pointers

So the behavior change isn’t about redefining segmentation semantics, but about using a signal that remains valid in a mixed 32-bit / 64-bit environment. Keeping `w1 & 1` as the gate would make native pointer support inherently unreliable.

JeodC:
So in rereading, I feel my above comment was a tad abstract. Since I've been spelunking in a project, I can give a real example.

The bit 0 protocol requires every component in the pipeline (ROM extractors, asset exporters, runtime importers) to coordinate on setting `w1 |= 1` for segmented addresses. If any path misses it, SegAddr returns the raw segmented value (e.g. 0x03000018) as a pointer. If the stripping (& 0xFE) is ever removed or missed, the resolved address is off by 1 byte.

In practice, some games have display lists contain G_DL commands targeting segment 3 (render mode tables). Under the bit 0 protocol, the importer must set `w1 |= 1` on these. But when the interpreter resolves a G_DL target that's 1 byte off, it reads misaligned Gfx commands — garbage opcodes corrupt the execution stack, producing crashes with invalid cmd pointers like 0xFFFFFFFFFF0B7DC1 (note: bit 0 set in the crash address itself).

So, I opt for a different approach.

The N64 address space caps at 256MB (0x0FFFFFFF). Every valid segmented address fits below this. Every native 64-bit host pointer (heap, stack, mmap) is above it. The check `w1 > 0x0FFFFFFF` is a property of the address itself — no flags to set, no flags to strip, no pipeline coordination needed. It works regardless of alignment, platform, or which component produced the address.

Now here's the caveat: any existing `w1 |= 1` in exporters (like Torch's DisplayListFactory for MK64 segment pass-through) should be removed to match. Otherwise the bit passes through to the offset calculation and corrupts the resolved address.

In short, the threshold approach eliminates an entire class of bugs by making segment detection self-contained in SegAddr. The current bit 0 protocol is a distributed contract — every exporter, importer, and handler must participate correctly or addresses silently corrupt. The threshold check requires zero cooperation from the rest of the pipeline: if it came from the N64, it's below 0x0FFFFFFF; if it came from the host, it's above. That's it.

JeodC:
I am debating whether to close this, as it's a fundamental change for LUS that in the end is helpful, but requires all ports remove the opt-in they are currently using.

Kenix3:
I'm still debating this myself. I've always been unhappy with the hacky nature of the test. I'm generally fine with making ports update to LUS features. It's one of the reasons I want to finish 2.x.

JeodC:
Oh whoops this was on `develop`, I'll move it to another branch and reopen if you want it reopened.

MegaMech:
Just want to add my two cents. This definitely sounds like a reasonable solution that fixes errors of the 'unspecified' sort. *I think* SpaghettiKart has completely moved away from segmented data?

PR #1010, opened 2026-02-16
Fixed point filtering on metal

KiritoDv:
This should be the last mac related bug, thank you again @crowell

### DIFF
--- a/src/fast/backends/gfx_metal.cpp
+++ b/src/fast/backends/gfx_metal.cpp
@@ -353,6 +353,7 @@ void GfxRenderingAPIMetal::UploadTexture(const uint8_t* rgba32_buf, uint32_t wid
 void GfxRenderingAPIMetal::SetSamplerParameters(int tile, bool linear_filter, uint32_t cms, uint32_t cmt) {
     TextureDataMetal* texture_data = &mTextures[mCurrentTextureIds[tile]];
     texture_data->linear_filtering = linear_filter;
+    texture_data->filtering = !linear_filter ? FILTER_LINEAR : FILTER_THREE_POINT;
 
     // This function is called twice per texture, the first one only to set default values.
     // Maybe that could be skipped? Anyway, make sure to release the first default sampler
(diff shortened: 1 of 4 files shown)


src/fast/backends/gfx_metal.cpp
```diff
@@ -353,6 +353,7 @@ void GfxRenderingAPIMetal::UploadTexture(const uint8_t* rgba32_buf, uint32_t wid
 void GfxRenderingAPIMetal::SetSamplerParameters(int tile, bool linear_filter, uint32_t cms, uint32_t cmt) {
     TextureDataMetal* texture_data = &mTextures[mCurrentTextureIds[tile]];
     texture_data->linear_filtering = linear_filter;
+    texture_data->filtering = !linear_filter ? FILTER_LINEAR : FILTER_THREE_POINT;
```
briaguya0, line 356: this feels backwards, why are we setting `filtering` to `FILTER_LINEAR` if `linear_filter` is falsy?

KiritoDv: That is just for linear vs bilinear, three point uses linear filtering + shader changes

KiritoDv: OpenGL does the same

briaguya0: it looks like you added that in https://github.com/Kenix3/libultraship/pull/873, it still feels very unintuitive,

it'd be nice to have comments documenting this. i left "no idea" comments after restructuring this to an if statement, if you can fill those in that'd go a long way.
```cpp
    if (linear_filter) {
        // no idea how to explain why linear_filter means set linear filter to true and filtering to three point
        texture_data->linear_filtering = true;
        texture_data->filtering = FILTER_THREE_POINT;
    } else {
        // no idea how to explain why !linear_filter means set filtering to FILTER_LINEAR
        texture_data->linear_filtering = false;
        texture_data->filtering = FILTER_LINEAR;
    }
```

briaguya0 (approved):

PR #1013, opened 2026-02-21
Impl ResourceManager::WriteResource

MegaMech:
`ArchiveManager::WriteFile()` has always had one glaring issue. It did not properly reset the file in the cache.

Thus,

`ResourceManager::WriteResource()` is now the proper function to add a new file or update a file in an o2r. This function will unload the resource after writing to disk. This will force LUS to load the file next time it gets used. Prior to this PR, it was required to reset the entire file system, or restart the game.

`ArchiveManager::WriteFile()` Now only writes the file to disk, and ensures the file is indexed.

This allows `WriteFile` to be used more generically for other use-cases as seen requested prior.

## OLD USAGE
(This usage is technically still valid. Especially, if you're writing to a lot of files, and wanting to reset the file system later)
```cpp
bool wrote = GameEngine::Instance->context->GetResourceManager()->GetArchiveManager()->WriteFile(track->Archive, sceneFile, bytes);
if (wrote) {
    // Tell the cache this file needs to be reloaded
    auto resource = GameEngine::Instance->context->GetResourceManager()->GetCachedResource(sceneFile);
    if (resource) {
        resource->Dirty();
    }
}
```
## NEW USAGE
```cpp
// ResourceIdentifier will make it more accurate to make sure you're getting the right file.
// This will help prevent unloading or loading the wrong file.
Ship::ResourceIdentifier id(sceneFile, 0, track->Archive);
// Write file to disk
auto rm = GameEngine::Instance->context->GetResourceManager();
//                                  Unload file after write
bool wrote = rm->WriteResource(id, bytes, true);
if (!wrote) {
    SPDLOG_INFO("[SceneManager] [SaveLevel] Failed to write file!");
}
```

### DIFF
--- a/src/ship/resource/ResourceManager.cpp
+++ b/src/ship/resource/ResourceManager.cpp
@@ -408,7 +408,7 @@ size_t ResourceManager::UnloadResource(const std::string& filePath) {
     return UnloadResource({ filePath, mDefaultCacheOwner, mDefaultCacheArchive });
 }
 
-bool ResourceManager::WriteResource(const ResourceIdentifier& identifier, const std::vector<uint8_t>& data) {
+bool ResourceManager::WriteResource(const ResourceIdentifier& identifier, const std::vector<uint8_t>& data, bool unloadFile) {
     std::shared_ptr<Archive> archive = identifier.Parent;
 
     if (!archive) {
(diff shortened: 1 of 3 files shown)


src/ship/resource/ResourceManager.cpp
```diff
@@ -408,6 +408,26 @@ size_t ResourceManager::UnloadResource(const std::string& filePath) {
...
+
+    if (!archive) {
+        return false;
+    }
+
+    if (!mArchiveManager->WriteFile(archive, identifier.Path, data)) {
+        return false;
+    }
+
+    UnloadResource(identifier);
```
Kenix3: Why unload the resource after writing? Seems to me it should be a flag at the minimum.

PR #1015, opened 2026-03-07
Add mipmap support via bypass, stride-aware decoding, and runtime palettes to Fast interpreter

JeodC:
## Summary

**Stride-aware texture decoding**: Replace linear 1D decode loops with 2D stride-aware iteration in ImportTextureIA4, ImportTextureIA8, ImportTextureI8, and ImportTextureCi4. Sub-texture loads from larger source images (texture atlases, CI textures with shared palettes) now index correctly using fullImageLineSizeBytes as the source stride. Matches the pattern already used by ImportTextureRgba16 and ImportTextureI4.

**LoadBlock DRAM stride**: When SetTextureImage provides a real pixel width (not the dummy width=1 from gDPLoadTextureBlock), compute the actual per-line DRAM stride and store it. Import functions and UV normalization use this stride for textures whose pixel width isn't 8-byte aligned, falling back to TMEM tile stride when per-line info is unavailable.

**Mipmap support**: Clamp imported texture dimensions to SetTileSize bounds so mipmap pyramid loads (which include all LOD levels in a single LoadBlock) only decode the base mip level. Force both texture slots to the base tile when first_tile_index >= 2 (mipmap indicator), preventing the second slot from reading stale data. Apply the same dimension clamping to UV normalization in the triangle draw loop, and use effective_tile for vertex UV shift/offset lookups.

**TLUT format override**: Check other_mode_h for active TLUT mode (G_TT_RGBA16/G_TT_IA16) before cache key construction and format dispatch. When TLUT is active, override fmt to G_IM_FMT_CI so palette-indexed textures aren't decoded as RGBA.

**Runtime palette support**: Rewrite GfxDpLoadTlut to copy TLUT data into palette_staging buffers at the correct byte offset based on tile TMEM address, supporting multi-palette CI4 models. Add palette_staging[2][256] to RDP struct. Add null palette guards to ImportTextureCi4 and ImportTextureCi8. Fix CI4 cache key to only include the relevant palette half, avoiding spurious invalidation when the other half changes.

**TMEM slot fallback**: When a tile has no texture loaded in its assigned TMEM slot, fall back to the other slot (tmem_index ^ 1). The N64 RDP's 4KB TMEM is shared across all tiles; this approximates that behavior for multi-tile rendering setups.

**Zero-dimension guards**: Early return in UploadTexture (D3D11 and OpenGL) when width or height is 0, preventing API errors from truncated integer division.

**Safe \_\_OTR\_\_ signature check**: Replace strncmp with byte-by-byte comparison using short-circuit evaluation. Games may pass raw heap pointers as texture addresses; strncmp is unsafe on arbitrary memory. Also filter obviously invalid addresses (null, low, truncated 32-bit, kernel-range) before attempting the check.

**Unresolved segment skip**: Skip SetTextureImage when SegAddr returns an address still in the N64 segmented range, indicating the segment register hasn't been initialized yet.

**Raw texture scale defaults**: Initialize h_byte_scale and v_pixel_scale to 1 for non-OTR textures, preventing GfxDpLoadBlock from multiplying size_bytes by 0.

**OTR filepath handler guard**: Validate w1 is a valid \_\_OTR\_\_ path before dispatching OTR filepath handlers, preventing crashes from non-OTR GBI commands that happen to match OTR opcodes.

### DIFF
--- a/include/fast/backends/gfx_direct3d_common.h
+++ b/include/fast/backends/gfx_direct3d_common.h
@@ -155,7 +155,7 @@ class GfxRenderingAPIDX11 final : public GfxRenderingAPI {
 
     std::vector<struct TextureData> mTextures;
     int mCurrentTile;
-    uint32_t mCurrentTextureIds[SHADER_MAX_TEXTURES];
+    uint32_t mCurrentTextureIds[SHADER_MAX_TEXTURES]{};
 
     std::vector<FramebufferDX11> mFrameBuffers;
 
--- a/include/fast/backends/gfx_direct3d_common.h
+++ b/include/fast/backends/gfx_direct3d_common.h
@@ -176,6 +176,11 @@ class GfxRenderingAPIDX11 final : public GfxRenderingAPI {
     Microsoft::WRL::ComPtr<ID3D11SamplerState> mLastSamplerStates[SHADER_MAX_TEXTURES] = { nullptr, nullptr };
 
     D3D_PRIMITIVE_TOPOLOGY mLastPrimitaveTopology = D3D_PRIMITIVE_TOPOLOGY_UNDEFINED;
+
+    // Cached staging texture for ReadFramebufferToCPU — avoids CreateTexture2D/Release per frame
+    Microsoft::WRL::ComPtr<ID3D11Texture2D> mReadbackStaging;
+    uint32_t mReadbackStagingW = 0;
+    uint32_t mReadbackStagingH = 0;
 };
 
 std::string gfx_direct3d_common_build_shader(size_t& numFloats, const CCFeatures& cc_features,
--- a/include/fast/backends/gfx_metal.h
+++ b/include/fast/backends/gfx_metal.h
@@ -199,6 +199,7 @@ class GfxRenderingAPIMetal final : public GfxRenderingAPI {
     size_t mCoordBufferSize;
     MTL::Function* mDepthComputeFunction;
     MTL::Function* mConvertToRgb5a1Function;
+    MTL::ComputePipelineState* mConvertToRgb5a1PipelineState = nullptr;
 
     // Current state
     struct ShaderProgramMetal* mShaderProgram;
--- a/include/fast/backends/gfx_metal.h
+++ b/include/fast/backends/gfx_metal.h
@@ -199,7 +199,6 @@ class GfxRenderingAPIMetal final : public GfxRenderingAPI {
     size_t mCoordBufferSize;
     MTL::Function* mDepthComputeFunction;
     MTL::Function* mConvertToRgb5a1Function;
-    MTL::ComputePipelineState* mConvertToRgb5a1PipelineState = nullptr;
 
     // Current state
     struct ShaderProgramMetal* mShaderProgram;
--- a/include/fast/backends/gfx_opengl.h
+++ b/include/fast/backends/gfx_opengl.h
@@ -107,7 +107,7 @@ class GfxRenderingAPIOGL final : public GfxRenderingAPI {
     void SetPerDrawUniforms();
 
     std::vector<TextureInfo> textures;
-    GLuint mCurrentTextureIds[SHADER_MAX_TEXTURES];
+    GLuint mCurrentTextureIds[SHADER_MAX_TEXTURES]{};
     GLuint mLastBoundTextures[SHADER_MAX_TEXTURES] = {};
     uint8_t mCurrentTile;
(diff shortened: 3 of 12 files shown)


Kenix3 (approved):
This looks good to me. Does anyone else have any thoughts?

briaguya0:
i'd love to see a testing PR using this, just want to make sure released ports still work as expected with it (or we at least know what changes we need to make to those to support this)

Malkierian:
> i'd love to see a testing PR using this, just want to make sure released ports still work as expected with it (or we at least know what changes we need to make to those to support this)

I assume you mean, then, that you'd want a PR on each current Torch port?

BoiledElectricity:
I tested  PR my Harvest Moon 64 PC port (BlueFeather). It fixed a
crash on my Windows DX11 build. The zero-init, bounds guard, and moving the call inside the
  usedTextures check all make sense. Would love to see these land upstream!

BoiledElectricity:
@JeodC  Also two edge cases I hit on Harvest moon port that might be worth adding, no worries if they're out of
  scope for this PR:

  1. CI8 result width still uses texture_tile but the stride-aware loop packs at
  loaded_texture width, wrong upload dimensions for tightly-packed LoadTile textures.
  2. GfxSpTri1 tex_width , same mismatch on the sampling side, causes diagonal shredding.

  Both are no-ops when strides match. Happy to share diffs if useful, otherwise I can open
  a separate PR.
  here is what we currently get, 
Screenshot 2026-03-17 at 1 03 06 PM

 After (result width uses loaded_texture for LoadTile, texture_tile fallback for
  LoadBlock):
  
Screenshot 2026-03-17 at 1 04 27 PM

JeodC:
I am thinking this PR can easily be built on for other ports. However, I am not currently working on other ports so it is out of my own scope. If you want to submit a pull request with additions to my fork of LUS and target the branch for *this* pull request, go for it.

Otherwise, I am with @briaguya0 and would like to see if this breaks current ports before merge or piling on more edge case solutions.

Love to see Harvest Moon booting!

BoiledElectricity:
> I am thinking this PR can easily be built on for other ports. However, I am not currently working on other ports so it is out of my own scope. If you want to submit a pull request with additions to my fork of LUS and target the branch for _this_ pull request, go for it.
> 
> Otherwise, I am with @briaguya0 and would like to see if this breaks current ports before merge or piling on more edge case solutions.
> 
> Love to see Harvest Moon booting!

Sounds good! And thank you! its been quite the experience haha

briaguya0:
> I assume you mean, then, that you'd want a PR on each current Torch port?

i don't think torch vs non-torch matters for this one, a testing PR to soh would do the trick imo

briaguya0:
tested this on https://github.com/briaguya0/Shipwright/commit/32287802272dc6c955ba49072f0fe1a404c4865b, build from https://github.com/briaguya0/Shipwright/actions/runs/23546513299 resulted in

image

while latest dev (https://github.com/HarbourMasters/Shipwright/commit/8c4d4738cc2aae9e16b183b9c9ff702f8432b74c) worked as expected

image

JeodC:
I am thinking @JamesAnthonyDellaMorte's solution will fix this, but I need to figure out what they did.

JeodC:
> tested this on [briaguya0/Shipwright@3228780](https://github.com/briaguya0/Shipwright/commit/32287802272dc6c955ba49072f0fe1a404c4865b), build from https://github.com/briaguya0/Shipwright/actions/runs/23546513299 resulted in
> 
> image
> while latest dev ([HarbourMasters/Shipwright@8c4d473](https://github.com/HarbourMasters/Shipwright/commit/8c4d4738cc2aae9e16b183b9c9ff702f8432b74c)) worked as expected
> 
> image

The two commits just added resolved this. :) Thanks for testing!

JeodC:
Cherry picked from #1018:

- **DX11 `ReadFramebufferToCPU`**: Cache the staging texture to avoid per-frame `CreateTexture2D`/`Release`. Handle dimension mismatch between source texture and requested output size.
- **OpenGL `ReadFramebufferToCPU`**: Read as `GL_UNSIGNED_BYTE` then convert to RGBA5551, working around an NVIDIA driver issue where `GL_UNSIGNED_SHORT_5_5_5_1` writes 8 bytes per pixel instead of 2.
- **`GfxDpFillRectangle`**: Use `mNativeDimensions` instead of hardcoded 319/239 for fullscreen fill rect widescreen expansion.

PR #1017, opened 2026-03-15
perf: avoid string allocations in CheckCache hot path via heterogeneo…

koplas:
…us map lookup

  Every call to CheckCache(const std::string&) previously constructed a full
  ResourceIdentifier, copying the path string onto the heap, computing a hash,
  then immediately freeing it on cache hits.

  Fix using C++20 heterogeneous unordered_map lookup:

  - Add ResourceLookupKey: a lightweight non-owning struct holding a string_view, owner, and raw Archive* — zero heap allocation.
  - Add is_transparent to ResourceIdentifierHash (adds operator() for ResourceLookupKey) and add ResourceIdentifierEqual with is_transparent (handles ResourceIdentifier vs ResourceLookupKey comparisons).
  - Update mResourceCache to use ResourceIdentifierEqual as its key equality.
  - CheckCache(const std::string&) now does the find() via ResourceLookupKey, skipping ResourceIdentifier construction entirely on cache hits.
  - Align CalculateHash() to use std::hash<std::string_view> for hash consistency with ResourceLookupKey across all standard library implementations.
  - Use constexpr length in OtrSignatureCheck, removing a redundant strlen().

### DIFF
--- a/include/ship/resource/ResourceManager.h
+++ b/include/ship/resource/ResourceManager.h
@@ -3,6 +3,7 @@
 #include <unordered_map>
 #include <unordered_set>
 #include <string>
+#include <string_view>
 #include <list>
 #include <vector>
 #include <mutex>
@@ -47,8 +48,28 @@ struct ResourceIdentifier {
     size_t mHash;
 };
 
+// Lightweight non-owning key for cache lookups — avoids copying the path string.
+// Used with heterogeneous unordered_map::find() to skip ResourceIdentifier construction.
+struct ResourceLookupKey {
+    std::string_view Path;
+    uintptr_t Owner = 0;
+    const Archive* ParentRaw = nullptr;
+    size_t Hash = 0;
+
+    ResourceLookupKey(std::string_view path, uintptr_t owner, const std::shared_ptr<Archive>& parent);
+};
+
 struct ResourceIdentifierHash {
+    using is_transparent = void;
     size_t operator()(const ResourceIdentifier& rcd) const;
+    size_t operator()(const ResourceLookupKey& key) const;
+};
+
+struct ResourceIdentifierEqual {
+    using is_transparent = void;
+    bool operator()(const ResourceIdentifier& a, const ResourceIdentifier& b) const;
+    bool operator()(const ResourceIdentifier& a, const ResourceLookupKey& b) const;
+    bool operator()(const ResourceLookupKey& a, const ResourceIdentifier& b) const;
 };
 
 class ResourceManager {
@@ -131,7 +152,7 @@ class ResourceManager {
 
   private:
     std::unordered_map<ResourceIdentifier, std::variant<ResourceLoadError, std::shared_ptr<IResource>>,
-                       ResourceIdentifierHash>
+                       ResourceIdentifierHash, ResourceIdentifierEqual>
         mResourceCache;
     std::shared_ptr<ResourceLoader> mResourceLoader;
     std::shared_ptr<ArchiveManager> mArchiveManager;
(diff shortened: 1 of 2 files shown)


koplas:
Github is currently broken and I can't convert this to a draft. This doesn't really fix the stutter issues. Please ignore for now.

vanfanel:
> Github is currently broken and I can't convert this to a draft. This doesn't really fix the stutter issues. Please ignore for now.

Ok, I will test when it's ready :)

briaguya0:
> This doesn't really fix the stutter issues.

i assume the stutter issues you're referring to are the ones i tried to make a workaround for with https://github.com/briaguya0/Shipwright/pull/374 and @Malkierian was looking into with https://github.com/HarbourMasters/Shipwright/pull/4652

my understanding is that the actual performance hit happens because we're loading high resolution textures saved in n64 texture format, then translating those into things that can be rendered, and all this is happening "on demand" as resources are needed for a scene.

both malk and I tried to address this by preloading resources needed for scenes. people have discussed mipmapping as another possible solution but that would require some more extensive changes.

my general thought is that the changes in this PR would impact both original and high resolution textures, so unless you're seeing stuttering with the original resources it'll be tough to see what impact this is having

koplas:
> > This doesn't really fix the stutter issues.
> 
> i assume the stutter issues you're referring to are the ones i tried to make a workaround for with [briaguya0/Shipwright#374](https://github.com/briaguya0/Shipwright/pull/374) and @Malkierian was looking into with [HarbourMasters/Shipwright#4652](https://github.com/HarbourMasters/Shipwright/pull/4652)
> 
> my understanding is that the actual performance hit happens because we're loading high resolution textures saved in n64 texture format, then translating those into things that can be rendered, and all this is happening "on demand" as resources are needed for a scene.
> 
> both malk and I tried to address this by preloading resources needed for scenes. people have discussed mipmapping as another possible solution but that would require some more extensive changes.

I'm currently testing two patches, one that reduces per-frame allocations and another that adds OpenGL shader caching. Based on my profiling so far, if neither yields a meaningful improvement, the next area to investigate would be resource loading, similar to the work you've done previously.

> my general thought is that the changes in this PR would impact both original and high resolution textures, so unless you're seeing stuttering with the original resources it'll be tough to see what impact this is having

@vanfanel Do you experience the stutter when using original resolution textures?

vanfanel:
@koplas Yes, all my tests are done using the original game textures.

koplas:
I am closing this. As the majority of the time is spent in the OS and not the game, the benefits of this pull request are minimal. But I will open a second pull request that considerably reduces the amount of temporary allocations, which could help on systems with slow allocators.
My suspicion is that the stutter @vanfanel experiences can only solved by the game side, if we preload the resources, which is something @briaguya0 and @Malkierian already tried in the past.

vanfanel:
> I am closing this. As the majority of the time is spent in the OS and not the game, the benefits of this pull request are minimal. But I will open a second pull request that considerably reduces the amount of temporary allocations, which could help on systems with slow allocators. My suspicion is that the stutter @vanfanel experiences can only solved by the game side, if we preload the resources, which is something @briaguya0 and @Malkierian already tried in the past.

Hm, what is a system with slow allocators exactly?

I mean: I am on a Ryzen 7 8700G with 2 x 32GB DDR5 5600 RAM, which is a pretty fast machine by today standards, and I run Debian GNU/Linux on it, a very light system on which only a lightweight Wayland compositor runs (latest labwc 0.9.6 from last week), Linux kernel 6.19.5 and latest stable MESA 26.0.3. No telemetry, no strange services.

So, what could make a system "slow" for allocation?

koplas:
A large portion of the [flamegraph](https://github.com/HarbourMasters/Shipwright/issues/6048#issuecomment-4050570951) is dominated by allocator overhead. The most expensive functions (such as `CheckCache`) generate significant temporary allocations, which puts pressure on the allocator and accounts for most of the time spent there. Whether the latency spike stems from the driver alone or a combination of the driver and game logic is hard to say without being able to reproduce it on the same hardware.

PR #1022, opened 2026-03-20
Avoid temporary string allocations in hot per-frame paths

koplas:
- ConsoleVariable::Get: use transparent hash/equal on mVariables so find(const char*) no longer constructs a temporary std::string
- ResourceIdentifier: add move constructor so altPath strings are moved rather than copied; use += instead of operator+ to avoid a second allocation
- Interpreter::GetBaseTexturePath: return std::string_view instead of std::string to eliminate per-call copies; use std::less<> on mMaskedTextures for zero-alloc heterogeneous lookup

Can help on platforms with slow allocators.

### DIFF
--- a/include/fast/interpreter.h
+++ b/include/fast/interpreter.h
@@ -9,6 +9,7 @@
 #include <vector>
 #include <stack>
 #include <string>
+#include <string_view>
 
 #include "fast/lus_gbi.h"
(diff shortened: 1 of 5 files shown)


Kenix3 (approved):

Kenix3:
Merge lending conflict resolution

PR #1023, opened 2026-03-21
Use sysctl to lookup kern.clockrate on OpenBSD & Mac

serprex:
see https://github.com/HarbourMasters/Shipwright/pull/6364#issuecomment-4103107166

use with mac too

### DIFF
--- a/include/fast/backends/gfx_sdl.h
+++ b/include/fast/backends/gfx_sdl.h
@@ -1,5 +1,9 @@
 #pragma once
 
+#if defined(__OpenBSD__) || defined(__APPLE__)
+#define BSD_TICKS_ENABLED 1
+#endif
+
 #include "gfx_window_manager_api.h"
 namespace Fast {
 class GfxWindowBackendSDL2 final : public GfxWindowBackend {
@@ -57,6 +61,9 @@ class GfxWindowBackendSDL2 final : public GfxWindowBackend {
     int mSdlToLusTable[512];
     float mMouseWheelX = 0.0f;
     float mMouseWheelY = 0.0f;
+#ifdef BSD_TICKS_ENABLED
+    int mBsdTick; // store kern.clockrate's tick (microseconds) to adjust sleep timing
+#endif
     // OTRTODO: These are redundant. Info can be queried from SDL.
     int mWindowWidth = 640;
     int mWindowHeight = 480;
--- a/include/fast/backends/gfx_sdl.h
+++ b/include/fast/backends/gfx_sdl.h
@@ -1,9 +1,5 @@
 #pragma once
 
-#if defined(__OpenBSD__) || defined(__APPLE__)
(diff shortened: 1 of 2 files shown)


src/fast/backends/gfx_sdl2.cpp
```diff
@@ -666,14 +679,13 @@ void GfxWindowBackendSDL2::SyncFramerateWithTime() const {
...
 
-#ifdef _WIN32
     t = qpc_to_100ns(SDL_GetPerformanceCounter());
+#ifdef _WIN32
     while (t < next) {
         YieldProcessor(); // TODO: Find a way for other compilers, OSes and architectures
         t = qpc_to_100ns(SDL_GetPerformanceCounter());
     }
 #endif
-    t = qpc_to_100ns(SDL_GetPerformanceCounter());
```
serprex, line 676: unrelated, but avoids an unnecessary extra call on windows

include/fast/backends/gfx_sdl.h
```diff
@@ -1,5 +1,9 @@
 #pragma once
 
+#if defined(__BSD__) || defined(__APPLE__)
```
fabienr: `__BSD__` isn't defined, `BSD` isn't standard (gcc only) and `unix` include Linux which isn't wanted.
Replace with `__OpenBSD__` works but only for openbsd.

```
$ cpp -dM /dev/null | grep -e unix -e BSD
#define __OpenBSD__ 1
#define __VERSION__ "OpenBSD Clang 19.1.7"
#define __unix 1
#define __unix__ 1
#define unix 1
```

src/fast/backends/gfx_sdl2.cpp
```diff
@@ -649,9 +663,8 @@ void GfxWindowBackendSDL2::SyncFramerateWithTime() const {
 #ifdef _WIN32
     // We want to exit a bit early, so we can busy-wait the rest to never miss the deadline
     left -= 15000UL;
-#elif defined(__APPLE__)
-    // Use macOS scheduler interval on macOS
-    left -= 10000UL;
+#elif BSD_TICKS_ENABLED
+    left -= mBsdTick * 1000;
```
fabienr: left is based on qpc_to_100ns, convert with * 10 instead of * 1000

fabienr:
With the two comments addressed, it works on openbsd.

serprex:
https://github.com/giampaolo/psutil/issues/1892#issuecomment-748491277 this person reports tick=10000 on mac mini, which would give 10x fudge factor than what was previously used

serprex:
if we can't get mac testing on this I'd be fine having it be openbsd-only

Kenix3 (approved):

serprex:
friend confirmed they get tick=10000 on mac, so I'm going to label mac as an OS that lies & restrict this to openbsd

PR #1026, opened 2026-03-24
Fix performance regression when using texture packs

koplas:
- Reintroduce check for alt-asset. When a texture pack was active every alt-asset path lookup would result in a cache miss and be queued to the thread pool.

- Revert to copy variant instead of raw pointer to avoid race condition.

Closes #1025

PR #1027, opened 2026-03-24
perf: avoid memory copies in resource loading

koplas:
The resource loading pipeline was making up to three copies of asset data between zip decompression and final use:

1. ReadResourceInitDataLegacy copied the entire file body into a new vector just to strip the OTR header, and separately copied the header into another vector for parsing.

2. TextureFactory allocated a new uint8_t[] buffer and memcpy'd image pixel data out of the already-decompressed file buffer.

This improves performance for large resource packs.

### DIFF
--- a/include/fast/resource/type/Texture.h
+++ b/include/fast/resource/type/Texture.h
@@ -1,6 +1,8 @@
 #pragma once
 
 #include "ship/resource/Resource.h"
+#include <memory>
+#include <vector>
 
 #define TEX_FLAG_LOAD_AS_RAW (1 << 0)
 #define TEX_FLAG_LOAD_AS_IMG (1 << 1)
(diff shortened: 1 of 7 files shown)


koplas:
@Djipi71, @leggettc18 can you also test this pull request in combination with #1026?

leggettc18:
> @Djipi71, @leggettc18 can you also test this pull request in combination with #1026?

In a few hours when I'm off work, yeah.

leggettc18:
So far I'm not noticing any difference on my PC (which is good as it was fine with just 1026). I forgot to double-check the numbers with just 1026 alone, so I'm rebuilding that now to see if there's any noticeable difference in memory usage or anything.

leggettc18:
Yeah I'm not really able to notice any difference between just 1026 and with this merged in. That's probably a good thing, since like I said, 1026 was already good so it getting better isn't really something I'd be able to notice very easily.

koplas:
> Yeah I'm not really able to notice any difference between just 1026 and with this merged in. That's probably a good thing, since like I said, 1026 was already good so it getting better isn't really something I'd be able to notice very easily.

Good to know, it is more an improvement for big textures. I am marking this as ready for review.

Kenix3 (approved):

PR #1033, opened 2026-03-25
tests: add initial unit test suite

briaguya0:
Adds a minimal, opt-in unit test suite as a starting point for testing libultraship.

`LUS_BUILD_TESTS` defaults to `OFF` so ports consuming LUS as a library are unaffected.

## Blocked on

- https://github.com/Kenix3/libultraship/pull/1029 — the prism update includes a fix that prevents prism from leaking its own CTest registration into consumers
- https://github.com/Kenix3/libultraship/pull/1030 — fix for `StringHelper::Strip` erase count bug surfaced by these tests
- https://github.com/Kenix3/libultraship/pull/1032 — fix for `StringHelper::Implode` returning `""` unconditionally, surfaced by these tests

The fixes from #1030 and #1032 are included in this branch for now — we'll update based on how those reviews go.

## What's tested and why

The three test files cover the [starting points](https://gist.github.com/briaguya0/f7f8f068069089a17967459b58741c70#starting-points) identified in the exploration gist — targets with no SDL, OpenGL, or `Context` dependencies:

- **BinaryReader/BinaryWriter** — endianness, round-trips, NaN handling, seek modes
- **StringHelper** — split, strip, replace, predicate helpers
- **CRC64** — golden-value pinning. The algorithm feeds every archive lookup; a silent change would break all OTR/O2R files.

## For reviewers

- Is the CMake structure right — `LUS_BUILD_TESTS` placement, test target link dependencies?
- GoogleTest v1.16.0 — any concerns?
- CI currently runs tests on Linux only — acceptable to merge in this state?

### DIFF
--- a/.github/workflows/test-validation.yml
+++ b/.github/workflows/test-validation.yml
@@ -1,0 +1,59 @@
+name: test-validation
+on:
+  push:
+  pull_request:
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+jobs:
+  test-linux:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y $(cat .github/workflows/apt-deps.txt)
+    - name: ccache
+      uses: hendrikmuhs/ccache-action@v1.2.14
+      with:
+        key: ${{ runner.os }}-test-ccache-${{ github.ref }}-${{ github.sha }}
+        restore-keys: |
+          ${{ runner.os }}-test-ccache-${{ github.ref }}
+          ${{ runner.os }}-test-ccache-
+    - name: Install latest SDL
+      run: |
+        export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
+        wget https://www.libsdl.org/release/SDL2-2.24.1.tar.gz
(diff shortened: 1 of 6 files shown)


Kenix3:
Google test should be fine. Wondering if we want a separate section in the readme for utilities that don't get compelled in.

briaguya0:
alright, rebased the branch so we only have the new test changes now that the fixes from those PRs are in.

_____

sounds like for the "for reviewers" questions

> Is the CMake structure right — `LUS_BUILD_TESTS` placement, test target link dependencies?

:grey_question: still an open question

> GoogleTest v1.16.0 — any concerns?

:white_check_mark: sounds like we're cool with using this

> CI currently runs tests on Linux only — acceptable to merge in this state?

:grey_question: still an open question. good thing is that google test appears to run on/support all our supported platforms
* https://google.github.io/googletest/platforms.html
* https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md

_____

other note, google test [is BSD-3](https://github.com/google/googletest/blob/main/LICENSE) which is fine

briaguya0:
> Wondering if we want a separate section in the readme for utilities that don't get compelled in.

something like https://github.com/briaguya0/libultraship/commit/82c3478154841af5fbc4d7bd8f7679e17deb7fd3 ?

Kenix3:
Yeah definitely cool on Google test. The small quibble(which this documents intent to fix) can be addressed as a separate concern.

Kenix3 (commented):
I will add that I also like the Cmake structure.

PR #1035, opened 2026-03-26
Add AudioDma registry to clamp osPiStartDma reads to source blob bounds

JeodC:
On N64, DMA reads from ROM which has no bounds — audio sample reads (fixed 0x200/0x270 byte chunks) can extend past the source data. On PC the source is a heap allocation with exact bounds. `AudioDma_Register` records soundfont blob address ranges, and `AudioDma_Clamp` in `osPiStartDma` limits memcpy to the available data, zero-filling the remainder.

This prevents audio leaks for any port that uses the N64 `libaudio` directly and was discovered with ASAN.


Kenix3 (approved):

Kenix3:
Looks like clang-format needs to be run

JeodC:
> Looks like clang-format needs to be run 

Done

PR #1040, opened 2026-04-01
Backend safety guards: null-checks, zero-dimension guards, and resource leak fixes

JeodC:
This is 1/7 of splitting up #1038 into easily digestible chunks for easier review and bisecting.

- Value-initialize `mCurrentTextureIds` arrays in DX11 and OpenGL to prevent stale out-of-bounds reads
- Move `PSSetSamplers` inside the `usedTextures[i]` guard so it doesn't index with a stale texture ID
- Early-return from `UploadTexture` in all three backends (DX11, OpenGL, Metal) when width or height is zero
- Default `h_byte_scale` and `v_pixel_scale` to 1 for raw N64 textures to avoid divide-by-zero in CI paths
- Null-check the `Context->Window->Gui` chain in DXGI `wnd_proc` and `WM_KILLFOCUS` for teardown safety
- Guard `mPendingFrameStats.rbegin()` against empty container
- Iteratively erase frame-stat containers in `Destroy()` to avoid MSVC recursive `_Erase_tree` stack overflow
- Unmap and Release the DX11 staging texture on early-return in `ReadFramebufferToCPU`

### DIFF
--- a/include/fast/backends/gfx_direct3d_common.h
+++ b/include/fast/backends/gfx_direct3d_common.h
@@ -155,7 +155,7 @@ class GfxRenderingAPIDX11 final : public GfxRenderingAPI {
 
     std::vector<struct TextureData> mTextures;
     int mCurrentTile;
-    uint32_t mCurrentTextureIds[SHADER_MAX_TEXTURES];
+    uint32_t mCurrentTextureIds[SHADER_MAX_TEXTURES]{};
 
     std::vector<FramebufferDX11> mFrameBuffers;
 
--- a/include/fast/backends/gfx_direct3d_common.h
+++ b/include/fast/backends/gfx_direct3d_common.h
@@ -155,7 +155,7 @@ class GfxRenderingAPIDX11 final : public GfxRenderingAPI {
 
     std::vector<struct TextureData> mTextures;
     int mCurrentTile;
-    uint32_t mCurrentTextureIds[SHADER_MAX_TEXTURES]{};
+    uint32_t mCurrentTextureIds[SHADER_MAX_TEXTURES] = {};
 
     std::vector<FramebufferDX11> mFrameBuffers;
 
--- a/include/fast/backends/gfx_opengl.h
+++ b/include/fast/backends/gfx_opengl.h
@@ -107,7 +107,7 @@ class GfxRenderingAPIOGL final : public GfxRenderingAPI {
     void SetPerDrawUniforms();
 
     std::vector<TextureInfo> textures;
-    GLuint mCurrentTextureIds[SHADER_MAX_TEXTURES]{};
+    GLuint mCurrentTextureIds[SHADER_MAX_TEXTURES] = {};
     GLuint mLastBoundTextures[SHADER_MAX_TEXTURES] = {};
     uint8_t mCurrentTile;
     int8_t mLastActiveTexture = -1;
--- a/include/fast/backends/gfx_opengl.h
+++ b/include/fast/backends/gfx_opengl.h
@@ -107,7 +107,7 @@ class GfxRenderingAPIOGL final : public GfxRenderingAPI {
     void SetPerDrawUniforms();
 
     std::vector<TextureInfo> textures;
-    GLuint mCurrentTextureIds[SHADER_MAX_TEXTURES];
+    GLuint mCurrentTextureIds[SHADER_MAX_TEXTURES]{};
     GLuint mLastBoundTextures[SHADER_MAX_TEXTURES] = {};
     uint8_t mCurrentTile;
     int8_t mLastActiveTexture = -1;
--- a/src/fast/backends/gfx_direct3d11.cpp
+++ b/src/fast/backends/gfx_direct3d11.cpp
@@ -1047,6 +1047,8 @@ void GfxRenderingAPIDX11::ReadFramebufferToCPU(int fb_id, uint32_t width, uint32
     ThrowIfFailed(mContext->Map(staging, 0, D3D11_MAP_READ, 0, &resource));
 
     if (!resource.pData) {
+        mContext->Unmap(staging, 0);
+        staging->Release();
         return;
     }
 
--- a/src/fast/backends/gfx_direct3d11.cpp
+++ b/src/fast/backends/gfx_direct3d11.cpp
@@ -727,8 +727,8 @@ void GfxRenderingAPIDX11::DrawTriangles(float buf_vbo[], size_t buf_vbo_len, siz
                     mLastSamplerStates[i] = mTextures[mCurrentTextureIds[i]].sampler_state.Get();
                 }
             }
+            mContext->PSSetSamplers(i, 1, mTextures[mCurrentTextureIds[i]].sampler_state.GetAddressOf());
         }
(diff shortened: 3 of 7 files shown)


include/fast/backends/gfx_opengl.h
```diff
@@ -107,7 +107,7 @@ class GfxRenderingAPIOGL final : public GfxRenderingAPI {
     void SetPerDrawUniforms();
 
     std::vector<TextureInfo> textures;
-    GLuint mCurrentTextureIds[SHADER_MAX_TEXTURES];
+    GLuint mCurrentTextureIds[SHADER_MAX_TEXTURES]{};
```
briaguya0: @Kenix3 do you have a preference between `{}` vs `= {}` for these? the line right under this uses `= {}`. i don't think it's something `clang-format` (or even `clang-tidy`) handles.

src/fast/backends/gfx_direct3d11.cpp
```diff
@@ -708,6 +712,10 @@ void GfxRenderingAPIDX11::DrawTriangles(float buf_vbo[], size_t buf_vbo_len, siz
 
     for (int i = 0; i < SHADER_MAX_TEXTURES; i++) {
         if (mShaderProgram->usedTextures[i]) {
+            // Guard against stale/invalid texture IDs
+            if (mCurrentTextureIds[i] >= mTextures.size()) {
+                continue;
+            }
```
briaguya0: this check feels like it's making a lot of assumptions. for example (i don't think it's possible for this to happen with the code as we have it now, but it's the first thing that came to my mind):
* texture A is added to mTextures, gets id 1
* texture B is added to mTextures, gets id 2
* texture C is added to mTextures, gets id 3
* somehow texture A is removed from mTextures, now texture C goes in and hits the `continue`

i think just adding a bit more detail to the comment explaining why something like that wouldn't happen would go a long way towards making this feel less scary

JeodC:
Ok I ran `find src include -name "*.cpp" -o -name "*.h" | sed 's| |\\ |g' | xargs clang-format -i` and it still failed, I suck at this.

src/fast/backends/gfx_dxgi.cpp
```diff
@@ -357,7 +357,10 @@ static LRESULT CALLBACK gfx_dxgi_wnd_proc(HWND h_wnd, UINT message, WPARAM w_par
     char fileName[256];
     Ship::WindowEvent event_impl;
     event_impl.Win32 = { h_wnd, static_cast<int>(message), static_cast<int>(w_param), static_cast<int>(l_param) };
-    Ship::Context::GetInstance()->GetWindow()->GetGui()->HandleWindowEvents(event_impl);
+    auto ctx = Ship::Context::GetInstance();
+    if (ctx && ctx->GetWindow() && ctx->GetWindow()->GetGui()) {
+        ctx->GetWindow()->GetGui()->HandleWindowEvents(event_impl);
+    }
```
briaguya0, line 363: this guard makes perfect sense, but i don't love it as a pattern. what i'm really hoping to avoid is seeing a pattern like this replace every `Ship::Context::GetInstance()->GetX()->GetY()`. @Kenix3 any thoughts on this one? maybe something with the component system you're working on like a "safegetcomponent" or something?

src/fast/backends/gfx_dxgi.cpp
```diff
@@ -1068,7 +1073,14 @@ bool GfxWindowBackendDXGI::CanDisableVsync() {
...
 void GfxWindowBackendDXGI::Destroy() {
-    // TODO: destroy _any_ resources used by dxgi, including the window handle
+    // Iteratively clear frame-stat containers to avoid MSVC's recursive _Erase_tree
+    // stack overflow on deep std::set/std::map trees during destruction.
+    while (!mPendingFrameStats.empty()) {
+        mPendingFrameStats.erase(mPendingFrameStats.begin());
+    }
+    while (!mFrameStats.empty()) {
+        mFrameStats.erase(mFrameStats.begin());
+    }
```
briaguya0, line 1084: fix seems reasonable, just curious about how `mPendingFrameStats` is getting big enough to hit the stack overflow issue

briaguya0 (commented):
thanks for breaking these up! left a couple comments with questions but overall this seems great!

include/fast/backends/gfx_opengl.h
JeodC: I'm all for consistency. I swapped it to the latter.

src/fast/backends/gfx_direct3d11.cpp
JeodC: Expanded the comment as requested.

src/fast/backends/gfx_dxgi.cpp
JeodC: Fwiw I only needed this guard here. Maybe it won't become a pattern at all?

src/fast/backends/gfx_dxgi.cpp
JeodC: The caps (40 for pending, 3 for stats) get enforced during presents, but if `Destroy()` runs during a messy teardown — say the window closes before the first present even finishes — the trimming code may never have had a chance to run. And honestly even at modest sizes, MSVC's `_Erase_tree` recurses through every node and `DXGI_FRAME_STATISTICS` isn't tiny, so it doesn't take a huge tree to eat through the stack.

include/fast/backends/gfx_direct3d_common.h
```diff
@@ -155,7 +155,7 @@ class GfxRenderingAPIDX11 final : public GfxRenderingAPI {
 
     std::vector<struct TextureData> mTextures;
     int mCurrentTile;
-    uint32_t mCurrentTextureIds[SHADER_MAX_TEXTURES];
+    uint32_t mCurrentTextureIds[SHADER_MAX_TEXTURES]{};
```
briaguya0:
```suggestion
    uint32_t mCurrentTextureIds[SHADER_MAX_TEXTURES] = {};
```

just to keep this consistent between files

src/fast/backends/gfx_dxgi.cpp
briaguya0: adding that context to the comment (something like "during messy teardowns, it doesn't take a huge tree to eat through the stack") would be perfect

src/fast/backends/gfx_dxgi.cpp
briaguya0: made a follow up issue for it https://github.com/Kenix3/libultraship/issues/1048

briaguya0 (commented):
tiny nit, one comment about a comment, and made a follow up issue. this is 99% ready to merge imo

src/fast/backends/gfx_dxgi.cpp
JeodC: Done.

src/fast/backends/gfx_dxgi.cpp
Kenix3: Yeah I think this is something better fixed with the component system. I'm not sure what a safe return would be, though. I don't really like the idea of having base implementations that have noop in order to get a legal object.

PR #1047, opened 2026-04-03
Event System with Debugger

KiritoDv:
This PR aims to port the event system used in some lus ports to be part of the core systems


Kenix3 (approved):
Anyone else have opinions in this?

PR #1051, opened 2026-04-04
Implemented dynamic loading

KiritoDv:
This PRs aims to implement the PoC system made on ghostship for use on all ports


Kenix3:
While doing an in depth review one of the first things I've noticed is new dependencies. We need to update our readme for these. Makes sense to credit everyone equally

include/libultraship/libultra/gbi.h
```diff
@@ -178,6 +178,9 @@
 #define G_FILLWIDERECT 0x38
 #define G_REGBLENDEDTEX 0x3f
 #define G_MOVEMEM_OTR 0x42
+#define G_LOADBLOCK_WIDE 0x47
```
Kenix3, line 181: Why are these changes in a PR about modding?

include/ship/resource/archive/Archive.h
```diff
@@ -17,6 +17,26 @@ namespace Ship {
...
+struct ArchiveManifest {
+    std::string Name;
+    std::string Icon;
+    std::string Author;
+    std::string Version;
+    std::string Website;
+    std::string Description;
+    std::string License;
+
+    uint32_t CodeVersion;
```
Kenix3, line 29: What's the difference between code version and game version?

include/ship/resource/archive/Archive.h
```diff
@@ -48,11 +71,16 @@ class Archive : public std::enable_shared_from_this<Archive> {
     void SetLoaded(bool isLoaded);
     void SetGameVersion(uint32_t gameVersion);
     void IndexFile(const std::string& filePath);
+    void Validate();
+    std::vector<uint8_t> HexToBytes(const std::string& hex);
```
Kenix3: Feels like this should be in the StringHelper class instead

include/ship/scripting/ScriptLoader.h
```diff
@@ -0,0 +1,24 @@
...
+    ScriptLoader() : mHandle(nullptr), mTempFile("") {
+    }
+
+    std::string GenerateTempFile();
+    void Init(const std::string& path);
+    void* GetFunction(const std::string& name);
+    void Unload();
+
+  protected:
+    ScriptHandle_t mHandle;
```
Kenix3: Data is essentially always private in LUS. What reason do you need protected here?

include/ship/security/KeystoreSystem.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include <string>
+#include <vector>
+#include <cstdint>
+#include <unordered_map>
+
+namespace Ship {
+
+class KeystoreSystem {
```
Kenix3: Wondering if this should just be called "Keystore"

I don't really like the system suffix. They'll all be a Component eventually

src/ship/events/EventSystem.cpp
Kenix3, line 1: I know that this already got in with the name EventSystem. Not sure I like that name, but it's better than KeystoreSystem

src/ship/resource/archive/Archive.cpp
```diff
@@ -47,6 +52,44 @@ void Archive::Load() {
...
+    if (m != nullptr && m->IsLoaded) {
+        try {
+            auto json = nlohmann::json::parse(m->Buffer->begin(), m->Buffer->end());
+            mManifest.Name = json["name"].get<std::string>();
+            mManifest.Icon = json.value("icon", "");
+            mManifest.Author = json.value("author", "Unknown");
+            mManifest.Version = json.value("version", "1.0");
+            mManifest.Website = json.value("website", "https://github.com/Kenix3/libultraship");
+            mManifest.Description = json.value("description", "No description provided.");
+            mManifest.License = json.value("license", "MIT");
```
Kenix3: Definitely default to an empty license (or the all rights reserved text) instead of MIT

src/ship/resource/archive/Archive.cpp
```diff
@@ -90,6 +139,18 @@ uint32_t Archive::GetGameVersion() {
     return mGameVersion;
 }
 
+const ArchiveManifest& Archive::GetManifest() {
```
Kenix3, line 144: Why return a reference here? Do you want to make the private data writable?

src/ship/resource/archive/Archive.cpp
```diff
@@ -115,10 +176,110 @@ void Archive::IndexFile(const std::string& filePath) {
     (*mHashes)[CRC64(filePath.c_str())] = filePath;
 }
 
-std::shared_ptr<File> Archive::LoadFile(uint64_t hash) {
-    const std::string& filePath =
-        *Context::GetInstance()->GetResourceManager()->GetArchiveManager()->HashToString(hash);
-    return LoadFile(filePath);
+void Archive::Validate() {
```
Kenix3, line 181: Why does a validate function have no return value? What happens if this archive is invalid?

src/ship/scripting/ScriptLoader.cpp
```diff
@@ -0,0 +1,184 @@
...
+    int fd = mkstemp(pathTemplate);
+    if (fd == -1) {
+        throw std::runtime_error("Failed to create temporary file: " + std::to_string(errno));
+    }
+
+#ifdef __linux__
+    fchmod(fd, 0755);
+#endif
+
+    // mkstemp creates a physical file on the hard drive.
```
Kenix3: Why are we leaving files on people's hard drives? Where do we clean these up? How do we clean these up on crashes? Generally not a good idea to write files out.

Kenix3 (commented):
Here are some other general things you could point me to in order to make this review go faster:
- Where can I see the concept of exposed functions from a mod? This is necessary for dependency.
- Where can I see how we create and package initial key stores?
- Where do we prompt for the user to accept a key when loading a mod from a new author for the first time?
- Where can I see the cmake changes that allow these code mods to be debugged?

Kenix3 (commented):
As this is a new feature and we have tests now, this should definitely come along with some new tests!

include/ship/resource/archive/Archive.h
KiritoDv: CodeVersion is for when we have breaking source code changes and GameVersion is the equivalent to the version file but inside the json file, which is the version of the rom

src/ship/resource/archive/Archive.cpp
KiritoDv: Its just to avoid copying stuff, since its marked as const it should be read-only

src/ship/resource/archive/Archive.cpp
KiritoDv: The archive is invalid always, Validate changes mIsChecksumValid and mIsSigned to be true

src/ship/scripting/ScriptLoader.cpp
KiritoDv: Not really, for mac and linux it gets deleted as soon as the library is loaded, thats the purpose of the unlink, and for windows the file is deleted when Unload is called

src/ship/resource/archive/Archive.cpp
KiritoDv: Changed to 'All rights reserved'

src/ship/events/EventSystem.cpp
KiritoDv: I like EventSystem but im open to changing it xD

include/ship/scripting/ScriptLoader.h
KiritoDv: Fixed it

include/libultraship/libultra/gbi.h
KiritoDv: Mostly because when i was working on the demo mods, i found f3d had a big bug that was hard to fix related to big textures and vtx idx being bigger than 16

KiritoDv:
@Kenix3 This is where you can see the exported functions 
- Library: https://github.com/Net64DD/Ghostship_LywxUtils
- Mod that uses it: https://github.com/Net64DD/Ghostship_DoomMod

src/ship/scripting/ScriptLoader.cpp
Kenix3: Will Unload be called on crash? If not that should be a part of our crash handler now. Unload and delete all mod libs.

Kenix3:
> @Kenix3 This is where you can see the exported functions
> 
>     * Library: https://github.com/Net64DD/Ghostship_LywxUtils
> 
>     * Mod that uses it: https://github.com/Net64DD/Ghostship_DoomMod

Awesome! Where can I see an example of the headers that are packaged? Where are they packaged to?

.github/workflows/apt-deps.txt
```diff
@@ -1 +1 @@
-libsdl2-dev libpng-dev libglew-dev libzip-dev zipcmp zipmerge ziptool ninja-build nlohmann-json3-dev libtinyxml2-dev libspdlog-dev
\ No newline at end of file
+libsdl2-dev libpng-dev libglew-dev libzip-dev zipcmp zipmerge ziptool ninja-build nlohmann-json3-dev libtinyxml2-dev libspdlog-dev tcc libtcc-dev
```
Kenix3, line 1: Please update the readme with these new dependencies

cmake/dependencies/common.cmake
```diff
@@ -110,4 +110,126 @@ FetchContent_Declare(
     GIT_REPOSITORY https://github.com/KiritoDv/prism-processor.git
     GIT_TAG 1de054450e7b3c5f777d2e3dfcb228ad120c329d
 )
-FetchContent_MakeAvailable(prism)
\ No newline at end of file
+FetchContent_MakeAvailable(prism)
+
+#=========== monocypher ===========
+FetchContent_Declare(
+    monocypher
```
Kenix3, line 117: Another new dependency to add to the readme

include/ship/scripting/ScriptSystem.h
```diff
@@ -0,0 +1,47 @@
...
+#include "ship/scripting/ScriptLoader.h"
+
+#include <unordered_map>
+#include <string>
+#include <optional>
+#include <functional>
+
+namespace Ship {
+
+enum class SafeLevel { ONLY_TRUSTED_SCRIPTS, WARN_UNTRUSTED_SCRIPTS, ALLOW_ALL_SCRIPTS };
```
Kenix3: Can we add one more level? Turn off all scripts.

Kenix3:
> @Kenix3 This is where you can see the exported functions
> 
>     * Library: https://github.com/Net64DD/Ghostship_LywxUtils
> 
>     * Mod that uses it: https://github.com/Net64DD/Ghostship_DoomMod

https://github.com/Net64DD/Ghostship_DoomMod/blob/main/src/objs/tv/logic.c
Does calling functions from another mod rely on this function wrapper, or is it just how you implemented this?

KiritoDv:
> > @Kenix3 This is where you can see the exported functions
> > ```
> > * Library: https://github.com/Net64DD/Ghostship_LywxUtils
> > 
> > * Mod that uses it: https://github.com/Net64DD/Ghostship_DoomMod
> > ```
> 
> https://github.com/Net64DD/Ghostship_DoomMod/blob/main/src/objs/tv/logic.c Does calling functions from another mod rely on this function wrapper, or is it just how you implemented this?

No, i added an easy wrapper but the real helper its on include/mod.h, you can take a look at the mod template i made: https://github.com/KiritoDv/Ghostship/tree/60421be6ef17125a19bf6cb681464f7bb3ec80a6/workspace/template

I also added a rule on the cmake to copy include folders into the zip

Kenix3:
> > > @Kenix3 This is where you can see the exported functions
> > > ```
> > > * Library: https://github.com/Net64DD/Ghostship_LywxUtils
> > > 
> > > * Mod that uses it: https://github.com/Net64DD/Ghostship_DoomMod
> > > ```
> > 
> > 
> > https://github.com/Net64DD/Ghostship_DoomMod/blob/main/src/objs/tv/logic.c Does calling functions from another mod rely on this function wrapper, or is it just how you implemented this?
> 
> No, i added an easy wrapper but the real helper its on include/mod.h, you can take a look at the mod template i made: https://github.com/KiritoDv/Ghostship/tree/60421be6ef17125a19bf6cb681464f7bb3ec80a6/workspace/template
> 
> I also added a rule on the cmake to copy include folders into the zip

Ok so in order to call a function from another mod you do need to use the `CALL_FUNC` macro?

KiritoDv:
> > > > @Kenix3 This is where you can see the exported functions
> > > > ```
> > > > * Library: https://github.com/Net64DD/Ghostship_LywxUtils
> > > > 
> > > > * Mod that uses it: https://github.com/Net64DD/Ghostship_DoomMod
> > > > ```
> > > 
> > > 
> > > https://github.com/Net64DD/Ghostship_DoomMod/blob/main/src/objs/tv/logic.c Does calling functions from another mod rely on this function wrapper, or is it just how you implemented this?
> > 
> > 
> > No, i added an easy wrapper but the real helper its on include/mod.h, you can take a look at the mod template i made: https://github.com/KiritoDv/Ghostship/tree/60421be6ef17125a19bf6cb681464f7bb3ec80a6/workspace/template
> > I also added a rule on the cmake to copy include folders into the zip
> 
> Ok so in order to call a function from another mod you do need to use the `CALL_FUNC` macro?

Yeah because of how the exposed function work, i added the macro, where you specify the mod name and the function to call

Issue #1037, opened 2026-03-30
Compilation fails with new clang version and old spdlog on macOS


koplas:
The compilation fails with the following output:
```
[1/1376] Building CXX object
_deps/spdlog-build/CMakeFiles/spdlog.dir/src/bundled_fmtlib_format.cpp.o
FAILED: [code=1] _deps/spdlog-build/CMakeFiles/spdlog.dir/src/bundled_fmtlib_format.cpp.o
/usr/bin/c++ -DCONTROLLERBUTTONS_T=uint32_t -DCVAR_ALLOW_BACKGROUND_INPUTS=\"gAllowBackgroundInputs\"
-DCVAR_AUDIO_CHANNELS_SETTING=\"gAudioChannelsSetting\" -DCVAR_CONSOLE_WINDOW_OPEN=\"gOpenWindows.Console\"
-DCVAR_CONTROLLER_CONFIGURATION_WINDOW_OPEN=\"gControllerConfigurationEnabled\"
-DCVAR_CONTROLLER_DISCONNECTED_WINDOW_OPEN=\"gOpenWindows.ControllerDisconnected\"
-DCVAR_CONTROLLER_REORDERING_WINDOW_OPEN=\"gOpenWindows.ControllerReordering\"
-DCVAR_ENABLE_MULTI_VIEWPORTS=\"gSettings.EnableMultiViewports\" -DCVAR_GAME_OVERLAY_FONT=\"gSettings.OverlayFont\"
-DCVAR_GFX_DEBUGGER_WINDOW_OPEN=\"gOpenWindows.GfxDebugger\" -DCVAR_IMGUI_CONTROLLER_NAV=\"gSettings.ControlNav\"
-DCVAR_INTERNAL_RESOLUTION=\"gSettings.InternalResolution\" -DCVAR_LOW_RES_MODE=\"gSettings.LowResMode\"
-DCVAR_MENU_BAR_OPEN=\"gSettings.OpenMenuBar\" -DCVAR_MSAA_VALUE=\"gSettings.MSAAValue\"
-DCVAR_PREFIX_ADVANCED_RESOLUTION=\"gSettings.AdvancedResolution\" -DCVAR_PREFIX_AUDIO=\"gAudioEditor\"
-DCVAR_PREFIX_CHEAT=\"gCheats\" -DCVAR_PREFIX_CONTROLLERS=\"gSettings.Controllers\" -DCVAR_PREFIX_COSMETIC=\"gCosmetics\"
[...]

PR #1060, opened 2026-04-12
Mac: Update spdlog

ShishuTheDragon:
[mac.cmake](https://github.com/Kenix3/libultraship/blob/46bdbdd143ca9a1fe23226d1df8027083b664e05/cmake/dependencies/mac.cmake#L3) currently pulls in **spdlog 1.14.1** which causes build issues on the latest version of macOS / Xcode / Clang.

Looks like [ios.cmake](https://github.com/Kenix3/libultraship/blob/46bdbdd143ca9a1fe23226d1df8027083b664e05/cmake/dependencies/ios.cmake#L40) is pulling in 1.16.0, so I upgraded the Mac one to match.

### When did this break?
It may have started breaking in the latest Xcode, version 26.4. That’s when Apple upgraded Clang from 17 to 21.

### Is this backwards compatible?
I believe so. Shipwright still builds after making this change (see https://github.com/HarbourMasters/Shipwright/pull/6491), and those workflows run on macos-14-arm64. From the SoH build log, it looks like they are using Clang 15.

### DIFF
--- a/cmake/dependencies/mac.cmake
+++ b/cmake/dependencies/mac.cmake
@@ -8,7 +8,7 @@ if (NOT ${spdlog_FOUND})
     FetchContent_Declare(
         spdlog
         GIT_REPOSITORY https://github.com/gabime/spdlog.git
-        GIT_TAG v1.14.1
+        GIT_TAG v1.16.0
         OVERRIDE_FIND_PACKAGE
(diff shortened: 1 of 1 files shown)


koplas:
Closes #1037

PR #1069, opened 2026-04-15
Implement async texture loading with deferred imports

koplas:
Replace synchronous LoadResourceProcess calls with async LoadResourceAsync throughout the Fast3D interpreter's texture pipeline (SetTImg, LoadBlock, LoadTile, LoadTLUT, s2dex bg ops).

When a texture resource isn't ready at import time, the import is deferred via PendingImport/PendingLoadTlut structs that snapshot the full loaded_texture and texture_tile state. On Flush(), the snapshotted state is temporarily restored so Import sub-functions read correct data, then swapped back afterward.

Key changes:
- RawTexMetadata.resource is now shared_future<IResource> instead of shared_ptr<Texture>; path/load_params stored directly in metadata
- ApplyLoadScale() extracted from GfxDpLoadBlock/GfxDpLoadTile to compute size_bytes/line_size_bytes/load_offset from load_params
- ProcessImportInternal() accepts optional texture/tile snapshots and saves/restores live state around sub-function calls
- CI textures deferred when TLUTs are still pending
- ImportTextureRaw fallthrough documented as intentional raw-data fallback when the resource isn't ready yet

This was tested with the Shipwright port and an 4k texture pack.

Performance improvements depend on the amount of cores and the resources that are loaded concurrently. In my test setup speed improvement was around 4x - 5x. Stuttering still occurs especially with high frame rates e.g. 120 fps, but is considerably reduced.

The changes are more extensive than anticipated. However, I hope they don't overwhelm anyone :).

### DIFF
--- a/include/fast/interpreter.h
+++ b/include/fast/interpreter.h
@@ -10,6 +10,7 @@
 #include <stack>
 #include <string>
 #include <string_view>
+#include <future>
 
 #include "fast/lus_gbi.h"
 #include "fast/types.h"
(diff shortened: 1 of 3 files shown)


koplas:
I've rebased the commit and will be testing it again. Other testers experiencing HD/4K texture stuttering are also welcome to try this build and share their feedback.

koplas:
Shipwright doesn't support the new version of libultraship with SDL3. Converting to a draft until I can test it again.

PR #1081, opened 2026-04-19
Re-introduce GfxDpSetBlendColor

bass3l:
This was previously introduced in PR #1021 but was reverted in #1046. This is tested in Paperboat and another WIP port and is working fine.


bass3l:
cc @JeodC

JeodC:
#1046 had `Remove broken use_blend_color fog path — standard fog (G_BL_CLR_FOG) is the only blender source that works with the current shader setup`, which turned out to be erroneous simply because the path was unused in the port being tested for the request. As a result, it was mistakenly assumed to be a cause of a bug and removed. Adding the path back was re-tested and was found to be sane.

Issue #1074, opened 2026-04-16
New tcc additions break MSVC compiling


JeodC:
Notably, the new tcc additions:

- Have build artifacts not in `.gitignore`
- Are not copied to Debug build folders for testing (attempting to run complains it's missing tcc.dll)
- Will not compile in MSVC due to an `__assert` overlap which has to be worked around by setting `NDEBUG` for tcc:
```cmake
    if(MSVC)
        if(CMAKE_GENERATOR_PLATFORM MATCHES "ARM64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "ARM64|aarch64")
            target_compile_definitions(libtcc1 PRIVATE __aarch64__ _WIN64)
            target_compile_definitions(libtcc  PRIVATE __aarch64__ TCC_TARGET_ARM64 _WIN64)
        else()
            target_compile_definitions(libtcc1 PRIVATE __x86_64__ _WIN64)
            target_compile_definitions(libtcc  PRIVATE __x86_64__ TCC_TARGET_X86_64 _WIN64)
        endif()
        target_compile_definitions(libtcc1 PRIVATE "__faststorefence=__faststorefence_tcc_unused")
        target_compile_definitions(libtcc PRIVATE NDEBUG) # New
    endif()
```

Shouldn't this system be opt-in rather than opt-out? Ports not yet using the scripting system are failing on the above, I don't think modifying CMakeLists per port should be necessary at this point.

KiritoDv:
Just to put what i said on the discord, here, the cmake changes are not needed in any way for the ports to build fine, you can see the CI builds for this two ports working fine https://github.com/HarbourMasters/Starship/actions/runs/24494256083/job/71585564483?pr=255
https://github.com/HarbourMasters/Ghostship/actions/runs/24491921177/job/71578506145

Among the main builds from LUS
win-x64 https://github.com/Kenix3/libultraship/actions/runs/24485084123/job/71558004478
win-arm https://github.com/Kenix3/libultraship/actions/runs/24485084123/job/71558004456

JeodC:
> [build] [18/70   1% :: 0.458] Linking C shared library D:\GitHub\SourceProjects\SomePort\libultraship\tcc.dll
> [build] FAILED: [code=4294967295] D:/GitHub/SourceProjects/SomePort/libultraship/tcc.dll D:/GitHub/SourceProjects/SomePort/libultraship/tcc.lib 
[...]

PR #1084, opened 2026-04-20
Flip tcc scripting to opt-in

JeodC:
Addresses #1074, tested building with a port that is not using this new system yet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -348,6 +348,7 @@ ZAPDUtils.lib
 cmake-build-**
 cmake_install.cmake
 tcc_c2str
+tcc_c2str.exe
 *.a
 *.o
 *.so
@@ -358,6 +359,8 @@ tcc_c2str
 *.stamp
 *.depend
 *.lib
+*.exp
+*.dll
 *.dylib
 *.tlog
 *.recipe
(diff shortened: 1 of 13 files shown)


JeodC:
CC @KiritoDv

KiritoDv:
Libtcc can't be static

JeodC:
> Libtcc can't be static

It can't? How come I only get the missing tcc.dll error in Debug builds but not in Release builds?

KiritoDv:
> > Libtcc can't be static
> 
> 
> 
> It can't? How come I only get the missing tcc.dll error in Debug builds but not in Release builds?

Its probably a setup issue but it can't because of the license which is lgpl

JeodC:
Ok, amended. Thanks. I am unsure of my changes re: Debug vs Release `tcc.dll` missing. Before this pull request, when I was testing 2ship, I compiled a Release build which when I ran the exe directly, it worked fine and there was no `tcc.dll` present. When I compiled a different port in VSCode and ran the Debug build from the VSCode integrated cmake run button, it threw the `tcc.dll missing` error.

KiritoDv:
Awesome, just one last thing, can you try it vs https://github.com/KiritoDv/Ghostship/tree/impl/c-mods ?

JeodC:
> Awesome, just one last thing, can you try it vs https://github.com/KiritoDv/Ghostship/tree/impl/c-mods ?

Sorry I haven't found time to test this yet.

PR #1089, opened 2026-04-26
decode G_SETKEYR/G_SETKEYGB/G_SETCONVERT into RDP state and route CENTER/SCALE/K4/K5 to dedicated combiner inputs

bass3l:
- Before: CENTER/SCALE aliased to G_CCMUX_1 / G_CCMUX_NOISE
- After:  CENTER/SCALE are routed to dedicated combiner inputs and the shader's standard (A - B) * C + D path produces (tex - center) * scale.


Kenix3:
Try merging in latest main to this branch and see if it fixes CI.

Issue #1072, opened 2026-04-15
#1046 breaks depth in MM


louist103:
<img width="1376" height="791" alt="Image" src="https://github.com/user-attachments/assets/b86e3996-7184-46ab-b519-6d8bb367369d" />

briaguya0:
just adding a link https://github.com/Kenix3/libultraship/pull/1046

JeodC:
Caused by
```cpp
    // N64 coverage simulation: When ALPHA_CVG_SEL is set, the RDP blender uses
    // coverage as its alpha input instead of the combiner output. On N64, coverage
    // for textured pixels is derived from texture alpha — zero-alpha texels produce
    // zero coverage and are never written, regardless of OPA/XLU blender mode.
    if (!use_alpha && (mRdp->other_mode_l & ALPHA_CVG_SEL)) {
        use_alpha = true;
        alpha_threshold = true;
    }
```

Will see if I can find a solution.

JeodC:
@louist103 This should be resolved with the above pr merge, can you confirm? It was working when I tested it.

louist103:
> [@louist103](https://github.com/louist103) This should be resolved with the above pr merge, can you confirm? It was working when I tested it.

Looks like this is fixed but the title screen still looks off. 

<img width="949" height="562" alt="Image" src="https://github.com/user-attachments/assets/557191c6-c8ff-4fd1-8f5e-e43d364b13d7" />

<img width="1098" height="610" alt="Image" src="https://github.com/user-attachments/assets/c33d22ac-e959-46f9-b400-3ad5d4e42cfd" />

JeodC:
> > [@louist103](https://github.com/louist103) This should be resolved with the above pr merge, can you confirm? It was working when I tested it.
> 
> Looks like this is fixed but the title screen still looks off. 
> 
> <img width="949" height="562" alt="Image" src="https://github.com/user-attachments/assets/557191c6-c8ff-4fd1-8f5e-e43d364b13d7" />
> 
> <img width="1098" height="610" alt="Image" src="https://github.com/user-attachments/assets/c33d22ac-e959-46f9-b400-3ad5d4e42cfd" />

Alright, more digging to do. Thanks, will take a look this week.

JeodC:
@louist103 This issue should be resolved as of #1090

PR #1090, opened 2026-04-26
Apply tile-size clamps only for mipmap pyramids and CLAMP tiles

JeodC:
PR #1043 added two unconditional clamps that overwrote texture dimensions with the rendered window from `SetTileSize `whenever the latter was smaller:

- `ImportTextureXxx `(Rgba16, Rgba32, Ci4, Ci8): clamps the GPU upload size
- `GfxSpTri1`: clamps tex_width/tex_height used for UV [0,1] normalization.

The intent was to handle mipmap pyramid loads, where LoadBlock deposits a base mip plus smaller mips into a single TMEM region. Without clamping, the inflated loaded_size / line_size overshoots tex_height, and UVs sample past the base mip into the smaller-mip bytes.

But the clamps fire for any draw where (lrs - uls + 4) / 4 < tex_width, including:

- Scrolling tiles that use gDPSetTileSize to offset UV sampling into a tiling texture
- Window-clamped sub-regions rendered with gDPSetTileSizeInterp's float values
- Animated water/cloud effects that change uls/ult per frame
- 
In all those cases the clamp shrinks tex_width (and the ImportTexture upload size) to the visible window, so WRAP mode tiles the wrong amount of texture and the visual changes per frame as the scroll values animate.

This change replaces the unconditional clamps with a discriminator. The mipmap-pyramid signature is loaded_pixels slightly larger than rendered_pixels (≈1.33× because mips sum to 4/3 of the base). Window-scrolling tiles have loaded_pixels ≈ rendered_pixels or loaded_pixels >> rendered_pixels. We clamp when:

- 1.0 < loaded_pixels / rendered_pixels < ~1.625 (pyramid-like ratio), **or**
- The tile's wrap mode has G_TX_CLAMP set (the unambiguous "render only this sub-region" signal)

### DIFF
--- a/src/fast/interpreter.cpp
+++ b/src/fast/interpreter.cpp
@@ -547,13 +547,21 @@ void Interpreter::ImportTextureRgba16(int tile, bool importReplacement) {
     uint32_t width = widthBytes / 2;
     uint32_t height = widthBytes > 0 ? sizeBytes / widthBytes : 0;
 
-    // Clamp to tile dimensions from SetTileSize (mipmap pyramids include all levels).
+    // Clamp to the rendered region only when the loaded buffer is ~1.33x of it (mipmap
+    // pyramid signature). Window-scrolling tiles have loaded ≈ rendered or loaded >> rendered;
+    // skip both. CLAMP wrap mode always opts in.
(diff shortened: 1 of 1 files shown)


JeodC:
Follow up for #1072 and unreported issue with Starship's water tiles in Corneria.

PR #1091, opened 2026-04-26
Fix G_EX_INVERT_CULLING to apply to all microcode types

PurpleHato:
G_EX_INVERT_CULLING is a LUS extension, not an N64 hardware feature, so it should work regardless of which microcode handler is active. Previously it was gated behind an F3DEX2-only check, preventing games using F3DEX (or other ucode variants) from using inverted culling. (usefull for modding or enenahcenemnts)


Kenix3:
Try merging in latest main to this branch and see if it fixes CI.

PR #1096, opened 2026-05-02
Replace mod() with clamp() in shader WRAP macro

JeodC:
Ship of Harkinian renders textures as missing on devices using PowerVR Rogue GPUs, others untested. Geometry is visible but many textured surfaces are transparent. Same binary works fine on Mali, Adreno, and desktop drivers.

I tracked it down to the fragment shader's WRAP macro:
`#define WRAP(x, low, high) mod((x)-(low), (high)-(low)) + (low)`

On Rogue, `mod()` here returns values outside `[low, high]`. The downstream `clamp(texel, 0.0, 1.0)` then snaps those to `0`. With `texel.a == 0` and `glBlendFunc(GL_SRC_ALPHA, ...)`, pixels contribute nothing to the framebuffer.

The Khronos GLSL spec [explicitly warns](https://registry.khronos.org/OpenGL-Refpages/gl4/html/mod.xhtml) that implementations may compute `mod()` imprecisely: "the error can be large due to the discontinuity in floor... can also cause the result to have a different sign than the infinitely precise result." Similar `mod()` bugs are documented on [Raspberry Pi VideoCore](https://forums.raspberrypi.com/viewtopic.php?t=189455), [Mesa swrast](https://bugs.freedesktop.org/show_bug.cgi?id=27286), and [Metal via Naga](https://github.com/gfx-rs/naga/issues/802).

Replace mod with clamp. It's strictly different math, but functionally equivalent for Harkinian's combiner output range (which as far as I can tell never goes far enough out to need wrap-around) and a subsequent clamp(0, 1) erases any remaining difference at output.

Original wrap:

image</br>

image</br>

After change:

image</br>

If there is a regression due to this change, it will most likely be visible in terrain decals, fog, or particle effects. I would greatly appreciate spot-check tests before merging, and the shader file doesn't require recompiling anything so anyone can test by replacing the shader file in the port o2r.

### DIFF
--- a/src/fast/shaders/opengl/default.shader.glsl
+++ b/src/fast/shaders/opengl/default.shader.glsl
@@ -130,7 +130,7 @@
     uniform int texture_filtering[2];
 
     #define TEX_OFFSET(off) @{texture}(tex, texCoord - off / texSize)
-    #define WRAP(x, low, high) mod((x)-(low), (high)-(low)) + (low)
+    #define WRAP(x, low, high) clamp((x), (low), (high))
 
     float random(in vec3 value) {
         float random = dot(sin(value), vec3(12.9898, 78.233, 37.719));


JeodC:
image

No noticeable regression on Windows OpenGL

JeodC:
I am not seeing any regressions in 2ship from this either, tested on my AYN Thor (opengles, linux arm64).

PR #1103, opened 2026-05-17
Untangle various parts from the Context destructor

### LINKED ISSUE #1102: Context destructor is not safe
The `Context` destructor causes several memory issues. At best its UB, and worst and usually it crashes. The ref count of `Context` and `Window` are set to zero before everything is done accessing them, causing the shared_ptr to return `nullptr` during cleanup.

louist103:
Various components were trying to access other components in `Context` during their destructors. `shared_ptr` would return `nullptr` leading to crashes.
This also fixes a race condition and double free inside `spdlog` related to shutting it down.
Fixes #1102. I've only tested on Linux. Someone will need to test Windows and MacOS

### DIFF
--- a/include/fast/Fast3dGui.h
+++ b/include/fast/Fast3dGui.h
@@ -42,6 +42,7 @@ typedef struct {
             uint32_t Height; ///< Framebuffer height in pixels.
         } Gx2;
     };
+    int32_t Backend;
 } GuiWindowInitData;
 
 /**
@@ -152,8 +153,8 @@ class Fast3dGui : public Ship::Gui {
      */
     ImTextureID GetTextureById(int32_t id);
 
-    std::weak_ptr<Interpreter> mInterpreter; ///< Weak reference to the Fast3D scripting interpreter.
     GuiWindowInitData mImpl;                 ///< Backend-specific window/context handles passed to Init().
+    std::weak_ptr<Interpreter> mInterpreter; ///< Weak reference to the Fast3D scripting interpreter.
 
   private:
     /** @brief Applies any pending resolution or MSAA changes to the render target. */
@@ -163,8 +164,7 @@ class Fast3dGui : public Ship::Gui {
      * @brief Returns the integer scaling factor applied to the game viewport.
      * @return Scaling multiplier (1 = native, 2 = 2×, etc.).
      */
-    int16_t GetIntegerScaleFactor();
-
     std::unordered_map<std::string, Ship::GuiTextureMetadata> mGuiTextures; ///< Cached GPU texture registry.
+    int16_t GetIntegerScaleFactor();
 };
 } // namespace Fast
--- a/include/fast/Fast3dGui.h
+++ b/include/fast/Fast3dGui.h
@@ -1,5 +1,7 @@
 #pragma once
 #include <SDL2/SDL.h>
+
+#include "Fast3dWindow.h"
 #include "ship/window/gui/Gui.h"
 #include "fast/WindowEvent.h"
 #include "fast/resource/type/Texture.h"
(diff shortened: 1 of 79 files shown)


include/fast/Fast3dGui.h
```diff
@@ -163,8 +164,7 @@ class Fast3dGui : public Ship::Gui {
      * @brief Returns the integer scaling factor applied to the game viewport.
      * @return Scaling multiplier (1 = native, 2 = 2×, etc.).
      */
-    int16_t GetIntegerScaleFactor();
-
     std::unordered_map<std::string, Ship::GuiTextureMetadata> mGuiTextures; ///< Cached GPU texture registry.
+    int16_t GetIntegerScaleFactor();
```
Pepe20129, line 169: This change separates the function from its documentation.

Pepe20129 (approved):
One small thing.

include/fast/Fast3dGui.h
```diff
@@ -42,6 +42,7 @@ typedef struct {
             uint32_t Height; ///< Framebuffer height in pixels.
         } Gx2;
     };
+    int32_t Backend;
```
briaguya0: why is this an int instead of using the `WindowBackend` enum?

include/fast/Fast3dGui.h
```diff
@@ -152,8 +153,8 @@ class Fast3dGui : public Ship::Gui {
      */
     ImTextureID GetTextureById(int32_t id);
 
-    std::weak_ptr<Interpreter> mInterpreter; ///< Weak reference to the Fast3D scripting interpreter.
     GuiWindowInitData mImpl;                 ///< Backend-specific window/context handles passed to Init().
+    std::weak_ptr<Interpreter> mInterpreter; ///< Weak reference to the Fast3D scripting interpreter.
```
briaguya0, line 158: looking for confirmation: i assume this reordering is to address destruction order issues and not just noise?

include/libultraship/bridge/resourcebridge.h
```diff
@@ -1,10 +1,10 @@
 #pragma once
 
 #include "stdint.h"
+#include "stddef.h"
 #include "ship/Api.h"
 
 #ifdef __cplusplus
-#include "fast/resource/type/Texture.h"
```
briaguya0, line 7: this isn't part of the fix is it? just unrelated cleanup?

include/ship/Context.h
```diff
@@ -40,15 +40,18 @@ class Keystore;
...
   public:
     /**
      * @brief Returns the currently active global Context instance.
-     * @return Shared pointer to the Context, or an empty pointer if none exists.
+     * @return Raw pointer to the Context. This should not be considered shared and should be considered invalid
+     * once the scope ends.
      */
-    static std::shared_ptr<Context> GetInstance();
-
+    static Context* GetRawInstance();
```
briaguya0, line 53: want to hear @Kenix3's thoughts on the rename here. if we didn't rename and kept `GetInstance` then ports that assume they're getting a shared pointer from `GetInstance` could possibly silently be wrong about that (but i can't think of an actual way a port uses `GetInstance` that would break with a raw pointer vs shared)

i'd personally prefer not to do the rename, we know it's a raw pointer because of the return type so adding `Raw` to the name feels redundant, and it also means a ton of lines need to be changed just to support the rename

include/libultraship/bridge/resourcebridge.h
louist103: It is unrelated. It was taking too long to rebuild so I needed to fix some header I can revert.

serprex (approved):

include/ship/Context.h
Kenix3: Name is fine.

PR #1105, opened 2026-05-18
feature(audio): add native PipeWire backend for Linux

bassdr:
PipeWireAudioPlayer replaces the SDL fallback on Linux with a native PipeWire playback stream, matching the architecture of the existing CoreAudio and WASAPI backends.

  - Ring buffer bridges the push model (DoPlay, main thread) and the pull model (on_process RT callback, PipeWire thread)
  - Underrun recovery: last chunk replayed with linear fade-out instead of silence click
  - pw_thread_loop provides the RT thread; no manual thread management
  - Config: pulse backend migrated to PIPEWIRE on Linux, pipewire added as a serialized backend name
  - AudioBackend enum gains PIPEWIRE value


KiritoDv:
Something funny, we had pipewire before but it got removed

briaguya0:
> Something funny, we had pipewire before but it got removed

we had pulseaudio, not pipewire (https://github.com/Kenix3/libultraship/pull/356)

that being said, for the pipewire part of this PR to land there'd need to be good justification for adding another backend to maintain. the PR description doesn't provide any reasoning for adding the new backend.

briaguya0:
if the reasoning is "SDL isn't using pipewire" then i'd much rather just expose the "tell SDL to use pipewire" setting https://wiki.libsdl.org/SDL2/FAQUsingSDL#linux_1

bassdr:
The reason would be the same as why we are not using the SDL backend either on windows and mac, with their dedicated audio backends (Wasapi and CoreAudio) that interract directly with the low level audio system on the platform. The pipewire backend is configuring the ring buffer size, it's using the pw_thread_loop thread instead of going though an intermediate layer, etc.

The first commit is fixing audio issues I had using SDL backend (I started by fixing the SDL backend first) and it should improve things for all platforms. I'm OK to improve further the SDL backend, just that I felt it's easier to just create a pipewire backend that "speaks" directly to the pipewire layer.

As for pulse, I agree with you. I don't think having a separate backend for this one worth it today, people are migrating to pipewire, and others that don't will be fine with the SDL backend.

Should I create separate PRs for things that are not related to Pipewire?

bassdr:
> if the reasoning is "SDL isn't using pipewire" then i'd much rather just expose the "tell SDL to use pipewire" setting https://wiki.libsdl.org/SDL2/FAQUsingSDL#linux_1

No, this is not the reason, my SDL is configured properly to use the pipewire backend. The SDL backend "works" for me, just had audio glitches sometimes and it was just easier to create a new backend after fixing things that were obvious.

briaguya0:
> Should I create separate PRs for things that are not related to Pipewire?

splitting this up would be good, yeah. it seems like quite a bit of this PR is `fix:` but adding the pipewire backend is solidly `feat:`, so reviewing/landing those separately would be good

bassdr:
> splitting this up would be good, yeah

Done, they are #1106 and #1107 , this one now only contains the new pipewire audio backend

bassdr:
Closing in favor of #1125

PR #1106, opened 2026-05-19
feature(audio): resample N64 audio to native 48 kHz output

bassdr:
Shipwright audio engine produces PCM at 32 kHz — the console's native sample rate. Most modern audio devices run at 48 kHz. Without explicit resampling, the OS or driver performs an implicit conversion of unknown quality, often introducing aliasing artifacts.
This PR adds a clean resampling step inside AudioPlayer::Play(), between the game's audio engine and the backend (SDL, PipeWire, WASAPI, CoreAudio). The resampler is transparent to all backends — they simply open the device at 48 kHz as they would normally.
The 32 kHz → 48 kHz ratio is exactly 3/2, which allows a compact polyphase filter (8 taps per phase) with good stopband rejection (~60 dB). The output buffer is stack-allocated to avoid any heap allocation on the audio hot path.

- AudioSettings gains SourceSampleRate (default 0 = passthrough)
- AudioPlayer::Play() resamples transparently before DoPlay() when SourceSampleRate != SampleRate
- GetDesiredBuffered() scales from source rate to output rate so OTRAudio_Thread fill logic remains coherent
- Resample output uses a fixed std::array<int16_t, 16384> — no heap allocation on the audio hot path
- Default SampleRate changed from 44100 to 48000 Hz
- AudioPlayer destructor made virtual to fix UB in derived class dtors

### DIFF
--- a/include/ship/audio/AudioPlayer.h
+++ b/include/ship/audio/AudioPlayer.h
@@ -4,6 +4,7 @@
 #include <string>
 #include <memory>
 #include "ship/audio/AudioChannelsSetting.h"
+#include "ship/audio/AudioResampler.h"
 #include "ship/audio/SoundMatrixDecoder.h"
 
 namespace Ship {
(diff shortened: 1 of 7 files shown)


bassdr:

PR #1107, opened 2026-05-19
fix(audio): stop the SDL backend dropping audio when the queue runs ahead

bassdr:
`DoPlay()` dropped its buffer outright once `Buffered()` reached 6000 frames. A slow frame then became an audible cut instead of extra latency, which is the wrong trade for the situation the threshold was meant to protect against.

The threshold was arbitrary, and the callers already bound how much they hand over - every port checks the buffered amount before calling `AudioPlayer_Play`, as @briaguya0 catalogued in the thread below (soh, 2ship, Starship, SpaghettiKart, Ghostship). SDL's own queue handles backpressure from there.

### Rebased onto the SDL3 migration

This PR predates #1191, and most of what it carried is now moot:

| original change | status |
| --- | --- |
| `SDL_Init` → `SDL_InitSubSystem` | already in main |
| `NULL` → `nullptr`, style, log format | already in main |
| read the channel count back from `have` | not applicable - SDL3 converts from the app-side spec, so the requested count is the right divisor in `Buffered()` |
| double the device buffer (`want.samples * 2`) | dropped - SDL3 has no per-device sample count in `SDL_AudioSpec`. `SDL_HINT_AUDIO_DEVICE_SAMPLE_FRAMES` is the equivalent knob, and SDL3's own guidance is to let it pick |

What is left is the one substantive fix, three lines. Happy to add the sample-frames hint in a follow-up if the underruns it was papering over turn out to still be there under SDL3 - worth measuring rather than assuming, since the stream model is not the SDL2 push queue this was tuned against.

### DIFF
--- a/src/ship/audio/SDLAudioPlayer.cpp
+++ b/src/ship/audio/SDLAudioPlayer.cpp
@@ -47,9 +47,6 @@ int SDLAudioPlayer::Buffered() {
 }
 
 void SDLAudioPlayer::DoPlay(const uint8_t* buf, size_t len) {
-    if (Buffered() < 6000) {
-        // Don't fill the audio buffer too much in case this happens
-        SDL_PutAudioStreamData(mStream, buf, static_cast<int>(len));
-    }
+    SDL_PutAudioStreamData(mStream, buf, static_cast<int>(len));
 }
 } // namespace Ship
--- a/src/ship/audio/SDLAudioPlayer.cpp
+++ b/src/ship/audio/SDLAudioPlayer.cpp
@@ -13,7 +13,8 @@ void SDLAudioPlayer::DoClose() {
     if (mDevice != 0) {
         // Pause playback first
         SDL_PauseAudioDevice(mDevice, 1);
-        // Clear any queued audio to prevent glitches when reopening
+
+        // Prevent stale audio when reopening
         SDL_ClearQueuedAudio(mDevice);
         SDL_CloseAudioDevice(mDevice);
         mDevice = 0;
@@ -21,29 +22,38 @@ void SDLAudioPlayer::DoClose() {
 }
 
 bool SDLAudioPlayer::DoInit() {
-    if (SDL_Init(SDL_INIT_AUDIO) != 0) {
+    if (SDL_InitSubSystem(SDL_INIT_AUDIO) != 0) {
         SPDLOG_ERROR("SDL init error: {}", SDL_GetError());
         return false;
     }
 
-    // Always open with the correct number of output channels
-    mNumChannels = this->GetNumOutputChannels();
-
     SDL_AudioSpec want, have;
     SDL_zero(want);
+
     want.freq = this->GetSampleRate();
     want.format = AUDIO_S16SYS;
-    want.channels = mNumChannels;
-    want.samples = this->GetSampleLength();
-    want.callback = NULL;
+    want.channels = this->GetNumOutputChannels();
 
-    mDevice = SDL_OpenAudioDevice(NULL, 0, &want, &have, 0);
+    // Increase backend/device buffering slightly
+    // to reduce rare underruns/glitches.
+    //
+    // IMPORTANT:
+    // This is NOT part of the emulator buffering logic.
+    // It only affects SDL/backend latency.
+    want.samples = GetSampleLength() * 2;
+
+    want.callback = nullptr;
+
+    mDevice = SDL_OpenAudioDevice(nullptr, 0, &want, &have, 0);
+
     if (mDevice == 0) {
-        SPDLOG_ERROR("SDL_OpenAudio error: {}", SDL_GetError());
+        SPDLOG_ERROR("SDL_OpenAudioDevice error: {}", SDL_GetError());
         return false;
     }
 
-    SPDLOG_INFO("SDL Audio initialized: {} channels, {} Hz", mNumChannels, this->GetSampleRate());
+    mNumChannels = have.channels;
+
+    SPDLOG_INFO("SDL audio initialized: {} Hz, {} channels, {} samples", have.freq, have.channels, have.samples);
 
     SDL_PauseAudioDevice(mDevice, 0);
     return true;
@@ -54,9 +64,6 @@ int SDLAudioPlayer::Buffered() {
 }
 
 void SDLAudioPlayer::DoPlay(const uint8_t* buf, size_t len) {
-    if (Buffered() < 6000) {
-        // Don't fill the audio buffer too much in case this happens
-        SDL_QueueAudio(mDevice, buf, len);
-    }
+    SDL_QueueAudio(mDevice, buf, len);
 }
 } // namespace Ship
--- a/include/ship/audio/SDLAudioPlayer.h
+++ b/include/ship/audio/SDLAudioPlayer.h
@@ -50,6 +50,6 @@ class SDLAudioPlayer final : public AudioPlayer {
 
   private:
     SDL_AudioDeviceID mDevice = 0; ///< Handle to the opened SDL audio device.
-    int32_t mNumChannels = 2;      ///< Number of output channels (2 for stereo, 6 for 5.1).
+    uint8_t mNumChannels = 2;      ///< Number of output channels (2 for stereo, 6 for 5.1).
 };
 } // namespace Ship


src/ship/audio/SDLAudioPlayer.cpp
```diff
@@ -13,37 +13,47 @@ void SDLAudioPlayer::DoClose() {
...
+        // Prevent stale audio when reopening
         SDL_ClearQueuedAudio(mDevice);
         SDL_CloseAudioDevice(mDevice);
         mDevice = 0;
     }
 }
 
 bool SDLAudioPlayer::DoInit() {
-    if (SDL_Init(SDL_INIT_AUDIO) != 0) {
+    if (SDL_InitSubSystem(SDL_INIT_AUDIO) != 0) {
```
briaguya0: this is not actually changing functionality, but is still an improvement

from https://wiki.libsdl.org/SDL2/SDL_Init#remarks
> [SDL_Init](https://wiki.libsdl.org/SDL2/SDL_Init)() simply forwards to calling [SDL_InitSubSystem](https://wiki.libsdl.org/SDL2/SDL_InitSubSystem)(). Therefore, the two may be used interchangeably. Though for readability of your code [SDL_InitSubSystem](https://wiki.libsdl.org/SDL2/SDL_InitSubSystem)() might be preferred.

src/ship/audio/SDLAudioPlayer.cpp
```diff
@@ -54,9 +64,6 @@ int SDLAudioPlayer::Buffered() {
 }
 
 void SDLAudioPlayer::DoPlay(const uint8_t* buf, size_t len) {
-    if (Buffered() < 6000) {
```
briaguya0, line 50: this appears to have existed since soh started (https://github.com/HarbourMasters/Shipwright/commit/d24c8453db1035f382e1b0853be00ebd281bbbdd), would be good to get some context on the reasoning behind having it in the first place. i agree checking against a magic number seems bad, but knowing *why* the check was happening in the first place would be nice.

src/ship/audio/SDLAudioPlayer.cpp
```diff
@@ -13,37 +13,47 @@ void SDLAudioPlayer::DoClose() {
...
-    want.callback = NULL;
+    want.channels = this->GetNumOutputChannels();
+
+    // Increase backend/device buffering slightly
+    // to reduce rare underruns/glitches.
+    //
+    // IMPORTANT:
+    // This is NOT part of the emulator buffering logic.
+    // It only affects SDL/backend latency.
+    want.samples = GetSampleLength() * 2;
```
briaguya0: curious about where this doubling is coming from. is this a recommended fix somewhere?

bassdr: having it == to the SampleLength is not giving any chance in cas there is a CPU load. If we increase it, it still has to be a power of 2, so we can increase it less than that. The more we increase the samples, the bigger is the latency we introduce, but it's not really perceptible with `* 2`. With `* 4`, it works too, but you start to hear it.
 - `* 1` → 1024 (~32 ms)
 - `* 2` → 2048 (~64 ms)
 - `* 4` → 4096 (~128 ms)

BTW this is the main reason why I decided to create a pipewire backwend in #1105, other backends does not have this issue as a callback is called when it needs more data instead of us filling a constant buffer.

src/ship/audio/SDLAudioPlayer.cpp
bassdr: The Buffered() < 6000 guard in DoPlay was the original overflow protection — it prevented SDL from accumulating an unbounded queue when the audio thread produced faster than the device consumed.

The magic number 6000 was likely chosen empirically to represent "enough buffer without too much latency," but it was never tied to DesiredBuffered or the actual sample rate.

The problem is that this guard lives at the wrong level. DoPlay is a backend primitive — it should not know anything about the producer's pacing logic. Overflow protection belongs in the producer, which has full context: it knows DesiredBuffered, the current fill level, and can make a meaningful decision about whether to advance the audio engine at all.

Also, the issue is that this if will suddenly stop playing audio when it happens, so we hear a clear cut during heavy CPU loads. We could look into decoupling this logic so we suffer less when CPU struggles, but this is for later.

Looking at the commit history, when d24c845 was written, OTRAudio_Thread had no fill-level awareness — it just called AudioPlayer_Play unconditionally every frame. The < 6000 guard in DoPlay was the only backstop against overflow. Today OTRAudio_Thread has the do/while loop with AudioPlayer_GetDesiredBuffered() which is the correct place for this logic.
Removing the guard from DoPlay is safe for Shipwright because OTRAudio_Thread now handles it properly. You're right that other libultraship consumers might not have equivalent producer logic — which is worth noting in the PR: removing this from SDLAudioPlayer is intentional but requires the caller to implement its own fill-level guard, as Shipwright does.

src/ship/audio/SDLAudioPlayer.cpp
bassdr: Agreed, I wrongly assumed SDL_InitSubSystem was only initializing the SubSystem (while SDL_Init would have initalize other stuff we don't need. I was wrong, thanks for pointing it out. TIL

src/ship/audio/SDLAudioPlayer.cpp
briaguya0: SDL3 changed `AudioSpec` to no longer include `samples` (https://wiki.libsdl.org/SDL3/SDL_AudioSpec, https://wiki.libsdl.org/SDL2/SDL_AudioSpec, https://wiki.libsdl.org/SDL3/README-migration#sdl_audioh), and from what i can tell it supports the callback pattern (https://examples.libsdl.org/SDL3/audio/02-simple-playback-callback/, https://github.com/libsdl-org/SDL/blob/main/examples/audio/02-simple-playback-callback/simple-playback-callback.c).

the move to SDL3 is also looking like it's finally almost unblocked now that https://github.com/libsdl-org/SDL_net/releases/tag/prerelease-3.1.0 exists

i'm ok with this landing as a temporary SDL2 fix, but i'd like to see if the improvements in SDL3 are enough to make adding another audio backend unnecessary (and ideally if it's good enough to let us use it everywhere instead of maintaining standalone backends for CoreAudio and WASAPI)

bassdr: I see where you are going, you'd prefer all platforms to use the SDL3 backend if possible. I think we should create a separate SDL3 backend in the meantime, and when it's stable we can migrate to this one. I personally prefer having os specific backends though, it's not like it's a lot of code either. A platform-specific ~200 lines of code is not the end of the world IMO.

But yes, why not ~60 that works for all platforms :)

bassdr: > i'm ok with this landing as a temporary SDL2 fix, but i'd like to see if the improvements in SDL3 are enough

I agree with you this is not ideal and can be considered a temporary fix. But I do agree we should merge the fix in the meantime.

IMO:
 - Temporary SDL2 fix
 - Create a new SDL3 backend (assuming we can have SDL2 and SDL3 side-by-side, TBC)
 - Check if it's awesome (it probably is!)
 - If it's working for all platforms, we can deprecate all other backends
 - If it's soso, we can think about merging my PipeWire backend and consider SDL2 backend a an OK fallback

BTW, I'm more interested in allowing modding the soundfont of the game, the audio backend fixes was ground work for this, I'm planning to have a look after.

briaguya0: > A platform-specific ~200 lines of code is not the end of the world IMO.

in general i agree. it's less about the maintenance surface and more about the question of where in the dependency stack interactions with OS level audio APIs should live. if the answer is "in SDL" then we can get the benefits of any low level improvements made in SDL, and if we find issues/see ways to improve it everyone using SDL audio benefits. it feels like an opportunity to be good open source citizens so to speak.

_____

> assuming we can have SDL2 and SDL3 side-by-side

we can't. the SDL3 audio improvements would need to happen after (or as a part of) migrating to SDL3

i'm planning on working to update
* https://github.com/HarbourMasters/Shipwright/pull/4859

this weekend, but the LUS submodule on the develop branch of soh is a couple months old at the moment so getting that up to date will need to happen first.

_____

the audio portion of LUS on that branch is a naive migration of what was in the file when i was working it last

https://github.com/Kenix3/libultraship/blob/2465c44d810d43422b16fb1c233a14314958a7a9/src/audio/SDLAudioPlayer.cpp#L31-L49

i'm assuming i'll land this PR _before_ updating the SDL3 branch so we'll be able to see if the issue comes back with the naive migration

src/ship/audio/SDLAudioPlayer.cpp
briaguya0: where `AudioPlayerPlayFrame` (or a port-specific thin wrapper for it: `AudioPlayer_Play`) is called in ports:

https://github.com/HarbourMasters/Shipwright/blob/772fe2bb92b3c75173435c98e17b7edc9ec2f7c0/soh/soh/OTRGlobals.cpp#L1044-L1045 (handles overflow)

https://github.com/HarbourMasters/2ship2harkinian/blob/48c022c36e06586ecf540e6a39e8842302685700/mm/2s2h/BenPort.cpp#L435-L436 (handles overflow the same way soh does)

https://github.com/HarbourMasters/Starship/blob/be6c807be42d155cdf1b369f4e55ce8670169371/src/port/Engine.cpp#L405-L406 (appears to have logic similar to soh/2ship)

https://github.com/HarbourMasters/SpaghettiKart/blob/efdaf9d8fdc99f79527220617ba0004fb882aad3/src/port/Engine.cpp#L508 (looks like the same pattern)

https://github.com/HarbourMasters/Ghostship/blob/463df67005499f8ca974189978d1a01d9ea56987/src/port/Engine.cpp#L879 (same pattern again)

so it looks like this is likely safe to remove

briaguya0 (approved):

src/ship/audio/SDLAudioPlayer.cpp
bassdr: I opened #1111 for the follow-up SDL3 backend. Feel free to comment there. I'll resove this because I think we should discuss the remaining either in the open issue or in Shipwright. I'll also create a draft PR because I did update LUS in my fork of Shipwright.

bassdr:
Opened #1189 as the port-maintenance counterpart (cherry-picks cleanly).

Kenix3 (approved):
Merge pending conflict resolution

bassdr:
@Kenix3 rebased onto main, conflicts resolved.

Worth a re-read before merging: the SDL3 migration in #1191 absorbed or obsoleted most of this PR, so it is now three lines - just removing the 6000-frame drop in `DoPlay()`. The description above has the item-by-item breakdown of what went where. The device-buffer doubling is the one deliberate omission: SDL3 has no `want.samples`, and its replacement hint is something SDL prefers to decide itself, so re-adding it belongs in a follow-up backed by measurements under the stream model rather than carried over from the SDL2 tuning.

@briaguya0 your approval was on the SDL2 diff, so it no longer describes what is here.

bassdr:
Closing this in favour of #1250, to keep the audio work down to one PR per branch instead of the four that were open.

@Kenix3 @briaguya0 - to be clear about what happens to the fix you approved: it is not dropped. Removing the `Buffered() < 6000` drop is the first thing #1250 does, for the same reason established in this thread, that the ports already bound their own buffering. #1250 then goes on to fix what the drop was standing in for, by giving `SDLAudioPlayer` its own ring buffer behind a pull callback.

Note your approvals here were against `a357d169`, before the rebase onto the SDL3 migration, so they no longer applied to what was in this PR anyway - the SDL3 migration had absorbed or obsoleted three of its four changes, leaving three lines.

If you would rather take the three-line fix on its own first and judge the pull model separately, say so and I will reopen this - it is a one-minute job and I would rather that than have the small fix held hostage to the larger argument.

PR #1109, opened 2026-05-19
Add fixedAspect flag to framebuffers

Pepper0ni:
Recently AdjXForAspectRatio was change to not get applied if resize was on, assuming that any framebuffer that resizes would want to be adjusted for the current windows aspect ratio, however an edge case in ship1's pause menu wanted both a resized viewport (to be full rendering resolution) and a fixed aspect ratio (as it is portrayed in a fixed size part of the UI), causing incorrect rendering in widescreen. 

This PR address this issue by adding a new flag named fixedAspect which specifically controls AdjXForAspectRatio. When true this causes AdjXForAspectRatio to not run, similar to resize = false before. The existing functionality can be replicated by setting fixedAspect to the inverse of resize in gfx_create_framebuffer.

Ports will need to add the extra argument to each gfx_create_framebuffer call after this merges.

### DIFF
--- a/include/fast/interpreter.h
+++ b/include/fast/interpreter.h
@@ -361,7 +361,7 @@ struct FBInfo {
     uint32_t applied_width, applied_height; // Up-scaled for the viewport
     uint32_t native_width, native_height;   // Max "native" size of the screen, used for up-scaling
     bool resize;                            // Scale to match the viewport
-    bool fixedAspect;                       // Preserve aspect ratio
+    bool forceFixedAspect;                  // Preserve aspect ratio even if resize is true
 };
 
 struct MaskedTextureEntry {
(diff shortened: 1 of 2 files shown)


ghost (commented):
The actual logic looks good -- just some better documentation suggestions.

include/fast/interpreter.h
```diff
@@ -361,6 +361,7 @@ struct FBInfo {
     uint32_t applied_width, applied_height; // Up-scaled for the viewport
     uint32_t native_width, native_height;   // Max "native" size of the screen, used for up-scaling
     bool resize;                            // Scale to match the viewport
+    bool fixedAspect;                       // Preserve aspect ratio
```
ghost: Reads like `fixedAspect = true` means "keep the aspect ratio as-is," but the code is actually applying a widescreen correction when `true` and skips it when `false`.

Something like `adjustForAspect` would read better, or flipping the logic so `fixedAspect = true` actually preserves the ratio like the comment implies.

src/fast/interpreter.cpp
```diff
@@ -1457,7 +1457,7 @@ float Interpreter::AdjXForAspectRatio(float x) const {
     // Skip widescreen adjustment for fixed-size off-screen FBs (HUD elements,
     // small capture buffers). But resizable FBs (resize=true) are capturing
     // the main scene and should match the window's aspect ratio.
```
ghost: The comment still explains the old `resize`-based logic.  Should be updated to describe `fixedAspect`.

Pepe20129 (approved):

briaguya0 (commented):
instead of making it so all consumers need to explicitly set `fixedAspect`, i'm thinking it makes sense to have this as `forceFixedAspect` and default it to false, then instead of _replacing_  `!resize` with `fixedAspect` we could have `!resize || forceFixedAspect`

briaguya0 (approved):

Issue #885, opened 2025-05-18
set `ImGui` `GamepadMode` to `AutoAll` instead of `AutoFirst`


briaguya0:
### imgui info:

https://github.com/ocornut/imgui/issues/6559

https://github.com/ocornut/imgui/blob/cdb5cbe6f8786ced227624a71bc8a677618e5327/backends/imgui_impl_sdl2.h#L41-L44

```cpp
// Gamepad selection automatically starts in AutoFirst mode, picking first available SDL_Gamepad. You may override this.
// When using manual mode, caller is responsible for opening/closing gamepad.
enum ImGui_ImplSDL2_GamepadMode { ImGui_ImplSDL2_GamepadMode_AutoFirst, ImGui_ImplSDL2_GamepadMode_AutoAll, ImGui_ImplSDL2_GamepadMode_Manual };
IMGUI_IMPL_API void     ImGui_ImplSDL2_SetGamepadMode(ImGui_ImplSDL2_GamepadMode mode, struct _SDL_GameController** manual_gamepads_array = nullptr, int manual_gamepads_count = -1);

```

### why i think autoall is better:

`AutoFirst` means `ImGui` will only allow the lowest indexed SDL controller to interact with the menu. there are plenty of scenarios where that controller isn't actually the one the player is using, and it gets even worse when dealing with disconnect/reconnect situations like https://github.com/HarbourMasters/Shipwright/issues/5437

by defaulting to `AutoAll` people will be able to navigate imgui menus in ports with any connected controller SDL knows about

KiritoDv:
This is pretty much why ios was broken too

briaguya0:
setting it to autoall does make it so multiple connected controllers work, but i still ran into an issue when trying to disconnect/reconnect

i set a breakpoint in https://github.com/ocornut/imgui/blob/cdb5cbe6f8786ced227624a71bc8a677618e5327/backends/imgui_impl_sdl2.cpp#L461-L466
```cpp
        case SDL_CONTROLLERDEVICEADDED:
        case SDL_CONTROLLERDEVICEREMOVED:
        {
            bd->WantUpdateGamepadsList = true;
            return true;
        }
```
and never hit it. i'm thinking that's because

https://github.com/Kenix3/libultraship/blob/6a3f6cd327b99f617b623e5b9a3afeae460aac2b/src/controller/physicaldevice/SDLAddRemoveDeviceEventHandler.cpp#L20-L32

[...]

PR #1112, opened 2026-05-30
Bind imgui's gamepad backend to the controller libultraship is using

JeodC:
If multiple controllers are connected. imgui would use the first one it found. In the event the first controller isn't actually a controller, or isn't the one in use, ControlNav becomes impossible. Instead, use the same gamepad that ControlDeck has already opened.

cc @KiritoDv

### DIFF
--- a/src/fast/Fast3dGui.cpp
+++ b/src/fast/Fast3dGui.cpp
@@ -10,6 +10,8 @@
 #include "ship/window/gui/resource/GuiTextureFactory.h"
 #include "ship/resource/File.h"
 
+#include <vector>
+
 #ifdef __APPLE__
 #include <SDL_hints.h>
 #include <SDL_video.h>
@@ -115,6 +117,9 @@ void Fast3dGui::ImGuiWMInit() {
         default:
             break;
     }
+
+    // Initial gamepad bind once the SDL2 backend exists (no-op for DX/Metal).
+    RefreshImGuiGamepads();
 }
 
 void Fast3dGui::ImGuiWMShutdown() {
@@ -239,6 +244,18 @@ void Fast3dGui::ImGuiWMNewFrame() {
         default:
             break;
     }
+}
+
+// Bind ImGui's SDL2 gamepad backend to the controller(s) the
+// ControlDeck has already opened
+void Fast3dGui::RefreshImGuiGamepads() {
+    auto window = Ship::Context::GetInstance()->GetWindow();
+    auto backend = window->GetWindowBackend();
+    if (backend != WindowBackend::FAST3D_SDL_OPENGL && backend != WindowBackend::FAST3D_SDL_METAL) {
+        return;
+    }
+
+    ImGui_ImplSDL2_SetGamepadMode(ImGui_ImplSDL2_GamepadMode_AutoAll, nullptr, 0);
 }
 
 void Fast3dGui::ImGuiRenderDrawData(ImDrawData* data) {
--- a/src/fast/Fast3dGui.cpp
+++ b/src/fast/Fast3dGui.cpp
@@ -10,8 +10,6 @@
 #include "ship/window/gui/resource/GuiTextureFactory.h"
 #include "ship/resource/File.h"
 
-#include <vector>
-
 #ifdef __APPLE__
 #include <SDL_hints.h>
 #include <SDL_video.h>
--- a/include/fast/Fast3dGui.h
+++ b/include/fast/Fast3dGui.h
@@ -133,6 +133,8 @@ class Fast3dGui : public Ship::Gui {
      */
     void LoadTextureFromResource(const std::string& name, std::shared_ptr<Ship::GuiTexture> texture);
 
+    void RefreshImGuiGamepads() override;
+
   protected:
     void ImGuiWMInit() override;
     void ImGuiWMShutdown() override;
--- a/include/ship/window/gui/Gui.h
+++ b/include/ship/window/gui/Gui.h
@@ -153,6 +153,11 @@ class Gui {
     /** @brief Re-enables ImGui gamepad navigation. */
     void UnblockGamepadNavigation();
 
+    /** @brief Re-binds the ImGui platform backend's gamepad list to the currently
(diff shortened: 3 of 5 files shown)


src/fast/Fast3dGui.cpp
```diff
@@ -228,9 +232,26 @@ void Fast3dGui::ImGuiWMNewFrame() {
     auto window = Ship::Context::GetInstance()->GetWindow();
     switch (window->GetWindowBackend()) {
         case WindowBackend::FAST3D_SDL_OPENGL:
-        case WindowBackend::FAST3D_SDL_METAL:
+        case WindowBackend::FAST3D_SDL_METAL: {
+            // Bind ImGui's SDL2 gamepad backend to the controller(s) the
+            // ControlDeck has already opened
+            auto controlDeck = Ship::Context::GetInstance()->GetControlDeck();
```
briaguya0: do we need to do this on every frame? i'd hope there'd be a better place to put this

JeodC: I found a better place to put it!

src/fast/Fast3dGui.cpp
```diff
@@ -241,6 +248,35 @@ void Fast3dGui::ImGuiWMNewFrame() {
...
 }
 
+// Bind ImGui's SDL2 gamepad backend to the controller(s) the
+// ControlDeck has already opened
+void Fast3dGui::RefreshImGuiGamepads() {
+    auto window = Ship::Context::GetInstance()->GetWindow();
+    auto backend = window->GetWindowBackend();
+    if (backend != WindowBackend::FAST3D_SDL_OPENGL && backend != WindowBackend::FAST3D_SDL_METAL) {
+        return;
+    }
```
briaguya0, line 254: curious as to why this check is here. we still use SDL for controllers on windows when rendering with directx.

src/fast/Fast3dGui.cpp
```diff
@@ -241,6 +248,35 @@ void Fast3dGui::ImGuiWMNewFrame() {
...
+
+    auto pads = cpdm->GetConnectedSDLGamepadsForPort(0);
+    std::vector<SDL_GameController*> arr;
+    for (const auto& kv : pads) {
+        arr.push_back(kv.second);
+    }
+    // Fall back to AutoAll if the port has no controllers yet, so the backend
+    // still discovers one rather than sitting on an empty manual list.
+    ImGui_ImplSDL2_SetGamepadMode(arr.empty() ? ImGui_ImplSDL2_GamepadMode_AutoAll : ImGui_ImplSDL2_GamepadMode_Manual,
+                                  arr.empty() ? nullptr : arr.data(), (int)arr.size());
```
briaguya0: not opposed to having this logic, but i want to make sure we aren't overengineering if we don't need to. the issue i noticed when looking into https://github.com/Kenix3/libultraship/issues/885 was that we're eating the `SDL_CONTROLLERDEVICEADDED` and `SDL_CONTROLLERDEVICEREMOVED` events so they never make it into ImGui.

in ImGui those events just set `WantUpdateGamepadsList`, and digging into `ImGui_ImplSDL2_SetGamepadMode` it appears [that sets it too](https://github.com/ocornut/imgui/blob/cdb5cbe6f8786ced227624a71bc8a677618e5327/backends/imgui_impl_sdl2.cpp#L727), so i'm thinking we might just be able to replace this entire block with

```cpp
ImGui_ImplSDL2_SetGamepadMode(ImGui_ImplSDL2_GamepadMode_AutoAll, nullptr, 0);
```

i'm not sure where `ImGui_ImplSDL2_GamepadMode_AutoFirst` is being set in the first place, so we'll probably want to figure that out too so we init ImGui in the right mode

briaguya0 (commented):
overall happy with the new placement of the refresh method, couple comments about implementation

src/fast/Fast3dGui.cpp
JeodC: The check isn't about how the game reads controllers, it's about which ImGui platform backend is live. `RefreshImGuiGamepads` only calls `ImGui_ImplSDL2_SetGamepadMode`, and the ImGui SDL2 backend is only initialized for the OpenGL/Metal window backends. On DX11 we init `ImGui_ImplWin32` instead, which does its own gamepad nav via XInput. The SDL2 backend isn't initialized on DX11, so it's better to guard.

src/fast/Fast3dGui.cpp
JeodC: On the `AutoFirst` question: it's not us; it's ImGui's own default, set inside `ImGui_ImplSDL2_Init`. Part of the fix is just calling `SetGamepadMode` at all, in the right mode, after init.

The block does one thing `AutoAll` doesn't: scope menu nav to the pad on port 0. With `AutoAll`, any connected controller can drive the menu; with Manual it's only player 1's. Single controller, they're identical.

If we don't care about multi-controller scoping I'll just drop it to the one-liner. If we do want it, the Manual/AutoAll-fallback is what gets it.

briaguya0: considering we default to all physical controllers working on port 1 I figure just having all controllers work in imgui fits better than trying to limit it. the only reason we have the current limitation is the imgui default

JeodC: Done.

src/fast/Fast3dGui.cpp
briaguya0: does DX have a similar problem? should we call the equivalent there instead of doing nothing?

JeodC: No, this bug only affects SDL2.

src/fast/Fast3dGui.cpp
```diff
@@ -10,6 +10,8 @@
 #include "ship/window/gui/resource/GuiTextureFactory.h"
 #include "ship/resource/File.h"
 
+#include <vector>
+
```
briaguya0: looks like we probably don't need this anymore

```suggestion
```

briaguya0 (approved):
one tiny thing then this should be good to merge. thanks for working with me through all my review comments!

PR #1114, opened 2026-05-30
fix(crash): replace exit(1) in ShutdownHandler with SDL_QUIT event

bassdr:
Calling exit() from a signal handler is not async-signal-safe: it triggers C++ destructors and atexit handlers which try to acquire mutexes already held by audio threads, causing a deadlock that the OS aborts as a crash. Push SDL_QUIT instead so the main event loop shuts down cleanly via the existing Close() path.

### DIFF
--- a/src/ship/debug/CrashHandler.cpp
+++ b/src/ship/debug/CrashHandler.cpp
@@ -217,7 +217,9 @@ static void ErrorHandler(int sig, siginfo_t* sigInfo, void* data) {
 }
 
 static void ShutdownHandler(int sig, siginfo_t* sigInfo, void* data) {
-    exit(1);
+    SDL_Event event;
+    event.type = SDL_QUIT;
+    SDL_PushEvent(&event);
 }
 
 #elif _WIN32


serprex (approved):

bassdr:
I just rebased on latest main to see if the CICD failures on MAC/IOS was caused by this. I can't explain how this could break them.

bassdr:
Opened #1188 as the port-maintenance counterpart, carrying just the `SDL_QUIT` ShutdownHandler fix. The "hold Window shared_ptr longer" commit isn't needed there since port-maintenance dispatches ImGui shutdown on the cached `mImpl.Backend` enum rather than the live window.

bassdr:
Rebased onto latest `main` (the branch had gone stale/conflicting). Dropped the second commit ("Hold on Window shared_ptr longer"): `main` now caches the window in an `mWindow` member held for the object's lifetime, so the extra shared_ptr holding is redundant and would fight that design. This PR is now just the `exit(1)` -> `SDL_QUIT` ShutdownHandler fix. Port-maintenance counterpart is #1188.

Kenix3 (approved):

PR #1115, opened 2026-05-30
Fix undefined shifts in the GBI macros and two divisions by zero

bassdr:
Two UB fixes found with UBSan. Rebased onto current `main` and cut down from the original catch-all, which was too broad to review.

**Unsigned bit macros.** `_SHIFTL(1, 0, 31)` evaluates `0x01 << 31` in `int`, overflows, then subtracts 1 from the result - two UBSan reports out of one macro (`left shift of 1 by 31 places cannot be represented in type 'int'`). The render mode flags, the blend macros and the button masks have the same problem at bit 31. Suffixing the constants with `u` puts the arithmetic where the wrap is defined.

Every value these macros expand to is unchanged - checked, not assumed: a program expanding `G_RM_*`, `GBL_c1`/`GBL_c2`, `GPACK_*`, `_SHIFTL`/`_SHIFTR` and `BTN_*` gives byte-identical output built against `main` and against this branch.

**Two divisions by zero.** A zero-length normal made `NormalizeVector` hand NaN to the lighting maths; a tile that was never loaded reports size zero, so the UV normalisation in `GfxSpTri1` put inf or NaN into the vertex buffer. Neither has a direction or scale worth preserving, so both fall back to zero.

Dropped from the original: the `tex_width2`/`tex_height2` rounding change - it clamps to a minimum of 1, which makes the `renderedPx > 0` gate below it fire where it currently skips, so it is a rendering change needing its own testing - plus the `#undef` blocks and the whitespace churn.

### DIFF
--- a/include/fast/lus_gbi.h
+++ b/include/fast/lus_gbi.h
@@ -164,8 +164,9 @@ constexpr int8_t RDP_G_TRI1_WIDE = OPCODE(0x49);
  */
 #define G_MAXFBZ 0x3fff /* 3b exp, 11b mantissa */
 
-#define GPACK_RGBA5551(r, g, b, a) ((((r) << 8) & 0xf800) | (((g) << 3) & 0x7c0) | (((b) >> 2) & 0x3e) | ((a)&0x1))
-#define GPACK_ZDZ(z, dz) ((z) << 2 | (dz))
+#define GPACK_RGBA5551(r, g, b, a) \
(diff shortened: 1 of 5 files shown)


Kenix3 (approved):

bassdr:
Rebased and trimmed rather than closed - the original was a catch-all that never got reviewed, which was my fault for bundling. None of it had been fixed upstream in the meantime, so the content was still live; it is now two commits over five files, each with a stated reason, and the behavioural rendering change is dropped out to be argued separately if it is wanted at all.

PR #1116, opened 2026-05-30
audio: optional float pipeline end-to-end (opt-in, s16 default)

bassdr:
AudioPlayer gains a parallel float-precision pipeline that consumers can
opt into via AudioSettings::UseFloatPipeline. The s16 path is unchanged
and remains the default, so existing libultraship consumers keep their
byte-exact contract; SoH flips to float when FluidSynth is enabled.

Based on top of #1106 
Closes #1106 

New entry points

- AudioPlayer::Play(const float*, size_t frames) alongside the legacy
  Play(const uint8_t*, size_t len). Each Play asserts it's called in
  the matching mode and drops the buffer with a warning otherwise.
- AudioPlayer::SetUseFloatPipeline(bool) — runtime mode switch via
  DoClose → flip → DoInit. Reverts on failure.
- AudioPlayer::SetMixSource(std::function<void(float*, int)>) — a
  secondary stereo source mixed in *after* the resampler so its
  contribution skips the rate-conversion step entirely. Returns false
  in s16 mode. AudioPlayer sums the source with a tanh-style soft-clip
  before any surround decode.
- audiobridge: AudioPlayerPlayFrame (legacy uint8_t) preserved;
  AudioPlayerPlayFrameF32 added for the float path.

DSP layer

- AudioResampler::Process gains a float overload that operates on
  interleaved float [-1, 1]. The original int16_t overload is preserved
  and wraps the float core with boundary conversions and clamping.
- SoundMatrixDecoder::Process likewise: native float overload plus a
  legacy uint8_t (s16) overload that converts at the boundaries. Both
  reuse the float-internal filter / phase / delay state.

Backends
- SDL / PipeWire / WASAPI / CoreAudio DoInit reads
  IsUsingFloatPipeline() and configures the device format (F32 vs S16).
  Buffered() divides by the matching sample size.
- PipeWire's ring init, sample width in OnProcess, and underrun
  fade-out math all branch on the same flag.

AudioPlayer reorder in float mode
- Stages: resample stereo → optional MixSource sum + soft-clip →
  surround decode (matrix-5.1) → DoPlay. Lets a secondary source
  produced at GetSampleRate() bypass the resampler.
- Resampler channel count differs between modes: stereo (2) in float
  mode (mix + surround decode follow), GetNumOutputChannels() in s16
  mode (legacy decode-first order preserved). RebuildResampler() picks
  the right value on Init / SetUseFloatPipeline / SetAudioChannels.

### DIFF
--- a/include/libultraship/bridge/audiobridge.h
+++ b/include/libultraship/bridge/audiobridge.h
@@ -22,12 +22,29 @@ API_EXPORT AudioChannelsSetting GetAudioChannels();
 API_EXPORT int32_t GetNumAudioChannels();
 
 /**
- * @brief Submits a frame of PCM audio to the audio output device.
+ * @brief Submits a frame of s16 PCM audio to the audio output device (legacy).
+ *
+ * Default entry point preserved for libultraship consumers on the s16
+ * audio path. Forwards to AudioPlayer::Play(uint8_t*, size_t); valid only
+ * when the player is in s16 mode (the default).
  *
  * @param buf Interleaved sample data (stereo: L,R,… or surround: FL,FR,C,LFE,SL,SR,…).
  * @param len Length of @p buf in bytes.
  */
 API_EXPORT void AudioPlayerPlayFrame(const uint8_t* buf, size_t len);
+
+/**
+ * @brief Submits a frame of float PCM audio to the audio output device.
+ *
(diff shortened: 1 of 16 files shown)


bassdr:
Close #1106

bassdr:

PR #1117, opened 2026-05-30
audio: Add FluidSynth MIDI renderer as an alternative to native audio renderer

bassdr:
## Summary

Adds an optional soft-synth abstraction layer that lets a consumer replace the engine's native PCM synthesis with a MIDI-driven backend (FluidSynth). Zero impact when disabled — when no synth is installed MidiSynthManager::GetActiveSynth() returns nullptr and the audio thread is the exact original PCM path.

Requires the float pipeline PR (this branch is based on it). FluidSynth renders at device output rate and feeds in via SetMixSource, bypassing the resampler.

Based on #1116 
Closes #1116 
Closes #1106 

## Testing

You can load any SF2 file and configure some instruments to play it.
I personnally used [MuseScore General](https://ftp.osuosl.org/pub/musescore/soundfont/MuseScore_General/) for my experimentation. Grab the .sf2 version. I can let the .sf3 file work later, FluidSynth supports it.

You can use this configuration to test, it sounds pretty good IMO: 
[fluidsynth_overrides.json](https://github.com/user-attachments/files/28580377/fluidsynth_overrides.json)

## Core abstraction

  - IMidiSynth: pure-virtual interface — NoteOn, NoteOff, ProgramChange, PitchBend, ControlChange, Render
  - MidiSynthManager: thread-safe singleton. SetSynth() must not be called from the audio thread. GetActiveSynth() returns nullptr when no synth is installed.
  - FluidSynth: IMidiSynth implementation backed by libfluidsynth. Gated by -DENABLE_FLUIDSYNTH=ON (default OFF); FluidSynth.cpp is excluded from the build when the option
  is off, so there is no link-time dependency unless explicitly requested.

## Graham-Smith volume curve (optional)

Also adds a mode so FluidSynth.

FluidSynth(sampleRate, linearVelocity) — when linearVelocity=true, replaces three SF2 default modulators on GEN_ATTENUATION with versions that keep the concave negative
curve shape but halve the amount (960 cB → 480 cB). This lifts quiet voices ~6 dB at mid-range without flattening dynamics. Technique credited to ANMP (GPL-2, github.com/derselbst/ANMP).
  
Also raises synth.gain from FluidSynth's stock 0.2 to 1.0 unconditionally — at 0.2 the synth output peak is ~5× quieter than a typical PCM mix partner.

## SetReverbParams and similar effects runtime method
  
Lets callers swap reverb presets without rebuilding the synth. Safe to call any time after construction.

## Support adding `*.sf2` and `*.sf3` files in mods

Adding SoundFonts to mods is now supported. They can be read directly from the `*.o2r` files. Adding a *.json with the same name of the soundfont will also provide a mapping that can be used to ship pre-configured instrument mapping alongside the soundfont.

## Design notes
  
  - SetSynth(nullptr) gracefully uninstalls the active synth; the audio thread reverts to native PCM on the next buffer.
  - Render() is called from the audio thread and must be real-time safe. The FluidSynth implementation holds a mutex shared with all event methods (NoteOn, ProgramChange, etc.).
  - FluidSynth reads synth.sample-rate once at construction via new_fluid_synth() — fluid_synth_set_sample_rate() is deprecated in FluidSynth 2.x and silently ignored.
  - The linearVelocity name is kept (rather than renamed to grahamSmith) for CVar key stability in consuming projects.


bassdr:

PR #1118, opened 2026-05-30
Skip some clamps and stride calculations for replacement textures

JeodC:
#1043 introduced a regression for alt assets which are much larger than native game textures. In accordance, these need to follow different rules. Regressions spotted in Ghostship, tested and verified resolved.

### DIFF
--- a/include/ship/resource/ResourceManager.h
+++ b/include/ship/resource/ResourceManager.h
@@ -224,6 +224,15 @@ class ResourceManager {
     size_t UnloadResource(const std::string& filePath);
 
     /**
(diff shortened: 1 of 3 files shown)


JeodC:
The latest commit https://github.com/Kenix3/libultraship/pull/1118/commits/7d782ac13e878f59ce398da413fb38ac1137651a is experimental. Banjo-Kazooie's bold font letters and numbers are created at runtime by computing `mask + sphere` assets. The end result has no asset path.

I tested the feature with a fairly poor scaling of the existing assets, and while they *seem* to work, they are no replacement for the HD work GhostlyDark does.

briaguya0 (approved):

PR #1120, opened 2026-06-06
Fix shutdown crash

xxAtrain223:
Pass the window down to the shutdown functions instead of using `Ship::Context::GetInstance()->GetWindow()` which is nulled out in the Context Destructor.

### DIFF
--- a/include/fast/Fast3dGui.h
+++ b/include/fast/Fast3dGui.h
@@ -137,9 +137,9 @@ class Fast3dGui : public Ship::Gui {
 
   protected:
     void ImGuiWMInit() override;
-    void ImGuiWMShutdown() override;
+    void ImGuiWMShutdown(Ship::Window* window) override;
     void ImGuiBackendInit() override;
-    void ImGuiBackendShutdown() override;
+    void ImGuiBackendShutdown(Ship::Window* window) override;
     void ImGuiBackendNewFrame() override;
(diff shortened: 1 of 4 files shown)


serprex:
does this conflict with #1114? Passing window around seems nice, tho #1114's SDL_QUIT seems correct outside all this

there's also #1103

xxAtrain223:
It would conflict, seems like two approaches to the same problem, although #1114 definitely covers more cases, it does look like SDL_QUIT should be done regardless.

My intention was to replicate the behavior from here: https://github.com/Kenix3/libultraship/commit/8813b8233d7209a530c95222cc578c1cab870064#diff-c90358fbeace7bef94d95ec22dbce42e98498acecd510e7437e13c333e513b8bL184-L206

That said, I'm not too familiar with this part of the code base, so I'd defer to someone who knows more.

bassdr:
#1114 covers more cases yes, I fixed two crashes in the same PR, SDL_QUIT and held on the shared_ptr longer too for the "normal" shutdown, both were crashing on me at shutdown so I fixed them both. Your approach is OK (actually better in the sense the ownership of the window is clearer) but it requires to modify all overriding functions, so my approach is more "lean".

xxAtrain223:
Closing in favor of other pull requests.

PR #1122, opened 2026-06-09
Skip N64 segmented-range guard in SETTIMG when a resource was loaded

briaguya0:
## Summary

`gfx_set_timg_handler_rdp` rejects any resolved address ≤ `0x0FFFFFFF` as an "unresolved N64 segmented address" (added in #1042). The check fires *after* a successful OTR resource load, where `i` has been reassigned to the host pointer `tex->ImageData`. On Linux with ASLR enabled, that heap pointer reliably lands well above the threshold, so the bug is invisible in normal use.

When ASLR is disabled — via `setarch --addr-no-randomize`, or implicitly by gdb's default `set disable-randomization on` — the heap starts low enough that legitimate `tex->ImageData` pointers fall under `0x0FFFFFFF`. The guard then rejects nearly every texture load, leaving stale F3D state behind. Downstream import functions read mismatched format/dimensions and emit garbled textures.

## Fix

Track whether a resource was loaded and only apply the range check in the no-resource fallback case, where `i` is still the raw `SegAddr` result that the guard is actually meant to filter.

## Reproduction

Before this change, with vanilla SoH + this libultraship:

```sh
cd build/soh
setarch --addr-no-randomize ./soh.elf
# or just run under gdb — same effect, since gdb disables ASLR by default
```

Result: garbled textures on the N64 logo intro, file select, and title screens. With the patch, all three render cleanly.

The visible upload divergence:

```
ASLRUPLOADSRC fn=ImportTextureI8 width=192 height=2   # ASLR on
ASLRUPLOADSRC fn=ImportTextureI8 width=384 height=2   # ASLR off — wrong texture in tmem[0]
```

## Test plan

- [x] Verified locally on Fedora 44 in a distrobox: corruption reproduces with `setarch --addr-no-randomize` on the parent commit, disappears with this patch applied.
- [ ] No regression in normal-ASLR runs (heap pointers are above the threshold either way; behavior unchanged).
- [ ] Windows path untouched (`!resourceLoaded` gates both the Linux and Windows branches identically).

### DIFF
--- a/src/fast/interpreter.cpp
+++ b/src/fast/interpreter.cpp
@@ -3852,6 +3852,7 @@ bool gfx_set_timg_handler_rdp(F3DGfx** cmd0) {
     // from the resource, but raw textures would leave them at 0.
     rawTexMetdata.h_byte_scale = 1;
     rawTexMetdata.v_pixel_scale = 1;
+    bool resourceLoaded = false;
 
     if ((i & 1) != 1) {
         if (gfx_check_image_signature(imgData) == 1) {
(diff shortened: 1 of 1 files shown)


briaguya0:
@JeodC hoping you can take a peek at this one. it seems `i <= 0x0FFFFFFF` isn't something we can rely on to mean "still in the N64 segmented range" 

this fixes it for me but i'd like to better understand this and figure out the best way to handle it.

JeodC:
> @JeodC hoping you can take a peek at this one. it seems `i <= 0x0FFFFFFF` isn't something we can rely on to mean "still in the N64 segmented range"
> 
> this fixes it for me but i'd like to better understand this and figure out the best way to handle it.

The `i <= 0x0FFFFFFF` guard exists because when a segment hasn't been set up yet, it can't resolve and SegAddr just hands the raw address right back. Those usually live in the low N64 range, so passing one into `GfxDpSetTextureImage` and dereferencing it later would crash. The guard is a cheap way of saying "this is an unresolved address, don't touch".

The assumption that nothing at or below ` 0x0FFFFFFF` could be a real resolved pointer is wrong on my part, then (see #1004). The check was only ever meant to vet raw SegAddr output, but by the time it runs `i` has already been reassigned to `tex->ImageData` on the OTR path, so it's being applied to a host heap pointer where the threshold means nothing.

ASLR is on by default on Linux, so the memory heap maps high. The issue is normally invisible. Your PR is the correct handling. However, the old guard incidentally caught a null/low `tex->ImageData` on a malformed resource; this PR now passes that straight through. It might not be a problem, but if we want that safety back it's more meaningful as an explicit `tex->ImageData != nullptr` check at the load site than as an address-range test.

More broadly, both discriminators leak in opposite directions: `w1 & 1` (tag) fails for odd host pointers, and `w1 > 0x0FFFFFFF` (magnitude) fails for low host pointers. Neither is robust because both inspect the integer values to guess provenance. In SpaghettiKart, the `ALIGNED8` and `ALIGNED16` macros make host pointers reliably even, which makes the tag method unambiguous *and* ASLR-independent. But it adds one more piece to the existing contract, on the port side.

Regarding this pull request: it's provenance-based so correct under ASLR on or off, and orthogonal to the SegAddr redesign I proposed in #1004. If we ever resume discussion on a redesign to use magnitude instead of tagging, we will also need an answer for the ASLR-off case.

briaguya0:
still not sure the best way to handle this in code, posting my local `.vscode/settings.json` workaround for now

```json
{
  "cmake.debugConfig": {
    "setupCommands": [
      { "description": "Keep ASLR on — SoH has an ASLR-sensitive bug", "text": "set disable-randomization off" }
    ]
  }
}
```

PR #1125, opened 2026-06-12
SDL3 audio backend

### LINKED ISSUE #1111: Create an SDL3 audio backend
This is a follow-up of #1107. For context: We want to see if an SDL3 audio backend would work fine on all platforms, allowing to deprecate other platform-specific audio backends, and potentially keep SDL2 as a fallback.

bassdr:
This brings SDL3 as an optional dependency to allow using SDL3 as an audio backend. This is working much better on Linux than the only SDL2 audio backend available.

As discussed in #1105 and #1107, currently the SDL audio backend has quite a lot of audio artifacts. On Windows and macOS, the dedicated WASAPI and CoreAudio backends make it less important, as  these backends are of good quality (fewer audio artifacts), but Linux only has the SDL2 backend available.
Since fixing SDL2 backend or merging my PipeWire backend (Linux modern audio api) does not seem to have a lot of interest, it was discussed an SDL3 backend would be a better fit for the project.

This PR does not need to migrate the whole project to SDL3, and is designed to work on all platforms (only tested on Linux so far), while we are waiting for a full migration.

To achieve SDL2 + SDL3 loading side-by-side, SDL3 is loaded dynamically (rather than linked) on purpose: SDL2 and SDL3 export overlapping SDL_* symbol names, so linking both into one binary makes the shared names (e.g. SDL_InitSubSystem) bind to SDL2 — breaking SDL3 init. Loading SDL3 with local visibility (RTLD_LOCAL) keeps its symbols isolated, and also makes it a soft runtime dependency.

Closes #1111

### DIFF
--- a/include/ship/audio/Audio.h
+++ b/include/ship/audio/Audio.h
@@ -9,7 +9,7 @@ namespace Ship {
 class Config;
 
 /** @brief Identifies the audio backend implementation in use. */
-enum class AudioBackend { WASAPI, SDL, COREAUDIO, NUL };
+enum class AudioBackend { WASAPI, SDL, SDL3, COREAUDIO, NUL };
 
 /**
  * @brief Manages audio playback through a platform-specific AudioPlayer.
--- a/include/ship/audio/AudioPlayer.h
+++ b/include/ship/audio/AudioPlayer.h
@@ -1,6 +1,6 @@
 #pragma once
-#include "stdint.h"
-#include "stddef.h"
+#include <cstdint>
+#include <cstddef>
 #include <string>
 #include <memory>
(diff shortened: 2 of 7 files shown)


bassdr:
Superseded by #1250 and closing.

#1191 put libultraship on SDL3, which removed the reason this PR was shaped the way it was: no SDL2/SDL3 symbol clash to dodge, so no runtime loader, no `ENABLE_SDL3` CMake option, no separate `AudioBackend::SDL3`, no pimpl hiding SDL3 types from SDL2 code. What was actually worth keeping - the pull callback, the ring buffer and the underrun masking - belongs in `SDLAudioPlayer` itself, which is what #1250 does in two files.

PR #1126, opened 2026-06-12
Fix warnings

Pepe20129:
Fix 2 warnings that are in headers and show up in port builds.

### DIFF
--- a/include/fast/lus_gbi.h
+++ b/include/fast/lus_gbi.h
@@ -134,6 +134,9 @@ constexpr int8_t RDP_G_TRI1_WIDE = OPCODE(0x49);
 
 /* macros for command parsing: */
 #define GDMACMD(x) (x)
+#ifdef GIMMCMD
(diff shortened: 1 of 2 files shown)


include/fast/lus_gbi.h
```diff
@@ -134,6 +134,9 @@ constexpr int8_t RDP_G_TRI1_WIDE = OPCODE(0x49);
 
 /* macros for command parsing: */
 #define GDMACMD(x) (x)
+#ifdef GIMMCMD
+#undef GIMMCMD
+#endif
```
briaguya0, line 139: we could just do

```suggestion
#undef GIMMCMD
```

because it won't cause problems if it isn't defined, but that's non-blocking

briaguya0 (approved):
@Pepe20129 i'm going to merge this, if you could make another PR with these changes to `main` so we don't lose these fixes over there that'd be perfect

PR #1128, opened 2026-06-12
Implemented mipmaps, moved some interpreter functions to the GPU, added vulkan and implemented shader packs

KiritoDv:
Depends on <https://github.com/Kenix3/libultraship/pull/1110> So im gonna rebase after that


louist103:
Shouldn't these be multiple PRs?

KiritoDv:
> Shouldn't these be multiple PRs?

A lot is gonna disappear when i rebase, thats why im saying its missing, also no because a lot of things depend on each other, probably the shader packs one, but i still need feedback regarding that

src/fast/shaders/vulkan/default.shader.glsl
```diff
@@ -0,0 +1,262 @@
+@prism(type='fragment', name='Fast3D Vulkan Shader', version='1.0.0', description='Fast3D combiner shader for Vulkan', author='Ghostship')
```
serprex, line 1: how's perf here? @natebuckareff went with ubershaders when taking a stab at vulkan backend due to shader compilation being so much heavier on vulkan

serprex:
> > Shouldn't these be multiple PRs?
> 
> A lot is gonna disappear when i rebase, thats why im saying its missing, also no because a lot of things depend on each other, probably the shader packs one, but i still need feedback regarding that

you can break up a PR into a series of a parts that depend on each other in sequence

louist103:
> > Shouldn't these be multiple PRs?
> 
> A lot is gonna disappear when i rebase, thats why im saying its missing, also no because a lot of things depend on each other, probably the shader packs one, but i still need feedback regarding that

I didn't read through the all the changes but I doubt you can't implement the mipmaps without also adding vulkan.

PR #1129, opened 2026-06-12
Fix NullAudioPlayer::Buffered() returning 0 (audio-thread spin/hang)

bassdr:
It discards all audio, so reporting 0 made the producer spin feeding it forever and starve the gfx thread. Return the desired buffered count, as the header already documents.


serprex (approved):

briaguya0 (approved):

serprex:
@bassdr could you also create this pointed at `port-maintenance` branch?

Kenix3 (approved):

PR #1131, opened 2026-06-15
Only populate event callers map in debug builds

JeodC:
In production, the event debugger is a high cost. The debugger wouldn't usually be used by consumers, and developers testing Release builds can just as easily disable the preprocessor gate.

### DIFF
--- a/src/ship/events/EventSystem.cpp
+++ b/src/ship/events/EventSystem.cpp
@@ -59,6 +59,8 @@ void EventSystem::CallEvent(const EventID id, IEvent* event, const char* file, c
         listener.Function(event);
     }
 
+#ifdef _DEBUG
+    // Event Debugger is a dev tool and shouldn't be compiled into Releases
     auto& info = registry.Callers[key];
 
     if (info.Path == nullptr) {
@@ -67,6 +69,7 @@ void EventSystem::CallEvent(const EventID id, IEvent* event, const char* file, c
     }
 
     info.Count++;
+#endif
 }
 
 } // namespace Ship


Malkierian:
Would it be OK for this to target port-maintenance, too?

JeodC:
> Would it be OK for this to target port-maintenance, too?

So long as nobody loses track of it and #1118 getting to main eventually. @briaguya0?

Malkierian:
Kenix said he would cherry pick as needed.

Malkierian:
Perhaps maintaining a list of PRs that need to be cherry-picked would be a good idea.

src/ship/events/EventSystem.cpp
```diff
@@ -67,6 +69,7 @@ void EventSystem::CallEvent(const EventID id, IEvent* event, const char* file, c
     }
 
     info.Count++;
+#endif
```
briaguya0, line 72: it's not clear to me how this is disabling the event debugger. it's just not adding file and line to info? if `#ifdef`ing these lines solves the perf issues i'm cool with landing this on port-maint, but it'd be good to think through what conditionally including the debugger should look like when getting the change to `main`

JeodC: `CallEvent()` runs for every event fired in a source port. The gated block here normally runs unconditionally inside that. Every event call does a hash-map lookup and a possible insert, plus bookkeeping.

I brought it up to @KiritoDv and they requested it be kept like this for now.

briaguya0: i understand how it's a performance improvement, what's not clear is how this counts as "disabling the event debugger"

the `EventDebuggerWindow` isn't touched by this PR, and that shows information about both the callers and the listeners.

looking at `Callers`
https://github.com/Kenix3/libultraship/blob/a3f1e102e1ed50a3f536fbca2e6e412170e7320b/include/ship/events/EventSystem.h#L24-L25

it seems a more accurate title/commit message for this would be "don't populate event callers map in release builds" or "only populate event callers map in debug builds"

JeodC: Agreed. I'll make that change.

PR #1133, opened 2026-06-17
Fix missing rumble/LED/gyro for PS4/PS5 controllers over Bluetooth

crocchetto:
Rumble, LED control, and gyro are unavailable when connected via Bluetooth because SDL2 gates them behind "extended input report" mode, off by default, which must be requested via hint before SDL_Init().

Tested on macOS 15.6.1, rumble/LED/gyro now work over Bluetooth.

### DIFF
--- a/src/libultraship/libultra/os.cpp
+++ b/src/libultraship/libultra/os.cpp
@@ -24,6 +24,10 @@ int32_t osContInit(OSMesgQueue* mq, uint8_t* controllerBits, OSContStatus* statu
     }
 
     SDL_SetHint(SDL_HINT_JOYSTICK_THREAD, "1");
+    SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_PS4, "1");
+    SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_PS4_RUMBLE, "1");
+    SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_PS5, "1");
+    SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_PS5_RUMBLE, "1");
     if (SDL_Init(SDL_INIT_GAMECONTROLLER) != 0) {
         SPDLOG_ERROR("Failed to initialize SDL game controllers ({})", SDL_GetError());
         exit(EXIT_FAILURE);


src/libultraship/libultra/os.cpp
```diff
@@ -24,6 +24,10 @@ int32_t osContInit(OSMesgQueue* mq, uint8_t* controllerBits, OSContStatus* statu
     }
 
     SDL_SetHint(SDL_HINT_JOYSTICK_THREAD, "1");
+    SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_PS4, "1");
```
briaguya0, line 27: this should already be set

from https://wiki.libsdl.org/SDL2/SDL_HINT_JOYSTICK_HIDAPI_PS4
> The default is the value of [SDL_HINT_JOYSTICK_HIDAPI](https://wiki.libsdl.org/SDL2/SDL_HINT_JOYSTICK_HIDAPI)

from https://wiki.libsdl.org/SDL2/SDL_HINT_JOYSTICK_HIDAPI
> "1": HIDAPI drivers are used (the default)

src/libultraship/libultra/os.cpp
```diff
@@ -24,6 +24,10 @@ int32_t osContInit(OSMesgQueue* mq, uint8_t* controllerBits, OSContStatus* statu
...
+    SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_PS4_RUMBLE, "1");
+    SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_PS5, "1");
```
briaguya0, line 29: same as `HIDAPI_PS4`

from https://wiki.libsdl.org/SDL2/SDL_HINT_JOYSTICK_HIDAPI_PS5
> The default is the value of [SDL_HINT_JOYSTICK_HIDAPI](https://wiki.libsdl.org/SDL2/SDL_HINT_JOYSTICK_HIDAPI)

src/libultraship/libultra/os.cpp
```diff
@@ -24,6 +24,10 @@ int32_t osContInit(OSMesgQueue* mq, uint8_t* controllerBits, OSContStatus* statu
     }
 
     SDL_SetHint(SDL_HINT_JOYSTICK_THREAD, "1");
+    SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_PS4, "1");
+    SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_PS4_RUMBLE, "1");
```
briaguya0, line 28: This doesn't come without side-effects

from https://wiki.libsdl.org/SDL2/SDL_HINT_JOYSTICK_HIDAPI_PS4_RUMBLE
> Extended input reports allow rumble on Bluetooth PS4 controllers, but break DirectInput handling for applications that don't use SDL.
> 
> Once extended reports are enabled, they can not be disabled without power cycling the controller.

src/libultraship/libultra/os.cpp
```diff
@@ -24,6 +24,10 @@ int32_t osContInit(OSMesgQueue* mq, uint8_t* controllerBits, OSContStatus* statu
...
+    SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_PS5, "1");
+    SDL_SetHint(SDL_HINT_JOYSTICK_HIDAPI_PS5_RUMBLE, "1");
```
briaguya0, line 30: same as `HIDAPI_PS4_RUMBLE`

briaguya0 (changes requested):
2 of the hints aren't needed, the other 2 we shouldn't set unconditionally.

i don't want people with ps4/ps5 controllers to need to power cycle them whenever they open soh and then want to play another game.

if we add these, it should be done in an opt-in way

PR #1139, opened 2026-07-01
Add ImGui FileBrowserWindow as a fallback file dialog

JeodC:
Relies on #1138 

Adds `Ship::FileBrowserWindow`, a self-contained ImGui file dialog for platforms that have no native file dialog or display server (Switch, Wii U, ARM Linux handhelds). It walks the filesystem and picks a file using only ImGui, and is navigable by mouse, keyboard, or controller.

**portable-file-dialogs remains the default.** This change is purely additive. Desktop ports keep using native OS dialogs (pfd) exactly as before. Nothing here routes anyone off them. The ImGui browser is opt-in, for the platforms where a native dialog isn't available. A consuming port picks per platform (e.g. native dialog on desktop, this browser on console/ARM); LUS provides the widget but does not impose the choice.

## Contract between LUS and the consuming port

LUS owns the widget; the port owns the semantics. The browser is domain-agnostic.

- **Registration.** `Gui` constructs one `FileBrowserWindow` at startup. It stays invisible and renders nothing until a request is queued, so there's no per-port setup.
- **Describe the pick** with `FileBrowserRequest`:
  - `Title`: the modal title.
  - `StartDir`: directory to open in; empty resumes the last directory browsed this session, then falls back to the working directory.
  - `Filters`: label + glob sets (`"*.z64"`, `"*"`); empty shows all files.
  - `Save` + `DefaultName`: save mode adds a filename field and returns `directory/filename`.
- **Receive the result** through `OnResult(std::optional<path>)`: the chosen path on confirm, `std::nullopt` on cancel. It fires **exactly once**, on the **render thread**, a later frame than the `Open()` call.
- **Mind lifetime.** Because `OnResult` can resolve frames later, anything it captures (a `this`, a buffer) must outlive the request. Capture a `shared_ptr` when you can't otherwise guarantee it.
- **Two entry points:**
  - `Open(request)`: queue and return immediately; safe from any thread.
  - `OpenBlocking(request)`: queue and block until the render loop resolves it; safe **only off** the render thread.
- **Drive your own flow** with `IsOpen()` while a pick is in flight (e.g. to hold a boot step or keep the render loop running).

## Supporting changes

To make the browser usable with a controller in pre-game UI (before the game boots its controller), this also:

- Brings up the SDL game-controller subsystem earlier, in `Context::InitControlDeck`, moved out of `osContInit` (which still runs `ControlDeck::Init()` since it needs the game's `controllerBits`). Controllers are then available for ImGui navigation of pre-game prompts.
- Adds `Gui::UpdateGamepadNavigation()`, a per-frame recompute of gamepad-nav enablement: on while a menu or any popup is open (and controller nav is enabled), off during gameplay so the game keeps the pad. Centralised because popups have no open/close event to hook.

## Platform notes

- The widget is POSIX-oriented (Home/Documents/… shortcuts, `/mnt`·`/media` device scanning). It compiles on all platforms but is intended for the console/ARM builds that actually use it; desktop keeps native dialogs.

https://github.com/user-attachments/assets/476a1d1b-a6f2-4328-85cf-9eca173696aa

### DIFF
--- a/src/ship/Context.cpp
+++ b/src/ship/Context.cpp
@@ -2,6 +2,7 @@
 #include "ship/controller/controldevice/controller/mapping/keyboard/KeyboardScancodes.h"
 #include <cstring>
 #include <iostream>
+#include <SDL2/SDL.h>
 #include <spdlog/sinks/rotating_file_sink.h>
 #include <spdlog/sinks/stdout_color_sinks.h>
 #include "ship/install_config.h"
@@ -268,6 +269,21 @@ bool Context::InitControlDeck(std::shared_ptr<ControlDeck> controlDeck) {
     if (GetControlDeck() == nullptr) {
         SPDLOG_ERROR("Failed to initialize control deck");
         return false;
+    }
+
+    // Bring up the SDL game-controller subsystem here rather than in osContInit, so controllers work
+    // in pre-game UI (e.g. navigating extraction prompts). osContInit still runs ControlDeck::Init(),
+    // which needs the game's controllerBits.
+    std::string controllerDb = LocateFileAcrossAppDirs("gamecontrollerdb.txt");
+    int mappingsAdded = SDL_GameControllerAddMappingsFromFile(controllerDb.c_str());
+    if (mappingsAdded >= 0) {
+        SPDLOG_INFO("Added SDL game controller mappings from \"{}\" ({})", controllerDb, mappingsAdded);
+    } else {
+        SPDLOG_ERROR("Failed to add SDL game controller mappings from \"{}\" ({})", controllerDb, SDL_GetError());
+    }
+    SDL_SetHint(SDL_HINT_JOYSTICK_THREAD, "1");
+    if (SDL_Init(SDL_INIT_GAMECONTROLLER) != 0) {
+        SPDLOG_ERROR("Failed to initialize SDL game controllers ({})", SDL_GetError());
     }
 
     return true;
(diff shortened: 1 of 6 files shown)


Malkierian:
Some conflicts came up.

JeodC:
> Some conflicts came up.

Rebased.

src/ship/Context.cpp
```diff
@@ -270,6 +271,21 @@ bool Context::InitControlDeck(std::shared_ptr<ControlDeck> controlDeck) {
...
 
+    // Bring up the SDL game-controller subsystem here rather than in osContInit, so controllers work
+    // in pre-game UI (e.g. navigating extraction prompts). osContInit still runs ControlDeck::Init(),
+    // which needs the game's controllerBits.
+    std::string controllerDb = LocateFileAcrossAppDirs("gamecontrollerdb.txt");
+    int mappingsAdded = SDL_GameControllerAddMappingsFromFile(controllerDb.c_str());
+    if (mappingsAdded >= 0) {
+        SPDLOG_INFO("Added SDL game controller mappings from \"{}\" ({})", controllerDb, mappingsAdded);
+    } else {
+        SPDLOG_ERROR("Failed to add SDL game controller mappings from \"{}\" ({})", controllerDb, SDL_GetError());
```
Malkierian: Not sure an error is warranted here, since the file is provided by ports, not LUS.

Malkierian (approved):

src/ship/Context.cpp
JeodC: That error isn't being informative about file presence, it's being informative about any potential errors adding the supposed mapping. I can downgrade it to warn though.

PR #1143, opened 2026-07-13
Fix Metal crash when LoadBlock uploads a texture as one giant line

serprex:
LoadBlock can record entire image as single line, so 256x256 texture becomes 65536x1 upload. If that exceeds GPU maximum texture size Metal crashes. When upload width is too wide, reshape. If buffer holds full image, use image's real dimensions; otherwise use tile-derived line size & height

See page 8 of https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf


serprex:
@JeodC

JeodC:
I don't have a Mac, but LGTM.

PR #1145, opened 2026-07-14
Metal: add PrewarmShaders to seed the shader cache in the background

quarrel07:
## Problem

On the Metal backend, the first use of each shader combination costs ~70 ms of synchronous MSL `newLibrary` compilation on the render path — a visible frame hitch plus an audio dropout. A game that touches ~40 combinations during boot → menus → gameplay accumulates a couple of seconds of stutter on every cold start (macOS caches compiled source per app, so warm relaunches are fine — but cache misses hit users after every game/OS update).

## Fix

- New optional `GfxRenderingAPI::PrewarmShaders(const uint64_t (*idPairs)[2], size_t pairCount)` virtual (default no-op), letting ports pass a baked list of shader ids captured from instrumented play.
- Metal implementation compiles each id's MSL source on a **detached background thread** (compile-and-discard; `MTL::Device` is thread-safe for resource creation). This seeds the OS shader cache, so the render thread's later compiles become sub-ms cache hits.
- Unknown/stale ids are harmless — anything not prewarmed still compiles lazily as before.

Verified in SpaghettiKart on macOS (Apple Silicon): with a cold shader cache, prewarming 42 combinations (84 variants) takes ~3.8 s in the background at boot, and instrumented runs show zero slow render-path compiles afterwards; the menu/race-start stutters are gone. A follow-up SpaghettiKart PR adds the port-side call.

### DIFF
--- a/include/fast/backends/gfx_metal.h
+++ b/include/fast/backends/gfx_metal.h
@@ -165,6 +165,7 @@ class GfxRenderingAPIMetal final : public GfxRenderingAPI {
     void UnloadShader(ShaderProgram* oldPrg) override;
     void LoadShader(ShaderProgram* newPrg) override;
     ShaderProgram* CreateAndLoadNewShader(uint64_t shaderId0, uint64_t shaderId1) override;
+    void PrewarmShaders(const uint64_t (*idPairs)[2], size_t pairCount) override;
     ShaderProgram* LookupShader(uint64_t shaderId0, uint64_t shaderId1) override;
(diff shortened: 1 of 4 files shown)


quarrel07:
Heads up, I pushed one more commit here. While stress testing on my fork I hit a rare startup crash caused by the prewarm thread sharing the shader builder's static state with the render thread. Made those thread_local, which fixes it. Details in the commit message.

Kenix3:
Any Mac people able to test this one out?

PR #1146, opened 2026-07-14
Normalize replacement-texture UVs by the original texture's dimensions

quarrel07:
## Problem

N64 games often point a tile at a region slightly **larger** than the texture it samples (e.g. a 27×19 tile over a 26×16 texture), relying on hardware CLAMP to repeat the edge texels. The interpreter derives UV normalization from the tile, which is correct for original assets — but hi-res replacement images (`TEX_FLAG_LOAD_AS_IMG`) span the **original texture's** texel space, so tile-derived normalization means the bottom/right of the replacement art is never sampled.

With 4K texture packs in SpaghettiKart this shows as every font glyph and menu banner visibly cropped at the bottom (~16% of each glyph missing); the same pack renders correctly in N64Recomp, which maps texels exactly.

## Fix

- `Fast::Texture` gains `OrigWidth`/`OrigHeight` (the replaced texture's dims; 0 = not a replacement). Ports set them when they load a replacement image — a follow-up SpaghettiKart PR does this from the binary texture header.
- `RawTexMetadata` carries them through all five metadata fill sites.
- `GfxSpTri1` normalizes UVs by the original dims for replacement draws; the sampler's clamp then covers the tile overhang exactly like N64 clamp did.
- **Framebuffer guard:** FB binds (`SelectTextureFb`) bypass the settimg/load path, leaving stale metadata in the TMEM slot. A new `RDP::fb_texture_selected` flag (set on FB selection, cleared on load/import of a regular texture) keeps the override away from FB-sourced draws — without it, FB copies (e.g. MK64's Luigi Raceway jumbotron feed) get shredded by a previous texture's metadata.

Verified in SpaghettiKart on macOS with the MK64-Reloaded 4K pack: glyphs/banners render full-height, original assets unchanged, and the jumbotron live feed is intact. No behavior change when no replacement is loaded (`OrigWidth == 0` short-circuits).

### DIFF
--- a/include/fast/interpreter.h
+++ b/include/fast/interpreter.h
@@ -229,6 +229,7 @@ struct LoadedVertex {
 
 struct RawTexMetadata {
     uint16_t width, height;
+    uint16_t orig_width = 0, orig_height = 0; // replaced texture's dims (hi-res packs); 0 = n/a
     float h_byte_scale = 1, v_pixel_scale = 1;
     std::shared_ptr<Fast::Texture> resource;
     Fast::TextureType type;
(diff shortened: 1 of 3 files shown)


quarrel07:
Rebased onto main after #1155 landed; no functional change.

PR #1151, opened 2026-07-17
Clamp masked over-loaded textures to their wrap period

JeodC:
A tile masked to NxM only addresses 2^mask texels, but some display lists load a larger block than the tile addresses (e.g. a 32x32 tile fed a 32x128 block). fast3d sized the texture from the loaded block, so UV normalization used the wrong dimensions and the surface stretched.

Clamp a masked axis back to its 2^mask wrap period. Only trust the mask when its period is at least the SetTileSize region, so a stale mask left by a framebuffer blit (which reuses a tile without a fresh SetTile) is ignored and the framebuffer isn't shrunk and tiled across the screen.

This restores the pre-#1090 clamp for the masked-WRAP over-load case without re-breaking the window-scroll textures #1090 was protecting.

### DIFF
--- a/include/fast/interpreter.h
+++ b/include/fast/interpreter.h
@@ -289,6 +289,7 @@ struct RDP {
         uint8_t fmt;
         uint8_t siz;
         uint8_t cms, cmt;
+        uint8_t masks, maskt; // mask exponents from SetTile; 2^mask is the WRAP/MIRROR period
         uint8_t shifts, shiftt;
         float uls, ult, lrs, lrt;
         uint16_t tmem; // 0-511, in 64-bit word units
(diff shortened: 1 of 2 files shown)


JeodC:
Before and after.

image
image

briaguya0 (approved):
i think this makes sense. i asked claude to take a look too and the only risk that it mentioned was
> It's a heuristic, not exact RDP emulation. A stale mask whose period happened to land in [render_region, loaded) could still mis-trigger a trim. Low risk given the reset-on-SetTile behavior and the s2dex paths passing 0, and it's strictly narrower than before, but it's the kind of thing that can only be confirmed by visual regression testing across the games this targets.

so that edge/corner case might be something to look out for in the future but definitely isn't something to block this on

PR #1154, opened 2026-07-19
Add Palette4bpp Support for Texture Loading into ImGui

Caladius:
This adds Palette4bpp texture type support for Fast3d, enabling textures of that format to be loaded into ImGui.

### DIFF
--- a/src/fast/Fast3dGui.cpp
+++ b/src/fast/Fast3dGui.cpp
@@ -740,7 +740,7 @@ void Fast3dGui::LoadGuiTexture(const std::string& name, const Fast::Texture& res
     mGuiTextures[name] = asset;
 }
 
-void Fast3dGui::LoadGuiTexture(const std::string& name, const std::string& path, std::string& palettePath, const ImVec4& tint) {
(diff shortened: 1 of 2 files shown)


src/fast/Fast3dGui.cpp
```diff
@@ -696,11 +740,11 @@ void Fast3dGui::LoadGuiTexture(const std::string& name, const Fast::Texture& res
     mGuiTextures[name] = asset;
 }
 
-void Fast3dGui::LoadGuiTexture(const std::string& name, const std::string& path, const ImVec4& tint) {
+void Fast3dGui::LoadGuiTexture(const std::string& name, const std::string& path, std::string& palettePath, const ImVec4& tint) {
```
Malkierian:
```suggestion
void Fast3dGui::LoadGuiTexture(const std::string& name, const std::string& path, const std::string& palettePath, const ImVec4& tint) {
```

Caladius:
replaced by https://github.com/Kenix3/libultraship/pull/1157

PR #1155, opened 2026-07-20
Round scaled viewport and scissor edges outward

quarrel07:
When AdjustVIewportOrScissor scales a rectangle into window pixels, x and width are scaled independently into integer fields, so both truncate. The converted right and bottom edges can land up to two pixels short of the true edge. With the game's fullscreen scissor that clips every scene draw slightly early, which shows up as an unrendered column at the window edge at some aspect ratios (only passes with a different scissor state still paint there, so it reads as a black strip that a few elements bleed into).

Fix: compute both edges in float and round them outward, so a scaled rectangle always covers the region the original rectangle did.

Found and play-verified in SpaghettiKart on macOS (the black edge column is gone at the aspect ratios that showed it), but the code path is shared by every port and backend.

### DIFF
--- a/src/fast/interpreter.cpp
+++ b/src/fast/interpreter.cpp
@@ -2200,10 +2200,20 @@ void Interpreter::AdjustVIewportOrScissor(XYWidthHeight* area) {
             area->y = mNativeDimensions.height - area->y;
         }
 
-        area->width *= RATIO_X(mActiveFrameBuffer, mCurDimensions);
-        area->height *= RATIO_Y(mActiveFrameBuffer, mCurDimensions);
-        area->x *= RATIO_X(mActiveFrameBuffer, mCurDimensions);
-        area->y *= RATIO_Y(mActiveFrameBuffer, mCurDimensions);
+        // Scale by computing both edges and rounding them outward: the fields
+        // are integers, so scaling width/x independently truncates and can
+        // leave the right/bottom edge up to 2px short of the true edge
+        // (visible as an unrendered column at the window edge).
+        {
+            const float rx = RATIO_X(mActiveFrameBuffer, mCurDimensions);
+            const float ry = RATIO_Y(mActiveFrameBuffer, mCurDimensions);
+            const float x2 = (area->x + (float)area->width) * rx;
+            const float y2 = (area->y + (float)area->height) * ry;
+            area->x = (int16_t)floorf(area->x * rx);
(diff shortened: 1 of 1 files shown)


MegaMech:
Good fix, but what about the lower half of this function?

image

Malkierian:
Really quick, what was the use-case that ended in this bug's discovery?

quarrel07:
@MegaMech 
> Good fix, but what about the lower half of this function?
> 
> image

Good catch, same truncating multiplies down there. Pushed https://github.com/Kenix3/libultraship/commit/ff30037ef0b7d1d8fdf61c5ac3d4a189a04e583a: the resize branch now rounds its edges outward the same way. To be fair, the window-edge symptom that led me here goes through the top path, so I don't have a repro exercising the framebuffer branch, but it's the identical arithmetic on the same integer fields.

quarrel07:
@Malkierian
> Really quick, what was the use-case that ended in this bug's discovery?

I noticed it when doing some QA on the SpaghettiKart macOS port: the last few pixel columns at the window edge stayed unrendered. Half of that turned out to be the game's own sky quad falling short (fixed port side in https://github.com/HarbourMasters/SpaghettiKart/pull/717), and the other half was the fullscreen scene scissor stopping a few pixels early, which traced back to the double truncation here.

src/fast/interpreter.cpp
```diff
@@ -2265,10 +2275,17 @@ void Interpreter::AdjustVIewportOrScissor(XYWidthHeight* area) {
         area->y = mActiveFrameBuffer->second.orig_height - area->y;
 
         if (mActiveFrameBuffer->second.resize) {
-            area->width *= RATIO_X(mActiveFrameBuffer, mCurDimensions);
-            area->height *= RATIO_Y(mActiveFrameBuffer, mCurDimensions);
-            area->x *= RATIO_X(mActiveFrameBuffer, mCurDimensions);
-            area->y *= RATIO_Y(mActiveFrameBuffer, mCurDimensions);
+            // Same outward rounding as above: the independent multiplies
+            // truncate per field and can leave the right/bottom edge short
+            // inside a resized framebuffer.
```
Malkierian: This comment can merely say "Same as above". The explanation is already there.

quarrel07: Done! Shortened in 07d2d1fd.

Malkierian (approved):

Kenix3 (approved):

PR #1156, opened 2026-07-20
fast3d: fix silently skipped CI texture uploads leaving poisoned cache entries

quarrel07:
Since #1043, ImportTextureCi8 bails unless both palette halves are non-null, and ImportTextureCi4 bails if its half is null. But partial TLUT loads are completely normal on N64. Plenty of games load 46 to 76 entries into half 0 and never touch half 1. Every CI8 draw using such a palette now skips its upload entirely.

Worse, the bail happens after TextureCacheLookup has already inserted the cache entry and bound a texture id, so the entry stays in the map pointing at a texture that never got pixel data. Fresh id: draws black. Id that's been through the free list: draws whatever texture last owned it. In SpaghettiKart's harbour intro that meant the ship's chrome engine rendering black, and random foreign art (minimap, UI) appearing on the ship after asset pack toggles recycled ids.

Two commits:
1. Restore N64 TMEM semantics for the palette area: it always exists, and an unloaded half is stale data (zeros here), never an error. palettes[0/1] now permanently point at the zeroed staging arrays and both bails are removed. Indexing into an unloaded half decodes from zeros, same as hardware reading stale TMEM.
2. Any import that fails to upload (null address, zero sized load, unsupported format) now erases the entry it just inserted, so the cache can never hand out a bound but never filled texture.

Verified in SpaghettiKart on macOS Metal: the harbour scene matches its pre #1043 appearance again and the toggle corruption is gone. The corruption reproduced on both Metal and OpenGL before the fix.


quarrel07:
Rebased onto main after #1155 landed; no functional change.

PR #1162, opened 2026-07-22
Lock the ArchiveManager file index for concurrent lookups

quarrel07:
ArchiveManager's hash-to-archive and hash-to-path maps are read and written with no synchronization, and the hot lookup path uses `operator[]`, which inserts an empty entry into the map on every miss. Misses happen constantly at runtime (every alt-asset probe, every texture pack extension probe), and ports increasingly call into the resource system from worker threads while the render thread does the same. Two threads missing at once can rehash the map under a concurrent reader, which is undefined behavior.

This isn't theoretical: in the SpaghettiKart macOS port, a corrupted lookup returned not-found for a kart palette that exists, ResourceManager negative-cached that result permanently, and the render thread crashed copying the palette from a null pointer.

The fix: `LoadFile` and `GetArchiveFromFile` use `find()` instead of `operator[]`, and a `shared_mutex` guards the index (shared for lookups, exclusive for `AddArchive`, `WriteFile`, and the VFS rebuild). `ResetVirtualFileSystem` now also clears `mDirectories`, which previously kept stale entries across an archive reload.

Verification, all on macOS Apple Silicon: three new tests in the existing gtest suite. On unmodified code, the two index-pollution tests fail deterministically (a missed lookup made `HasFile` report the path as present) and the concurrency test produces a few hundred ThreadSanitizer data-race reports pointing at these two maps. With the fix, the full suite passes with TSAN clean. The fix has also been running in the SpaghettiKart macOS build that hit the original crash.


quarrel07:
Adding a data point from the wild: a SpaghettiKart player on the stock Windows 1.0.0 build posted a crashlog in the spaghettikart-development Discord channel. Access violation about 100ms after race setup started (they had just hit restart race), reading a near-NULL address with no symbols in the trace. They'd finished several races fine before it hit, and it didn't reproduce after relaunching. That intermittent race-start timing matches the concurrent lookup race this PR locks against, with the failed lookup feeding game-side memcpy calls (kart palette / wheel tlut loads in SpaghettiKart's case). Here's the full log for reference if helpful:
[Crash Log.txt](https://github.com/user-attachments/files/30434128/Crash.Log.txt)

include/ship/resource/archive/ArchiveManager.h
```diff
@@ -243,6 +244,9 @@ class ArchiveManager : public Component {
     std::vector<std::shared_ptr<Archive>> mArchives;
     std::vector<uint32_t> mGameVersions;
     std::unordered_set<uint32_t> mValidGameVersions;
+    // Guards the virtual-filesystem index below: lookups run concurrently on
```
Kenix3: These comments are unnecessary

src/ship/resource/archive/ArchiveManager.cpp
```diff
@@ -48,7 +48,19 @@ std::shared_ptr<File> ArchiveManager::LoadFile(const std::string& filePath) {
 }
 
 std::shared_ptr<File> ArchiveManager::LoadFile(uint64_t hash) {
-    auto archive = mFileToArchive[hash];
+    // find(), not operator[]: misses are hot (alt-asset and texture-pack
```
Kenix3: Comments are unnecessary here

Kenix3 (changes requested):

include/ship/resource/archive/ArchiveManager.h
quarrel07: Removed, thanks.

src/ship/resource/archive/ArchiveManager.cpp
quarrel07: Gone as well. There are a few explanatory comments left in the new test file if you'd like those trimmed too, just say the word.

PR #1163, opened 2026-07-22
Recover saved window backend by name when saved IDs go stale

### LINKED ISSUE #1153: WindowBackend renumbering silently switches existing macOS configs from Metal to OpenGL
#1097 moved the `WindowBackend` enum into `fast/Fast3dWindow.h` and renumbered it from implicit 0-based (`DX11=0, OPENGL=1, METAL=2`) to explicit 1-based (`DX11=1, OPENGL=2, METAL=3`).

Configs written by older libultraship store the backend as an integer (`Window.Backend.Id`). A macOS user's existing config has `Id: 2`, meaning Metal under the old numbering. Current `Window.cpp` reads that as OpenGL, runs OpenGL, and re-saves `{"Id": 2, "Name": "OpenGL"}` on exit, making the downgrade permanent and invisible. Old DX11 configs (`Id: 0`) hit a similar mismatch on Windows.

**Repro** (SpaghettiKart, current libultraship, Apple Silicon):
1. In the config, set `"Window": { "Backend": { "Id": 2, "Name": "Metal" } }` (what an old install wrote for Metal)
2. Boot, open the backend dropdown: it shows OpenGL
3. Quit; the config now reads `{"Id": 2, "Name": "OpenGL"}`

**Impact:** measured with frame-time instrumentation in SpaghettiKart on an M4 Mac, same session on each backend: Metal runs a clean 30 Hz cadence (33.3 ms median, ~2-4% of frames over 40 ms); the silently-selected OpenGL runs with ~33% of frames over 40 ms plus audible audio stutter. It presents as "the game got choppy after upgrading" with nothing in the UI suggesting why.

Affects both main and port-maintenance (the renumbering commit is on both). A possible fix is to consult `Window.Backend.Name` when it disagrees with the id, or to migrate the stored id once. Happy to submit a PR either way, let me know what approach you'd prefer.

quarrel07:
Fixes #1153. Same fix as #1161, which I closed before it was properly verified; it now comes with tests and one hardening commit.

The config stores both `Window.Backend.Id` and `Window.Backend.Name`, but only the ID was read back. The WindowBackend renumbering shifted every ID, so existing configs silently changed renderer: a saved Metal config re-resolved as OpenGL, and a saved DirectX 11 config fell out of range entirely. `GetSavedWindowBackend` now resolves the saved name first through a new `GetWindowBackendIdByName` virtual and heals the stored ID when the two disagree. Unknown or missing names fall through to the existing ID validation unchanged.

Second commit: unit tests covering the case table with a real Context and Fast3dWindow against a temp config. The stale-ID recovery and heal is exercised with OpenGL on every platform and with the Metal case from the original report on macOS; matching configs pass through untouched; unknown names fall back to the old path. Making never-initialized windows destructible required null guards in `Interpreter::Destroy` and the Context destructor.

Third commit: startup no longer persists the resolved backend. Previously every launch wrote the resolved ID and name back to the config, which is exactly how the renumbering bug permanently overwrote saved Metal configs on first launch: users who ran an affected build once have a self-consistent wrong config that no migration can distinguish from a deliberate choice. Those users will need to re-pick their renderer one time. With this change, resolution bugs can misread a config but can never destroy it, so this class of loss can't recur.

Verified via the new unit tests on macOS Apple Silicon (5 pass, full suite 433/433). Not runtime-tested in a full port; happy to runtime-verify on a port before merge if you'd like.

### DIFF
--- a/include/fast/Fast3dWindow.h
+++ b/include/fast/Fast3dWindow.h
@@ -91,6 +91,7 @@ class Fast3dWindow : public Ship::Window {
     const char* GetKeyName(int32_t scancode) override;
 
     std::string GetWindowBackendName() override;
+    int32_t GetWindowBackendIdByName(const std::string& name) override;
 
     void SetCurrentDimensions(uint32_t width, uint32_t height) override;
     void SetCurrentDimensions(uint32_t width, uint32_t height, int32_t posX, int32_t posY) override;
--- a/include/ship/window/Window.h
+++ b/include/ship/window/Window.h
@@ -217,6 +217,15 @@ class Window : public Component {
      * override this to return backend-specific names such as "OpenGL", "Metal", or "DirectX 11".
      */
     virtual std::string GetWindowBackendName();
+    /**
+     * @brief Resolves a backend name (as produced by GetWindowBackendName) back to its ID.
+     * Subclasses that override GetWindowBackendName should override this with the reverse
+     * mapping. Used to recover saved configs when backend IDs are renumbered between
+     * versions: the stored name identifies the backend even when the stored ID has gone
+     * stale. Returns -1 when the name is unknown.
+     * @param name Backend name to resolve.
+     */
+    virtual int32_t GetWindowBackendIdByName(const std::string& name);
     /** @brief Returns the list of backends available on this platform. */
     std::shared_ptr<std::vector<int32_t>> GetAvailableWindowBackends();
     /**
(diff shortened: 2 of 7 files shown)


quarrel07:
Reimplemented for 2.0.0: after the SDL3 migration this couldn't rebase, so it's a fresh single commit against main with the same design as before. Name first recovery in GetSavedWindowBackend (with config healing when the ID is stale), backend changes only persist when deliberate so startup resolution never rewrites the config, and Fast3dWindow supplies the reverse name to ID mapping.

The tests got simpler in the move: the recovery logic now lives on the base Window class, so a minimal fake window with an injected config covers all the cases on every platform, including the original incident (a stale ID that's still valid but names the wrong renderer). One small extra fix came out of that: Gui::ShutDownImGui now tolerates a window that was constructed but never initialized, since the destructor hits it.

PR #1121, opened 2026-06-07
fix: round interpolated texture tile sizes

banteg:
## Summary

This fixes interpolated texture tile dimensions being truncated after `uls/ult/lrs/lrt` interpolation.

The bug showed up on macOS/Apple Silicon as animated water/lava textures visibly jittering when interpolation is active. In the Fire Temple lava repro, one replay phase derived a different rendered texture window from the interpolated tile coordinates:

```text
replay=1
tile0=(0.000,6.333,124.000,130.333)
renderSize0=(32, 31)
texSize0=(32, 31)
```

Adjacent phases for the same surface stayed `32x32`, so the visible symptom was an alternating `32x32` / `32x31` texture window.

```
replay0=(0.000,6.000,124.000,130.000) renderSize0=(32, 32)
replay1=(0.000,6.333,124.000,130.333) renderSize0=(32, 31) # bugged!
replay2=(0.000,6.667,124.000,130.667) renderSize0=(32, 32)
```

## Root Cause

Several paths in `src/fast/interpreter.cpp` derive rendered/imported tile dimensions with:

```cpp
(high - low + 4) / 4
```

Those results were then cast directly to integer types. With interpolated float coordinates, a texel count that is logically `32.0` can land slightly below the integer boundary, so truncation turns it into `31`.

That makes the texture clamp/import and draw-time texture size calculations disagree across interpolation phases, producing the observed water/lava animation jitter.

## Fix

Add `GetTileSizeFromCoordinates()` and use it for the tile-size calculations that derive dimensions from interpolated tile coordinates. It keeps non-positive sizes clamped to `0` and rounds the derived texel count with `lroundf`.

## Related Links

- Fixes Kenix3/libultraship#1119
- Shipwright report: HarbourMasters/Shipwright#6666

## Validation

- `cmake -S . -B build -DLUS_BUILD_TESTS=OFF`
- `cmake --build build --target libultraship -j 8`
- `cmake -S . -B build -DLUS_BUILD_TESTS=ON`
- `cmake --build build --target lus_tests -j 8`
- `ctest --test-dir build --output-on-failure`

PR #1164, opened 2026-07-23
fix: round interpolated texture tile sizes

### LINKED ISSUE #1119: Interpolated texture tile sizes can truncate and cause water/lava jitter
Linked Shipwright report: https://github.com/HarbourMasters/Shipwright/issues/6666

I investigated the Shipwright water/lava texture animation jitter locally and the repro points at LUS-side tile-size derivation rather than frame pacing or skipped interpolation.

In the Fire Temple lava case, the interpolation replay cycle stayed regular (`0 -> 1 -> 2 -> 0 ...`), but one replay phase derived a different rendered texture window size from the interpolated tile coordinates.

Bad phase example:

```text
replay=1
tile0=(0.000,6.333,124.000,130.333)
renderSize0=(32, 31)
texSize0=(32, 31)
```

Adjacent phases for the same lava surface stayed `32x32`:

```text
replay=0 tile0=(0.000,6.000,124.000,130.000) renderSize0=(32, 32)
replay=2 tile0=(0.000,6.667,124.000,130.667) renderSize0=(32, 32)
```

So the visual jitter appears to come from alternating between a `32x32` and `32x31` texture window. The intended height is still 32 texels, but the current conversion effectively truncates `(lrt - ult + 4) / 4` after interpolation. With interpolated float values like `6.333` and `130.333`, precision can land just below `32.0`, and the cast becomes `31`.

The conversions I found are in `src/fast/interpreter.cpp`, in the code paths that derive rendered/imported tile dimensions from `uls/ult/lrs/lrt`, including:

- `ImportTextureRgba16`
- `ImportTextureRgba32`
- `ImportTextureCi4`
- `ImportTextureCi8`
- `GfxSpTri1`

A minimal local fix that made the lava case stable was to round the derived texel count instead of truncating it:

```cpp
static uint32_t GetTileSizeFromCoordinates(float low, float high) {
    float size = (high - low + 4.0f) / 4.0f;
    if (size <= 0.0f) {
        return 0;
    }
    return static_cast<uint32_t>(lroundf(size));
}
```

Then use that helper instead of directly casting `(high - low + 4) / 4` to `uint32_t`/`int32_t` in those tile-size calculations.

I tested this locally in Shipwright; the lava scene stopped visibly alternating between good and bad texture phases. The local `soh` target built successfully after the change. I did not need to change interpolation scheduling or texture cache behavior for this specific symptom.

Additional Shipwright-side diagnostic comment: https://github.com/HarbourMasters/Shipwright/issues/6666#issuecomment-4638437420

banteg:
## Summary

This is the `port-maintenance` backport of #1121, requested in https://github.com/Kenix3/libultraship/pull/1121#issuecomment-5054224556.

It fixes interpolated texture tile dimensions being truncated after `uls/ult/lrs/lrt` interpolation. The bug showed up on macOS/Apple Silicon as animated water/lava textures visibly jittering when interpolation is active.

## Root cause

Several paths in `src/fast/interpreter.cpp` derive rendered/imported tile dimensions with:

```cpp
(high - low + 4) / 4
```

Those results were cast directly to integer types. With interpolated float coordinates, a texel count that is logically `32.0` can land slightly below the integer boundary, so truncation turns it into `31`.

## Fix

Add `GetTileSizeFromCoordinates()` and use it for tile-size calculations derived from interpolated coordinates. It keeps non-positive sizes clamped to `0` and rounds the derived texel count with `lroundf`.

Fixes #1119.

## Validation

- The original commit cherry-picks cleanly onto `port-maintenance`.
- `git diff --check upstream/port-maintenance..HEAD`
- `src/fast/interpreter.cpp` syntax-compiles successfully with the project’s CMake-generated C++20 flags and declared dependency headers (`c++ ... -fsyntax-only`).

### DIFF
--- a/src/fast/interpreter.cpp
+++ b/src/fast/interpreter.cpp
@@ -545,6 +545,14 @@ static uint32_t GetEffectiveLineSize(uint32_t lineSizeBytes, uint32_t fullImageL
     return tileLineSizeBytes;
 }
 
+static uint32_t GetTileSizeFromCoordinates(float low, float high) {
+    float size = (high - low + 4.0f) / 4.0f;
+    if (size <= 0.0f) {
+        return 0;
+    }
+    return static_cast<uint32_t>(lroundf(size));
+}
+
 void Interpreter::ImportTextureRgba16(int tile, bool importReplacement) {
     const RawTexMetadata* metadata = &mRdp->loaded_texture[mRdp->texture_tile[tile].tmem_index].raw_tex_metadata;
     const uint8_t* addr =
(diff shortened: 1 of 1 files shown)


Eblo:
Seeing the aforementioned issue in 2ship following a recent LUS bump, can confirm this fixes it for me locally.

Malkierian (approved):

PR #1166, opened 2026-07-24
Fix NullAudioPlayer::Buffered() returning 0 (audio-thread spin/hang) -- port-maintenance

bassdr:
It discards all audio, so reporting 0 made the producer spin feeding it forever and starve the gfx thread. Return the desired buffered count, as the header already documents.


bassdr:
wrong branch

PR #1168, opened 2026-07-24
Resolve .meta aliases by archive priority

### LINKED ISSUE #984: `.meta` files no longer work
### repro steps

* create archive with `myResource.meta` that points to `myResourceData.xml`
* try to load `myResource`
* see it fail because `myResource` does not exist

### investigation 

i did some testing (using the changes from <https://github.com/HarbourMasters/Shipwright/pull/3952/changes/f5d2e583444de41a1a43a00b9c3fca2bf782dc2d>) and it looks like it's failing here: <https://github.com/Kenix3/libultraship/blob/8c55f607f2249f3ac696fc0f7277553fe3ce75a6/src/ship/resource/ResourceManager.cpp#L140-L146>

after checking blame i'm thinking it's likely <https://github.com/Kenix3/libultraship/pull/790> that broke it

@Kenix3  it looks like you merged that one, do you know how we're supposed to handle the use case of
* game requests `myResourceDL`
* archive doesn't have a file named `myResourceDL`, just has a `myResourceDL.meta` that points to `myXMLToUseForMyResourceDL.xml`

because right now we just get `nullptr` from `LoadFileProcess` so `LoadResourceProcess` returns `nullptr` and sets the resource to error

___

it's still failing for me, which makes sense because it still isn't seeing `myResourceDL` when it calls into `LoadFileProcess` from `LoadResourceProcess`, it just moves the failure to

```cpp
    auto archive = mFileToArchive[hash];
    if (archive == nullptr) {
        return nullptr;
    }

    return archive->LoadFile(hash);
}
```

___

https://github.com/Kenix3/libultraship/pull/983 is a workaround, it solves the problem if we have *both* a `myResource` file and a `myResource.meta` file (like when using a mod archive to replace something), but not for the use case in the repro steps
because `myResourceDL` isn't hashed (it is when we have the `.meta` stripping), as opposed to in

```cpp
std::shared_ptr<File> O2rArchive::LoadFile(const std::string& filePath) {
    if (mZipArchive == nullptr) {
        SPDLOG_TRACE("Failed to open file {} from zip archive {}. Archive not open.", filePath, GetPath());
        return nullptr;
    }

    auto zipEntryIndex = zip_name_locate(mZipArchive, filePath.c_str(), 0);
    if (zipEntryIndex < 0) {
        SPDLOG_TRACE("Failed to find file {} in zip archive  {}.", filePath, GetPath());
        return nullptr;
    }
```

when it just can't find `myResourceDL` in the zip

### LINKED ISSUE #1165: `.meta` alias should fall back to the real same-path asset when its target is missing
# `.meta` alias should fall back to the real same-path asset when its target is missing

### Summary
When a resource is requested and a `<path>.meta` alias exists that repoints it to another
resource, but that target resource is **not present**, LUS should fall back to loading the
real asset at `<path>` (if one exists) instead of failing the load.

This is separate from #984, which is about `.meta` aliases resolving at all. This request is
about the fallback behavior when an alias is present but its target isn't.

### Motivation / use case
Cross-game link-mod compatibility between SoH (OoT) and 2Ship (MM), with **no changes to the mod**:

- `oot.o2r` requests `gLinkChildSkel` (OoT has both adult and child Link).
- `mm.o2r` requests `gLinkHumanSkel` (MM's human-form Link).
- A modder ships a plain SoH-style `mod.o2r` that contains only `gLinkChildSkel`.

To make that unmodified mod load in 2Ship, 2Ship ships the alias itself in its **base
archive** (a `gLinkHumanSkel.meta` pointing at `gLinkChildSkel`, or equivalent hardcoded
init data — a 2Ship-side change, not an LUS one). Then:

- **Mod loaded:** 2Ship's request for `gLinkHumanSkel` resolves via the alias to the mod's
  `gLinkChildSkel`. The mod needs zero `.meta` files and isn't touched.
- **No mod loaded (vanilla boot):** `gLinkChildSkel` doesn't exist, so the alias target is
  absent — 2Ship must fall back to loading its own real `gLinkHumanSkel` (vanilla MM Link).

That vanilla boot case is what this issue is about: shipping an alias in the base archive
must not break the game when the alias target isn't present.

### Current behavior
`.meta` resolution happens in `ResourceLoader::LoadResource`
(`src/ship/resource/ResourceLoader.cpp:203-212`): it reads `<path>.meta`, gets the target
path from it, and calls `LoadFileProcess(initDataFromMetaFile->Path)` for the target. If the
target file doesn't exist, `fileToLoad` becomes `nullptr` and the load fails at
`ResourceLoader.cpp:215-218` — there is no fallback to the real file at the original
`<path>`.

### Proposed / expected behavior
When a `<path>.meta` alias resolves to a target that cannot be loaded:
- If a real asset exists at `<path>`, load that instead (the alias is treated as
  "prefer the target, otherwise use me").
- If neither the target nor a real `<path>` asset exists, fail as today.

### Acceptance criteria
- [ ] `<path>` real asset present, `<path>.meta` present, target present → **target loads** (alias wins).
- [ ] `<path>` real asset present, `<path>.meta` present, target **absent** → **real `<path>` loads** (graceful fallback), rather than erroring.
- [ ] `<path>` real asset absent, `<path>.meta` present, target absent → fails as today.
- [ ] Assumes / depends on #984 being resolved first.

### Notes
- A minimal repro can be built with the logo download asset to exercise both the
  target-present and target-absent cases.
- Related: #984 (aliases resolving at all), #983 (partial fix).

briaguya0:
## Summary

Reworks `.meta` alias resolution so a request for `X` loads the **highest-priority provider**
among:

- the real asset at `X`, and
- the target of any `X.meta` alias,

ranked by the archive each **asset** lives in (last-loaded archive wins; ties go to the alias).

Fixes #984
Resolves #1165 

- **#984** — `.meta` aliases stopped resolving. Root cause: `Archive::IndexFile` stripped the
  `.meta` suffix, conflating a real `foo` with its `foo.meta`, so the loader's `foo.meta` lookup
  could never resolve (and a `.meta`-only resource was unreachable).
- **#1165** — a `.meta` alias must fall back to the real asset when its target is missing, and (the
  cross-game case) a lower-priority archive's `.meta` whose target lives in a *higher*-priority
  archive must win over an intervening real asset.

## Model

`X.meta → Y` makes `Y` an alternate provider of `X`. Resolving `X` compares the real `X` against
the aliased `Y` by the priority of the archive where each **asset** lives (not where the `.meta`
lives), and loads the higher-priority one; ties go to the alias. "Fall back to the real asset" is
simply the case where the alias's target doesn't exist, so the real asset is the only provider
left.

## Changes

1. **`Archive::IndexFile`** — index every file under its literal name (no `.meta` stripping), so a
   real `foo` and its `foo.meta` are distinct index entries.
2. **`Archive::Get/SetPriority` + `ArchiveManager::GetFilePriority`** — O(1) load-order priority of
   the archive that owns a given path.
3. **`ResourceManager::LoadResourceProcess`** — only bail when neither the real file nor a `.meta`
   exists, so a resource that exists only as a `.meta` reaches the loader.
4. **`ResourceLoader::ResolveMetaAlias` / `LoadResource`** — pick the higher-priority provider
   (real asset vs alias target); ties go to the alias.

## Testing

Verified end-to-end in Ship of Harkinian with a dedicated test kit (companion testing PR:
https://github.com/HarbourMasters/Shipwright/pull/6977) that aliases the boot ship-logo DisplayList and recolors the assets so the loaded
one is obvious. It covers the full matrix:

- single-mod cases: target-present, fallback, meta-only, no-`.meta` regression;
- layered archive-priority cases: the cross-game "flagship" (a `.meta` below a real, whose target
  is higher, wins), "real above the alias target" (real wins), and "target below the `.meta`"
  (resolved by normal global lookup).

Baseline (stock engine) and after-fix results were both captured; every case matches, with five
result flips vs. baseline proving the fix.

## Notes

- Targets `port-maintenance`.
- Minor behavior change from the `Archive::IndexFile` change: a resource that exists *only* as `X.meta` is no longer in the
  base-name index, so `HasFile(X)` returns false for it. Resolution still works (the loader checks
  `X.meta`), but callers that gate on existence before loading would skip such a resource.

### DIFF
--- a/include/ship/resource/ResourceLoader.h
+++ b/include/ship/resource/ResourceLoader.h
@@ -111,6 +111,16 @@ class ResourceLoader {
     std::shared_ptr<ResourceInitData> ReadResourceInitData(const std::string& filePath,
                                                            std::shared_ptr<File> metaFileToLoad);
 
+    /**
+     * @brief Resolves a `.meta` alias at filePath, if one wins over the real asset.
+     * @param filePath   Virtual path being loaded.
+     * @param fileToLoad In/out: the highest-priority real asset at filePath (may be null). On a
+     *                   successful alias, this is replaced with the aliased target's file.
+     * @return The alias target's init data if its archive is equal-or-higher priority than the
+     *         real asset's; otherwise nullptr (fileToLoad left unchanged).
+     */
+    std::shared_ptr<ResourceInitData> ResolveMetaAlias(const std::string& filePath, std::shared_ptr<File>& fileToLoad);
+
     /** @brief Creates a ResourceInitData with default/zeroed fields. */
     static std::shared_ptr<ResourceInitData> CreateDefaultResourceInitData();
 
--- a/include/ship/resource/archive/Archive.h
+++ b/include/ship/resource/archive/Archive.h
@@ -156,6 +156,15 @@ class Archive : public std::enable_shared_from_this<Archive> {
     bool IsLoaded();
 
     /**
+     * @brief Load-order priority assigned by the ArchiveManager when the archive is mounted
+     * (higher = higher priority; the last-loaded archive wins conflicts). -1 if unset.
+     */
+    int32_t GetPriority();
+
+    /** @brief Sets the load-order priority. Called by the ArchiveManager. */
+    void SetPriority(int32_t priority);
+
+    /**
      * @brief Opens the underlying archive file or directory for reading.
      * @return true on success.
      */
@@ -194,6 +203,7 @@ class Archive : public std::enable_shared_from_this<Archive> {
     bool mIsChecksumValid;
     bool mHasGameVersion;
     uint32_t mGameVersion;
+    int32_t mPriority = -1;
     ArchiveManifest mManifest;
     std::string mPath;
     std::shared_ptr<std::unordered_map<uint64_t, std::string>> mHashes;
--- a/include/ship/resource/archive/ArchiveManager.h
+++ b/include/ship/resource/archive/ArchiveManager.h
@@ -144,6 +144,14 @@ class ArchiveManager {
     GetArchiveFromFile(const std::string& filePath); // Retrieves a ptr to the archive that the asset is inside of
 
     /**
+     * @brief Load-order priority of the archive that owns the given file (higher = higher
+     * priority; the last-loaded archive wins). Used to rank a real asset against an alias target.
+     * @param filePath Virtual path of the file.
+     * @return Priority index, or -1 if no loaded archive has the file.
+     */
+    int32_t GetFilePriority(const std::string& filePath);
+
+    /**
      * @brief Lists virtual paths of all files matching the given search mask across all archives.
      * @param searchMask Glob pattern (empty = list everything).
      * @return Sorted, deduplicated list of matching paths.
--- a/src/ship/resource/ResourceLoader.cpp
+++ b/src/ship/resource/ResourceLoader.cpp
@@ -182,6 +182,29 @@ std::shared_ptr<ResourceInitData> ResourceLoader::ReadResourceInitData(const std
     return initData;
 }
 
+// Position File::BufferOffset at the resource body. A currently-headed binary resource keeps an OTR
+// header ahead of its body; when init data was supplied externally (a caller or a `.meta`) rather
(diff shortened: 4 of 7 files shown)


Kenix3 (approved):
LGTM

PR #1174, opened 2026-07-28
Component system v2

### LINKED ISSUE #764: `Context` should have a `Component` system and a list of `Component`s instead of a bunch of different classes.

Kenix3:
Resolves https://github.com/Kenix3/libultraship/issues/764

Run on Ghostship: https://github.com/HarbourMasters/Ghostship/pull/223


briaguya0:
> 🤖 **Note:** this comment was written by Claude (Claude Code), on @briaguya0's behalf.

Cross-referencing the earlier review: the pending review left on the now-closed #1148 flagged 11 issues in the first iteration of the component system — [see the inline comments here](https://github.com/Kenix3/libultraship/pull/1148#pullrequestreview-4761586308).

I re-checked each of those against this PR (`component_system_v2`), and **all 11 are addressed here.** Several fixes even mirror the review notes in-code (weak parent refs, the "hold the Logger aside and drop it last" teardown, the `visited`-set cycle guards). Mapping:

| # | Finding (from #1148) | Status in #1174 | Evidence |
|---|----------------------|-----------------|----------|
| 1 | **[high]** parent↔child strong-ref cycle → leak, `~Context`/config-save never runs | ✅ Fixed | Parents now stored non-owning: `ParentComponentList = BasicComponentList<std::weak_ptr<Component>>`; children stay owning. `PartList` gained configurable weak/shared storage. |
| 2 | **[high]** `TickableComponent::mContext` strong `shared_ptr<Context>` cycle | ✅ Fixed | The strong `shared_ptr<Context>` member is gone; it reuses `Part`'s `weak_ptr<Context>` context. |
| 3 | **[med]** AB/BA lock-order inversion (reciprocal add under held mutex) | ✅ Fixed | `PartList::Add` closes the mutex scope before calling `Added()`/`OnAdded()` (`PartList.h` ~L422–438), so the reciprocal add runs unlocked. |
| 4 | **[med]** `PropagateContextDown` unbounded recursion on a cycle | ✅ Fixed | Now takes `std::unordered_set<uint64_t>& visited` and bails on re-visit (`ComponentList.cpp` ~L14–18). |
| 5 | **[med]** `TickableList::Sort()` never called → tick ordering dead | ✅ Fixed | `TickableList::Added()` now calls `Sort()` (`TickableList.h` ~L48–50). |
| 6 | **[med]** Logger destroyed first → components log after spdlog shutdown | ✅ Fixed | `~Context` holds the Logger aside, `Remove(true)` for everything else, then `logger.reset()` last (`Context.cpp` ~L106–123). |
| 7 | **[low]** `ToTreeString` no cycle guard | ✅ Fixed | Delegates to `ToTreeStringImpl(depth, visited)` with a visited set (`Component.cpp` ~L34–48). |
| 8 | **[low]** BFS queues raw `Component*` from a temp snapshot (dangling risk) | ✅ Fixed | All six BFS helpers now use `std::queue<std::shared_ptr<const Component>>` and push the shared_ptr directly. |
| 9 | **[low]** `mIsInitialized` non-atomic data race | ✅ Fixed | `std::atomic<bool> mIsInitialized{false}` + `mInitMutex` under `COMPONENT_THREAD_SAFE` (`Component.h` ~L214–216). |
| 10 | **[low]** `Part::mContext` weak_ptr read/write race | ✅ Fixed | Added `mutable std::mutex mContextMutex` guarding `mContext` (`Part.h` ~L91–95). |
| 11 | **[low]** unguarded `SPDLOG_INFO` in `~ResourceManager` | ✅ Fixed | Now wrapped in `if (spdlog::default_logger())` (`ResourceManager.cpp` ~L81). |

Since this is a fresh reimplementation rather than a rebase of #1148 (the events layer was also rewritten — `Events`, `ListenerAction`, `EventActionList`), the above only confirms the previously-flagged issues are resolved; it is not a fresh review of code new to this PR.

PR #1175, opened 2026-07-28
Add opt-in memoization of texture path resolution

JeodC:
Profiling Lighthouse in heavy scenes (max draw distance, high interpolation targets) showed OTR texture path resolution as the largest avoidable cost in the display list walk: ~10 ms/frame in `LoadResourceProcess` covering string construction, hash, cache mutex, and map lookup, all re-resolving the same handful of textures the scene binds every frame. The answer only changes when resources reload, so everything after the first bind is repeated work.

The per-bind cost is identical across ports; Lighthouse surfaces it because BK's open, prop-dense worlds under a draw-distance multiplier produce an order of magnitude more texture binds per frame than room-bounded games at equivalent settings. Processors with a weak single-thread beware.

Here's the solution I found:

- `Interpreter::ResolveResourceCached(path)`: memoizes path pointer -> resource on `G_SETTIMG`; straight pass-through when disabled.
- The table clears with `TextureCacheClear()`.
- `Interpreter::SetResolvedResourceCacheEnabled(bool)` is the opt-in, following the existing setter convention. Off by default: zero behavior change for existing ports.

No memory tradeoff: the resource manager's own cache already holds these resources, so the memo pins nothing new. If multiple ports adopt this and it's deemed a success then the opt-in could be removed entirely.

### DIFF
--- a/include/fast/interpreter.h
+++ b/include/fast/interpreter.h
@@ -420,6 +420,7 @@ class Interpreter {
     ColorCombiner* LookupOrCreateColorCombiner(const ColorCombinerKey& key);
     void ShaderCacheClear();
     void TextureCacheClear();
+    std::shared_ptr<Ship::IResource> ResolveResourceCached(const char* path);
     bool TextureCacheLookup(int i, const TextureCacheKey& key);
     void TextureCacheDelete(const uint8_t* origAddr);
     void ImportTextureRgba16(int tile, bool importReplacement);
@@ -436,6 +437,10 @@ class Interpreter {
     void ImportTexture(int i, int tile, bool importReplacement);
     void ImportTextureMask(int i, int tile);
     void CalculateNormalDir(const F3DLight_t*, float coeffs[3]);
+    // Opt-in memoization of OTR texture-path resolution, keyed by display-list
+    // pointer and dropped with the texture cache. Safe for ports whose display
+    // lists carry stable path pointers; off by default.
+    void SetResolvedResourceCacheEnabled(bool enabled);
 
     void GfxSpMatrix(uint8_t params, const int32_t* addr);
     void GfxSpPopMatrix(uint32_t count);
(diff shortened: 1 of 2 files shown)


include/fast/interpreter.h
```diff
@@ -436,6 +437,10 @@ class Interpreter {
     void ImportTexture(int i, int tile, bool importReplacement);
     void ImportTextureMask(int i, int tile);
     void CalculateNormalDir(const F3DLight_t*, float coeffs[3]);
+    // Opt-in memoization of OTR texture-path resolution, keyed by display-list
+    // pointer and dropped with the texture cache. Safe for ports whose display
+    // lists carry stable path pointers; off by default.
+    void SetResolvedResourceCacheEnabled(bool enabled);
```
briaguya0, line 443: i wonder if this makes more sense to have as a build-time cmake option  or as a runtime setting like this. i don't have strong feelings one way or the other but figured it's worth discussing.

MegaMech: Is there a reason you wouldn't want to enable this by default? I think most the ports have proper asset ptrs?

JeodC: I made it flexible like this on purpose so it's easily adapted if it's more widely successful.

Malkierian: _Memoization_

Also, we can probably make it automatic in the future, but it's untested on other ports, so it's better to keep the current functionality in place and let ports manually test it as they can.

Malkierian (approved):

include/fast/interpreter.h
JeodC: [Memoization](https://en.wikipedia.org/wiki/Memoization) increases the memory footprint in exchange for runtime speed. Here, none of it is resource data. `mResourceCache` holds `shared_ptr<IResource>` strongly and never evicts outside an explicit unload, so the memo's `shared_ptr` copies pin nothing that wasn't already pinned. The cost is purely hash-table nodes. There's an argument that could be made for it to remain opt-in and have a toggleable cvar, but it's already negligible.

PR #1180, opened 2026-07-30
Fix GfxDebuggerWindow.cpp

MegaMech:
The gui load command now expects a str path. The pointer is no longer allowed. Replaced with empty str.

Couldn't find a way to put the path in at this spot.


MegaMech:
Closing because fix already pushed.

PR #1189, opened 2026-07-31
fix(audio): fix SDL backend buffer sizing causing audio cuts/glitches when CPU is struggling

bassdr:
Port-maintenance counterpart of #1107.

Fixes SDL-backend buffer sizing that caused audio cuts/glitches when the CPU is
struggling:

- Use `SDL_InitSubSystem(SDL_INIT_AUDIO)` instead of `SDL_Init()` so unrelated
  SDL subsystems aren't reinitialised.
- Double the hardware buffer size (`GetSampleLength() * 2`) to reduce backend
  underruns. This only affects SDL/backend latency, not the emulator buffering
  logic.
- Remove the hardcoded 6000-frame fill guard in `DoPlay()`; it was an arbitrary
  threshold that starved audio on slow frames. SDL's internal queue handles
  backpressure.
- Read the actual channel count back from SDL's `have` spec instead of assuming
  the requested value was honoured.
- Minor: `NULL` -> `nullptr`, style cleanup, clearer init log.

Applies cleanly onto port-maintenance.

### DIFF
--- a/include/ship/audio/SDLAudioPlayer.h
+++ b/include/ship/audio/SDLAudioPlayer.h
@@ -50,6 +50,6 @@ class SDLAudioPlayer final : public AudioPlayer {
 
   private:
     SDL_AudioDeviceID mDevice = 0; ///< Handle to the opened SDL audio device.
-    int32_t mNumChannels = 2;      ///< Number of output channels (2 for stereo, 6 for 5.1).
+    uint8_t mNumChannels = 2;      ///< Number of output channels (2 for stereo, 6 for 5.1).
 };
 } // namespace Ship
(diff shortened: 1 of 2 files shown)


bassdr:
Port-maintenance counterpart of #1107 (which targets `main`). Same commit, cherry-picked cleanly.

bassdr:
Superseded by #1251 and closing.

#1251 removes the same 6000-frame drop, and goes further: the ring buffer there is what the drop was crudely standing in for.

One part of this PR actively conflicts with that, which is worth recording. Doubling `want.samples` made sense while SDL owned the buffering - more device buffer, fewer underruns. Once the ring absorbs the producer's jitter, a bigger device buffer only means the device holds more audio ahead of us, which is latency we were trying to remove. #1251 keeps `want.samples` at `GetSampleLength()` deliberately.

PR #1190, opened 2026-07-31
Fix Windows crash handler producing unusable tracebacks

JeodC:
Windows crash reports print blank frames with null addresses, even with a matching PDB sitting next to the executable, and the one symbol that did resolve was typically a misleading nearest-export match rather than a real frame. Several bugs combined: `SymInitialize` was given `"debug"` as a search path (resolved against the cwd, not the executable's directory) with `SYMOPT_IGNORE_IMAGEDIR` suppressing the fallback, so the PDB never loaded; `SymFromAddr`'s return value was ignored, so a zeroed `SYMBOL_INFO` got printed as a frame; `SYMOPT_LOAD_LINES` was never set, so line numbers could never have worked; and `STACKFRAME64` was left zeroed on x64, working only by way of `StackWalk64`'s undocumented fallback to the `CONTEXT` record. This fixes all four, adds module base and RVA to every frame so tracebacks stay resolvable offline under ASLR, reports the operation and faulting address for access violations, and writes a minidump to `logs/<AppName>-crash.dmp` before the stack walk (DbgHelp allocates while walking and would otherwise pollute the captured state). No behavior change on Linux; ARM64 paths remain stubbed as before.

### DIFF
--- a/src/ship/debug/CrashHandler.cpp
+++ b/src/ship/debug/CrashHandler.cpp
@@ -11,6 +11,8 @@
 #include <excpt.h>
 
 #pragma comment(lib, "Dbghelp.lib")
+
+#include <filesystem>
 #endif
 
 namespace Ship {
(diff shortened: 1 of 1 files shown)


JeodC:
cc @Malkierian

Kenix3 (approved):

PR #783, opened 2025-01-05
try sdl3

briaguya0:
overall todo list on the soh PR
* https://github.com/HarbourMasters/Shipwright/pull/4809

sdl migration guide https://wiki.libsdl.org/SDL3/README/migration

PR #788, opened 2025-01-10
imgui 1.90.6 -> 1.91.6

briaguya0:
decided to handle the breaking `ImTextureID` change by patching `imconfig.h` to have
```c
#define ImTextureID void*
```
to switch us back to what we had before (which is working)

note: i early merged https://github.com/Kenix3/libultraship/pull/789 into this branch to get green checks on linux builds

soh pr:
* https://github.com/HarbourMasters/Shipwright/pull/4838

edit: moved to draft, trying another thing
edit: other thing worked, moving back out of draft

<details>
  <summary>stuff from before the `imconfig` patching</summary>
most of these changes are from 9b6a91ea36c0d5af375249a009f9d307959006da where i mentioned

> in https://github.com/ocornut/imgui/commit/92b94980c69ff3d3d7f5dc30c1c19abfb72db47e `ImTextureID` was updated from `void*` to `ImU64`, which is defined as `unsigned long long` (which conflicts with some implementations of `uint64_t` that have it as `unsigned long`)

i linked some info on this and some other breaking changes in https://github.com/Kenix3/libultraship/pull/770

i'm very open to feedback/ideas on how to handle this `ImGui` change - i know we have some other options but this one seemed reasonable
</details>

PR #1191, opened 2026-07-31
Migrate libultraship to SDL3

mckinlee:
## Summary

- move libultraship's window, input, controller, and SDL audio paths to SDL3
- update the ImGui backend and platform dependency setup for SDL3
- preserve gamepad handles across hotplug and correctly round display refresh rates
- retain centered mouse capture through SDL3's window mouse rectangle
- skip desktop GLEW initialization when building the OpenGL ES backend

## Why

This supersedes #783 with the migration replayed on current `main`. It carries forward the later [`sdl3-updated`](https://github.com/briaguya0/libultraship/tree/sdl3-updated) implementation used by Shipwright #4859 while closing the remaining SDL3 migration gaps.

## Testing

- Windows x64 Release build
- focused connected-device and SDL window backend tests (2/2)
- all 618 libultraship tests

### DIFF
--- a/.github/workflows/tidy-format-validation.yml
+++ b/.github/workflows/tidy-format-validation.yml
@@ -9,7 +9,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          fetch-depth: 0
+          fetch-depth: 2
       - name: Install dependencies
         run: |
           sudo apt-get update
@@ -52,7 +52,7 @@ jobs:
           mkdir clang-tidy-result
       - name: Analyze
         run: |
-          git diff -U0 "${{ github.event.pull_request.base.sha }}"...HEAD -- 'src' 'include' ':!include/libultraship/color.h' ':!include/ship/utils/color.h' ':!include/libultraship/libultra/*' ':!src/libultraship/libultra/*' ':!src/fast/*' ':!include/fast/*' ':!src/ship/utils/AppleFolderManager.mm' | clang-tidy-diff -p1 -path build -regex='.*\.(cpp|cc|cxx|mm)$' -export-fixes clang-tidy-result/fixes.yml
+          git diff -U0 HEAD^ -- 'src' 'include' ':!include/libultraship/color.h' ':!include/ship/utils/color.h' ':!include/libultraship/libultra/*' ':!src/libultraship/libultra/*' ':!src/fast/*' ':!include/fast/*' ':!src/ship/utils/AppleFolderManager.mm' | clang-tidy-diff -p1 -path build -export-fixes clang-tidy-result/fixes.yml
       - name: Save PR metadata
         run: |
           echo ${{ github.event.number }} > clang-tidy-result/pr-id.txt
--- a/.github/workflows/tidy-format-validation.yml
+++ b/.github/workflows/tidy-format-validation.yml
@@ -9,7 +9,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          fetch-depth: 2
+          fetch-depth: 0
       - name: Install dependencies
         run: |
           sudo apt-get update
@@ -25,12 +25,11 @@ jobs:
       - name: Install latest SDL
         run: |
           export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
-          wget https://www.libsdl.org/release/SDL2-2.24.1.tar.gz
-          tar -xzf SDL2-2.24.1.tar.gz
-          cd SDL2-2.24.1
-          ./configure
-          make -j 10
-          sudo make install
+          wget https://github.com/libsdl-org/SDL/releases/download/release-3.4.12/SDL3-3.4.12.tar.gz
+          tar -xzf SDL3-3.4.12.tar.gz
+          cmake -S SDL3-3.4.12 -B SDL3-3.4.12/build -GNinja -DCMAKE_BUILD_TYPE=Release
+          cmake --build SDL3-3.4.12/build --parallel 10
+          sudo cmake --install SDL3-3.4.12/build --prefix /usr/local
       - name: Install latest tinyxml2
         run: |
           export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
@@ -53,7 +52,7 @@ jobs:
           mkdir clang-tidy-result
       - name: Analyze
         run: |
-          git diff -U0 HEAD^ -- 'src' 'include' ':!include/libultraship/color.h' ':!include/ship/utils/color.h' ':!include/libultraship/libultra/*' ':!src/libultraship/libultra/*' ':!src/fast/*' ':!include/fast/*' ':!src/ship/utils/AppleFolderManager.mm' | clang-tidy-diff -p1 -path build -export-fixes clang-tidy-result/fixes.yml
+          git diff -U0 "${{ github.event.pull_request.base.sha }}"...HEAD -- 'src' 'include' ':!include/libultraship/color.h' ':!include/ship/utils/color.h' ':!include/libultraship/libultra/*' ':!src/libultraship/libultra/*' ':!src/fast/*' ':!include/fast/*' ':!src/ship/utils/AppleFolderManager.mm' | clang-tidy-diff -p1 -path build -regex='.*\.(cpp|cc|cxx|mm)$' -export-fixes clang-tidy-result/fixes.yml
       - name: Save PR metadata
         run: |
           echo ${{ github.event.number }} > clang-tidy-result/pr-id.txt
--- a/.github/workflows/tidy-format-validation.yml
+++ b/.github/workflows/tidy-format-validation.yml
@@ -52,7 +52,7 @@ jobs:
           mkdir clang-tidy-result
       - name: Analyze
         run: |
-          git diff -U0 HEAD^ -- 'src' 'include' ':!include/libultraship/color.h' ':!include/ship/utils/color.h' ':!include/libultraship/libultra/*' ':!src/libultraship/libultra/*' ':!src/fast/*' ':!include/fast/*' ':!src/ship/utils/AppleFolderManager.mm' | clang-tidy-diff -p1 -path build -export-fixes clang-tidy-result/fixes.yml
+          git diff -U0 HEAD^ -- 'src' 'include' ':!include/libultraship/color.h' ':!include/ship/utils/color.h' ':!include/libultraship/libultra/*' ':!src/libultraship/libultra/*' ':!src/fast/*' ':!include/fast/*' ':!src/ship/utils/AppleFolderManager.mm' ':!include/ship/port/mobile/MobileImpl.h' | clang-tidy-diff -p1 -path build -export-fixes clang-tidy-result/fixes.yml
       - name: Save PR metadata
         run: |
           echo ${{ github.event.number }} > clang-tidy-result/pr-id.txt
--- a/cmake/dependencies/linux.cmake
+++ b/cmake/dependencies/linux.cmake
@@ -1,16 +1,5 @@
 #=================== ImGui ===================
-find_package(SDL3 QUIET CONFIG COMPONENTS SDL3)
-if(NOT SDL3_FOUND)
-    include(FetchContent)
-    FetchContent_Declare(
-        SDL3
-        GIT_REPOSITORY https://github.com/libsdl-org/SDL.git
-        GIT_TAG release-3.4.12
-        OVERRIDE_FIND_PACKAGE
-    )
-    message("SDL3 not found. Downloading now...")
-    FetchContent_MakeAvailable(SDL3)
-endif()
+find_package(SDL3 REQUIRED CONFIG COMPONENTS SDL3)
 target_link_libraries(ImGui PUBLIC SDL3::SDL3)
 
 if (USE_OPENGLES)
--- a/cmake/dependencies/linux.cmake
+++ b/cmake/dependencies/linux.cmake
@@ -1,6 +1,17 @@
 #=================== ImGui ===================
-find_package(SDL2 REQUIRED)
-target_link_libraries(ImGui PUBLIC SDL2::SDL2)
+find_package(SDL3 QUIET CONFIG COMPONENTS SDL3)
+if(NOT SDL3_FOUND)
+    include(FetchContent)
+    FetchContent_Declare(
+        SDL3
+        GIT_REPOSITORY https://github.com/libsdl-org/SDL.git
+        GIT_TAG release-3.4.12
+        OVERRIDE_FIND_PACKAGE
+    )
+    message("SDL3 not found. Downloading now...")
+    FetchContent_MakeAvailable(SDL3)
+endif()
+target_link_libraries(ImGui PUBLIC SDL3::SDL3)
 
 if (USE_OPENGLES)
     target_link_libraries(ImGui PUBLIC ${OPENGL_GLESv2_LIBRARY})
--- a/include/fast/Fast3dGui.h
+++ b/include/fast/Fast3dGui.h
@@ -1,5 +1,5 @@
 #pragma once
-#include <SDL2/SDL.h>
+#include <SDL3/SDL.h>
 #include "ship/window/gui/Gui.h"
 #include "fast/WindowEvent.h"
 #include "fast/resource/type/Texture.h"
--- a/include/fast/Fast3dGui.h
+++ b/include/fast/Fast3dGui.h
@@ -77,10 +77,8 @@ class Fast3dGui : public Ship::Gui {
     /**
      * @brief Forwards a platform window event to the active ImGui backend.
      *
-     * Only Fast3D backends (gfx_sdl,
-     * gfx_dxgi) construct and dispatch WindowEvents.
-     * Callers retrieve the Gui via Window::GetGui() and
-     * dynamic_cast to Fast3dGui.
+     * Only Fast3D backends (gfx_sdl, gfx_dxgi) construct and dispatch WindowEvents.
+     * Callers retrieve the Gui via Window::GetGui() and dynamic_cast to Fast3dGui.
      * @param event Platform event (SDL or Win32) wrapped in a Fast::WindowEvent.
      */
     void HandleWindowEvents(Fast::WindowEvent event);
--- a/include/fast/Fast3dGui.h
+++ b/include/fast/Fast3dGui.h
@@ -77,9 +77,10 @@ class Fast3dGui : public Ship::Gui {
     /**
      * @brief Forwards a platform window event to the active ImGui backend.
      *
-     * Only Fast3D backends (gfx_sdl, gfx_dxgi) construct and dispatch WindowEvents.
-     * Callers retrieve the Gui via
-     * Window::GetGui() and dynamic_cast to Fast3dGui.
+     * Only Fast3D backends (gfx_sdl,
+     * gfx_dxgi) construct and dispatch WindowEvents.
+     * Callers retrieve the Gui via Window::GetGui() and
+     * dynamic_cast to Fast3dGui.
      * @param event Platform event (SDL or Win32) wrapped in a Fast::WindowEvent.
      */
     void HandleWindowEvents(Fast::WindowEvent event);
--- a/include/fast/Fast3dGui.h
+++ b/include/fast/Fast3dGui.h
@@ -12,7 +12,7 @@ class ResourceManager;
 } // namespace Ship
 
 // Fixes issue #926: HandleWindowEvents is only ever called from Fast3D backend code
-// (gfx_sdl2.cpp, gfx_dxgi.cpp) and must not be a virtual method on Ship::Gui.
+// (gfx_sdl.cpp, gfx_dxgi.cpp) and must not be a virtual method on Ship::Gui.
 // The WindowEvent type has been moved to the Fast namespace so that ship code does
 // not depend on any Fast3D or platform-specific types.
 
@@ -77,8 +77,9 @@ class Fast3dGui : public Ship::Gui {
     /**
      * @brief Forwards a platform window event to the active ImGui backend.
      *
-     * Only Fast3D backends (gfx_sdl2, gfx_dxgi) construct and dispatch WindowEvents.
-     * Callers retrieve the Gui via Window::GetGui() and dynamic_cast to Fast3dGui.
+     * Only Fast3D backends (gfx_sdl, gfx_dxgi) construct and dispatch WindowEvents.
+     * Callers retrieve the Gui via
+     * Window::GetGui() and dynamic_cast to Fast3dGui.
      * @param event Platform event (SDL or Win32) wrapped in a Fast::WindowEvent.
      */
     void HandleWindowEvents(Fast::WindowEvent event);
--- a/include/fast/WindowEvent.h
+++ b/include/fast/WindowEvent.h
@@ -1,7 +1,7 @@
 #pragma once
 
 // Fixes issue #926: ship/window/gui/Gui.h previously defined WindowEvent in the Ship namespace,
-// but this type is only ever constructed and consumed by Fast3D backend code (gfx_sdl2.cpp,
+// but this type is only ever constructed and consumed by Fast3D backend code (gfx_sdl.cpp,
 // gfx_dxgi.cpp) and processed by Fast3dGui::HandleWindowEvents. Moving it to the Fast namespace
 // enforces the correct include hierarchy: fast can include fast, ship cannot include fast.
 
@@ -10,9 +10,9 @@ namespace Fast {
 /**
  * @brief Platform-agnostic wrapper around a window system event.
  *
- * Constructed by the Fast3D backends (gfx_sdl2, gfx_dxgi) and passed to
- * Fast3dGui::HandleWindowEvents() so that the ImGui backend can process input
- * events without the ship layer depending on any Fast3D or platform-specific type.
+ * Constructed by the Fast3D backends (gfx_sdl, gfx_dxgi) and passed to
+ * Fast3dGui::HandleWindowEvents() so that the
+ * ImGui backend can process input events without the ship layer depending on any Fast3D or platform-specific type.
  */
 typedef union {
     struct {
--- a/include/fast/WindowEvent.h
+++ b/include/fast/WindowEvent.h
@@ -10,9 +10,11 @@ namespace Fast {
 /**
  * @brief Platform-agnostic wrapper around a window system event.
  *
- * Constructed by the Fast3D backends (gfx_sdl, gfx_dxgi) and passed to
- * Fast3dGui::HandleWindowEvents() so that the
- * ImGui backend can process input events without the ship layer depending on any Fast3D or platform-specific type.
+ * Constructed by the Fast3D backends (gfx_sdl,
+ * gfx_dxgi) and passed to
+ * Fast3dGui::HandleWindowEvents() so that the ImGui backend can process input
+ * events
+ * without the ship layer depending on any Fast3D or platform-specific type.
  */
 typedef union {
     struct {
--- a/include/fast/WindowEvent.h
+++ b/include/fast/WindowEvent.h
@@ -10,11 +10,9 @@ namespace Fast {
 /**
  * @brief Platform-agnostic wrapper around a window system event.
  *
- * Constructed by the Fast3D backends (gfx_sdl,
- * gfx_dxgi) and passed to
+ * Constructed by the Fast3D backends (gfx_sdl, gfx_dxgi) and passed to
  * Fast3dGui::HandleWindowEvents() so that the ImGui backend can process input
- * events
- * without the ship layer depending on any Fast3D or platform-specific type.
+ * events without the ship layer depending on any Fast3D or platform-specific type.
  */
 typedef union {
     struct {
--- a/include/ship/controller/controldevice/controller/mapping/factories/RumbleMappingFactory.h
+++ b/include/ship/controller/controldevice/controller/mapping/factories/RumbleMappingFactory.h
@@ -45,14 +45,10 @@ class RumbleMappingFactory {
                                   std::shared_ptr<ControlDeck> controlDeck);
 
     /**
-     * @brief Creates default SDL rumble mappings for a specific physical device type.
-     * @param physicalDeviceType
-     * Type of physical device receiving the mapping.
+     * @param physicalDeviceType Type of physical device receiving the mapping.
      * @param portIndex Controller port index (0-based).
-     *
      * @param consoleVariable    ConsoleVariable for persisting mappings.
-     * @param controlDeck        ControlDeck
-     * for physical device access.
+     * @param controlDeck        ControlDeck for physical device access.
      * @return A default SDL rumble mapping for SDL gamepads, or an empty vector for other device types.
      */
     static std::vector<std::shared_ptr<ControllerRumbleMapping>>
--- a/include/ship/controller/controldevice/controller/mapping/factories/RumbleMappingFactory.h
+++ b/include/ship/controller/controldevice/controller/mapping/factories/RumbleMappingFactory.h
@@ -46,11 +46,12 @@ class RumbleMappingFactory {
 
     /**
      * @brief Creates default SDL rumble mappings for a specific physical device type.
-     * @param physicalDeviceType SDL gamepad type.
-     * @param portIndex          Controller port index (0-based).
+     * @param physicalDeviceType Type of physical device receiving the mapping.
+     * @param portIndex Controller port
+     * index (0-based).
      * @param consoleVariable    ConsoleVariable for persisting mappings.
      * @param controlDeck        ControlDeck for physical device access.
-     * @return Vector of rumble mappings suitable for the device type.
+     * @return A default SDL rumble mapping for SDL gamepads, or an empty vector for other device types.
      */
     static std::vector<std::shared_ptr<ControllerRumbleMapping>>
     CreateDefaultSDLRumbleMappings(PhysicalDeviceType physicalDeviceType, uint8_t portIndex,
--- a/include/ship/controller/controldevice/controller/mapping/factories/RumbleMappingFactory.h
+++ b/include/ship/controller/controldevice/controller/mapping/factories/RumbleMappingFactory.h
@@ -46,7 +46,7 @@ class RumbleMappingFactory {
 
     /**
      * @brief Creates default SDL rumble mappings for a specific physical device type.
-     * @param portIndex Controller port index (0-based).
+     * @param physicalDeviceType Type of physical device receiving the mapping.
      * @param portIndex Controller port index (0-based).
      * @param consoleVariable    ConsoleVariable for persisting mappings.
      * @param controlDeck        ControlDeck for physical device access.
--- a/include/ship/controller/controldevice/controller/mapping/factories/RumbleMappingFactory.h
+++ b/include/ship/controller/controldevice/controller/mapping/factories/RumbleMappingFactory.h
@@ -46,11 +46,13 @@ class RumbleMappingFactory {
 
     /**
      * @brief Creates default SDL rumble mappings for a specific physical device type.
-     * @param physicalDeviceType Type of physical device receiving the mapping.
-     * @param portIndex Controller port
-     * index (0-based).
+     * @param physicalDeviceType
+     * Type of physical device receiving the mapping.
+     * @param portIndex Controller port index (0-based).
+     *
      * @param consoleVariable    ConsoleVariable for persisting mappings.
-     * @param controlDeck        ControlDeck for physical device access.
+     * @param controlDeck        ControlDeck
+     * for physical device access.
      * @return A default SDL rumble mapping for SDL gamepads, or an empty vector for other device types.
      */
     static std::vector<std::shared_ptr<ControllerRumbleMapping>>
--- a/include/ship/controller/controldevice/controller/mapping/factories/RumbleMappingFactory.h
+++ b/include/ship/controller/controldevice/controller/mapping/factories/RumbleMappingFactory.h
@@ -45,7 +45,8 @@ class RumbleMappingFactory {
                                   std::shared_ptr<ControlDeck> controlDeck);
 
     /**
-     * @param physicalDeviceType Type of physical device receiving the mapping.
+     * @brief Creates default SDL rumble mappings for a specific physical device type.
+     * @param portIndex Controller port index (0-based).
      * @param portIndex Controller port index (0-based).
      * @param consoleVariable    ConsoleVariable for persisting mappings.
      * @param controlDeck        ControlDeck for physical device access.
--- a/include/ship/controller/controldevice/controller/mapping/factories/RumbleMappingFactory.h
+++ b/include/ship/controller/controldevice/controller/mapping/factories/RumbleMappingFactory.h
@@ -46,8 +46,8 @@ class RumbleMappingFactory {
 
     /**
      * @brief Creates default SDL rumble mappings for a specific physical device type.
-     * @param physicalDeviceType The type of physical device (e.g., SDL2 gamepad, Nintendo controller).
-     * @param portIndex          The controller port index (0-based).
+     * @param physicalDeviceType SDL gamepad type.
+     * @param portIndex          Controller port index (0-based).
      * @param consoleVariable    ConsoleVariable for persisting mappings.
      * @param controlDeck        ControlDeck for physical device access.
      * @return Vector of rumble mappings suitable for the device type.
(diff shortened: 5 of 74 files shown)


briaguya0:
> [!NOTE]
> This review was generated by Claude (Claude Code), driven by @briaguya0. The branch archaeology and diff comparisons below were mechanically verified against the repos; the judgment calls are worth a human second opinion.

## Review: SDL3 migration coverage

outdated

<details>
<summary>original comment</summary>

I checked this against the LUS-side changes that HarbourMasters/Shipwright#4859 actually consumes, to confirm nothing was lost in the replay onto current `main`.

### Note on the "replaces the earlier SDL3 work in #783" framing

Worth correcting in the description, since it sent me down the wrong path initially. There are four SDL3 branches on `briaguya0/libultraship`:

| branch | date | based on `main` @ | what it is |
|---|---|---|---|
| `play-with-sdl3` | Feb 2024 | `0833afad` | 1-commit first poke |
| `try-sdl3` | Jan 3–4 2025 | `78d951e4` | **← what #783 points at** |
| `sdl3` | Jan 11 2025 | `9a974e00` | fresh redo |
| `sdl3-updated` | Sep 27 2025 | `7f737f8` merged in | **← what Shipwright#4859 pins** |

`merge-base(try-sdl3, sdl3-updated)` is `78d951e4` — their common *main* ancestor. They share zero commits; `sdl3` was not a rebase of `try-sdl3` but a from-scratch redo a week later, re-running the same `rename_symbols.py` / `rename_headers.py` / `rename_macros.py` pass. `sdl3-updated` then descends from `sdl3`.

So #783's branch is two generations behind what SoH consumes, and diffing this PR against #783's changed-files view is misleading. The right baseline is `7f737f8...2465c44`. **Suggest rewording to "supersedes #783" and linking `sdl3-updated` instead.**

For completeness, nothing substantive is stranded in #783. Of its 17 unique files: 12 are pure `__APPLE__` → `SDL_PLATFORM_APPLE` cosmetic sweeps; `ConsoleWindow.cpp` is the `ImGui::GetKeyIndex()` removal from the imgui 1.91.6 bump that landed separately as #788; the `gfx_metal*` / `gfx_opengl` / `gfx_sdl2` entries are the same files pre-`backends/` move; and `SDLMapping.cpp` + `deviceindex/*` covered the `ShipDeviceIndex` subsystem that has since been deleted from `main` in favor of `ConnectedPhysicalDeviceManager`.

### Coverage vs `sdl3-updated`: complete

Every file that branch touched has a counterpart here once the `src/` → `src/ship` / `src/fast` / `include/` reorg is accounted for. Grepping this branch for residual SDL2 API (`SDL2/`, `SDL_GameController*`, `SDL_CONTROLLER_*`, `SDL_SysWM*`, `SDL_GL_GetDrawableSize`, `SDL_WINDOWEVENT`, `imgui_impl_sdl2`, `SDL_bool`, `SDL_QueueAudio`, `SDL_EventState`, …) returns zero hits outside one URL in a log string.

### Where this PR finishes what that branch left stubbed

| | `sdl3-updated` | this PR |
|---|---|---|
| LED capability | `if (false)` in the factory and `if (true) continue;` in `SDLLEDMapping::SetLEDColor` — **LED silently disabled** | `SDL_PROP_GAMEPAD_CAP_RGB_LED_BOOLEAN` + `SDL_SetGamepadLED` |
| Rumble capability | `if (false)` with a todo | `SDL_PROP_GAMEPAD_CAP_RUMBLE_BOOLEAN` |
| Device add/remove | `event.cdevice.which` — no such union member in SDL3 | `event.gdevice.which` |
| Exclusive vs windowed fullscreen | todo; bare `SDL_SetWindowFullscreen(mWnd, on)`, dropping `CVAR_SDL_WINDOWED_FULLSCREEN` | `SDL_GetClosestFullscreenDisplayMode` + `SDL_SetWindowFullscreenMode(nullptr)` for borderless |
| Refresh rate | `*refresh_rate = mode->refresh_rate;` — no null check, no 60 fallback | null-checked, `> 0.0f`, falls back to 60 |
| Mobile text input | untouched; SDL3 requires a window arg | threaded through `Mobile::ImGuiProcessEvent(SDL_Window*, bool)` |
| `SDL_GetJoysticks` array | never freed | `SDL_free`'d, plus new `CloseConnectedSDLGamepads()` closing the gamepads |
| GLES/Emscripten GLEW | kept `#if !defined(__linux__) && !defined(__OpenBSD__)` around `glewInit()` | `FAST_GFX_OPENGL_USE_GLEW`, tied to the header branches that actually select GLEW |
| Audio teardown | `DoClose` not migrated | `SDL_ClearAudioStream` / `SDL_DestroyAudioStream`, `Buffered()` null-guarded |
| `common.cmake` | left stale `${SDL2_INCLUDE_DIRS}` | removed |
| SDL version | `preview-3.1.8` / `3.2.22` | `release-3.4.12` |

Also only here: `openbsd.cmake`, `windows-vcpkg.cmake`, the imgui patch retargeted to `imgui_impl_sdl3.cpp`, `GetPrimaryMonitorRect`, the `SDL_WarpMouseInWindow` float cast, null-guarded `Destroy()`, and iOS static SDL.

**Keeping `__APPLE__` instead of `SDL_PLATFORM_APPLE` was the right call** and is not a gap. That rename was buggy on the old branches: `src/config/Config.cpp`, `src/window/gui/StatsWindow.cpp`, and `src/graphic/Fast3D/gfx_pc.cpp` used `#ifdef SDL_PLATFORM_APPLE` while including no SDL header, so those macOS branches silently compiled out. `__APPLE__` is a compiler builtin and needs no include.

### Two things to confirm before merge

1. **`SetMouseCapture` drops the cursor-clip workaround.** `main` has a manual `SDL_SetWindowMouseRect` clamp with the comment *"`SDL_HINT_MOUSE_RELATIVE_MODE_CENTER` isn't working as expected. Revisit on SDL3."* This PR removes it for plain `SDL_SetWindowRelativeMouseMode`. That reads like the intended follow-through, but it's a behavior change that deserves an actual mouse-capture test rather than an assumption — and `mCursorClip` in `include/fast/backends/gfx_sdl.h` is now an unused member that should probably go with it.

2. **One line from the SoH branch isn't carried over**: `SDLAxisDirectionToButtonMapping::EraseFromConfig` there also cleared `"%s.ShipDeviceIndex"`. That was an unrelated drive-by, and it looks moot now since `SaveToConfig` no longer writes that key — just confirming it's intentional and no legacy config key is left orphaned.

</details>

briaguya0:
> [!NOTE]
> Follow-up generated by Claude (Claude Code), driven by @briaguya0.

## Correction on the refresh-rate row: small regression vs `main`, not just an improvement over `sdl3-updated`

appears to have been addressed in https://github.com/Kenix3/libultraship/pull/1191/changes/f1cebdb739d9d64d407027ca4794a406971d80b0

<details>
<summary>Original Comment</summary>

I flagged `GetActiveWindowRefreshRate` above as strictly better than the SoH branch. That's true, but I understated it — digging into where the 60Hz fallback came from turned up a separate one-off that this PR newly introduces relative to `main`.

### Why the fallback exists

`git log -S` points at #828 (2025-03-22): *"Assume 60Hz if SDL reports 0Hz — this fixes a crash on MacOS, where SDL sometimes outputs 0Hz for secondary displays on startup."* Worth knowing it's a real crash fix, since `Fast3dWindow::GetCurrentRefreshRate` has no in-tree callers and is consumed downstream for frame pacing. This PR preserves it correctly, and `> 0.0f` is the right float form of the guard. The `sdl3-updated` branch dropped it outright, so that part stands.

### The new issue: truncation instead of rounding

On `main` (`src/fast/backends/gfx_sdl2.cpp:279`) LUS never does a float→int conversion — SDL2's `refresh_rate` is already an `int` that SDL rounds inside its own backends:

- X11 — `SDL_x11modes.c:198`: `SDL_round((double)info->dotClock / (hTotal * vTotal))`
- Cocoa — `SDL_cocoamodes.m:89`: `(int)(CGDisplayModeGetRefreshRate(vidmode) + 0.5)`
- Windows — `SDL_windowsmodes.c:184`: raw `dmDisplayFrequency`

SDL3 changed that contract. It keeps the rate as an exact rational and derives a float quantized to two decimals (`SDL_video.c:811`):

```c
mode->refresh_rate = ((100 * (Sint64)mode->refresh_rate_numerator) / mode->refresh_rate_denominator) / 100.0f;
```

For 60000/1001 that is `(100*60000)/1001 = 5994`, `/100.0f` → exactly `59.94f`. SDL3 hands over the real value and leaves the integer policy to the caller; this PR picks truncation via `static_cast<uint32_t>`.

| display @ 59.94Hz | SDL2 (`main`) | SDL3 + `static_cast` (this PR) |
|---|---|---|
| Linux / X11 | **60** | **59** ← changed |
| macOS | **60** | **59** ← changed |
| Windows | 59 | 59 — unchanged |

So it's a new off-by-one on Linux and macOS specifically (Windows' `dmDisplayFrequency` was already reporting 59). Same applies to any other non-integer mode — 143.98, 164.83, etc. It surfaces as subtle frame-pacing drift rather than an obvious break, which is why it's easy to miss.

### Suggested fix

One word, and it matches what `gfx_dxgi.cpp:723` already does (`(uint32_t)roundf(mDetectedHz)`):

```cpp
*refresh_rate = mode != nullptr && mode->refresh_rate > 0.0f ? static_cast<uint32_t>(roundf(mode->refresh_rate)) : 60;
```
</details>

briaguya0:
> [!NOTE]
> Follow-up generated by Claude (Claude Code), driven by @briaguya0.

## `RefreshConnectedSDLGamepads`: the leak fix is correct, but the close-all/reopen-all is load-bearing on ImGui

appears to have been addressed

<details>
<summary>original comment</summary>

Expanding on the `CloseConnectedSDLGamepads()` change, because there are two separate properties here and only one is fully self-contained.

### The leak fix is real, and belongs on `main` regardless of SDL3

`main` does `mConnectedSDLGamepads.clear()` and then re-opens every present pad, and `~ConnectedPhysicalDeviceManager()` is empty — `SDL_GameControllerClose` is never called anywhere in the class. Both SDL versions refcount opens rather than handing back a fresh handle:

- SDL2 `SDL_gamecontroller.c:2178-2180` — already open → `++gamecontroller->ref_count`, return existing
- SDL3 `SDL_gamepad.c:2669-2671` — identical

So on `main` every refresh is `+1` per still-connected pad with no matching `-1`. The count climbs monotonically and never returns to zero, so the pads are never actually closed and their device handles stay claimed for the process lifetime — including after a physical unplug, since the removal path also just refreshes.

This PR's close/open pair is properly balanced: net 0 for a pad that stayed, `-1` for one that vanished, `+1` for a new one. The leak is genuinely fixed, this is version-independent, and it could be cherry-picked to `main` on its own.

### The part worth a second look

Blanket close-all-then-reopen-all means every hotplug event closes and immediately reopens every already-connected pad. `SDL_CloseGamepad` is refcounted (`SDL_gamepad.c:3623`), but if the count reaches zero it calls `SDL_CloseJoystick`, which at *its* zero (`SDL_joystick.c:1859-1877`) does a full teardown:

```c
SDL_DestroyProperties(joystick->props);          // caps/props object destroyed
if (joystick->rumble_expiration) SDL_RumbleJoystick(joystick, 0, 0, 0);
CleanupSensorFusion(joystick);                   // gyro/accel state
joystick->driver->Close(joystick);               // real HID/XInput handle closed
joystick->hwdata = NULL;
```

Rumble stopped, sensor fusion dropped, HID handle released — then a full reopen. On Bluetooth pads that's the shape of thing that produces a visible input blip.

**In the current flow it does not reach zero**, because ImGui independently holds a reference. `Fast3dGui::RefreshImGuiGamepads` sets `ImGui_ImplSDL3_SetGamepadMode(AutoAll)`, and `imgui_impl_sdl3.cpp:798-809` opens every gamepad and retains the handles in `bd->Gamepads`. Sequence on hotplug:

1. `SDLAddRemoveDeviceEventHandler::UpdateElement` peeps `GAMEPAD_ADDED` → `RefreshConnectedSDLGamepads()` → close-all (**2 → 1**, no teardown) → reopen-all (**back to 2**)
2. `changed = true` → `RefreshImGuiGamepads()` → sets `WantUpdateGamepadsList = true`
3. next `NewFrame` → ImGui closes its own refs (2 → 1) and reopens (→ 2)

So no disconnect in practice. But that safety is incidental rather than designed — it holds only while ImGui's backend keeps those pads open, and stops holding under `ImGui_ImplSDL3_GamepadMode_Manual`, for a consumer running without `Fast3dGui`, or in any window between gamepad-subsystem init and the first `NewFrame`. In those cases LUS holds the only reference and each hotplug becomes a real teardown/reopen of every connected pad.

Scope is bounded: `RefreshConnectedSDLGamepads` has exactly one entry point (`SDLAddRemoveDeviceEventHandler::UpdateElement`), so this is hotplug-only, not per-frame.

### Suggestion

Worth noting first: in SDL3 `SDL_GetJoysticks` returns **instance IDs** directly, where SDL2's `SDL_NumJoysticks`/index API returned *device indices*. That makes the `openedInstanceId` re-derivation via `SDL_GetJoystickID(SDL_GetGamepadJoystick(gamepad))` redundant — and it's what allows keying the map before opening, so an already-open pad never needs to be reopened at all.

Sketch (untested, guards and bookkeeping elided where unchanged):

```cpp
void ConnectedPhysicalDeviceManager::RefreshConnectedSDLGamepads() {
    static SDL_GUID sZeroGuid;

    int32_t joystickCount = 0;
    SDL_JoystickID* joysticks = SDL_GetJoysticks(&joystickCount);
    if (joysticks == nullptr) {
        return; // enumeration failed; leave the existing set alone rather than dropping every pad
    }

    std::unordered_set<int32_t> seen;

    for (int32_t i = 0; i < joystickCount; i++) {
        const SDL_JoystickID instanceId = joysticks[i];

        // ... zero-GUID and SDL_IsGamepad guards unchanged ...

        SDL_Gamepad* gamepad = nullptr;
        auto existing = mConnectedSDLGamepads.find(instanceId);
        if (existing != mConnectedSDLGamepads.end()) {
            gamepad = existing->second; // already open: keep the handle, don't reopen
        } else {
            gamepad = SDL_OpenGamepad(instanceId);
            if (gamepad == nullptr) {
                SPDLOG_ERROR("SDL_OpenGamepad error (GUID: {}): {}", deviceGuidCStr, SDL_GetError());
                continue;
            }
        }

        seen.insert(instanceId);

        // ... name lookup and mIgnoredInstanceIds bookkeeping unchanged, keyed on instanceId ...
    }

    SDL_free(joysticks);

    // release only what actually went away
    for (auto it = mConnectedSDLGamepads.begin(); it != mConnectedSDLGamepads.end();) {
        if (seen.count(it->first) == 0) {
            SDL_CloseGamepad(it->second);
            it = mConnectedSDLGamepads.erase(it);
        } else {
            ++it;
        }
    }
}
```

Never touches a pad that stayed plugged in, still fixes the leak, and makes the teardown impossible by construction rather than by luck.

Two notes on the sketch: the `joysticks == nullptr` early return now preserves the existing set instead of having already dropped it, which seems better than treating a transient enumeration failure as "everything unplugged" — but it is a deliberate behavior change worth a look. And the destructor's `CloseConnectedSDLGamepads()` is correct as-is and should stay, including the `SDL_WasInit(SDL_INIT_GAMEPAD)` guard, since the destructor can run after `SDL_Quit()` has already invalidated those handles.

</details>

.github/workflows/tidy-format-validation.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          fetch-depth: 2
+          fetch-depth: 0
```
briaguya0: i'm open to this (switching to use `github.event.pull_request` instead of doing a deeper fetch) but it feels unrelated to the sdl3 work, so it should be a separate PR

cmake/dependencies/linux.cmake
```diff
@@ -1,6 +1,17 @@
...
+    include(FetchContent)
+    FetchContent_Declare(
+        SDL3
+        GIT_REPOSITORY https://github.com/libsdl-org/SDL.git
+        GIT_TAG release-3.4.12
+        OVERRIDE_FIND_PACKAGE
+    )
+    message("SDL3 not found. Downloading now...")
+    FetchContent_MakeAvailable(SDL3)
+endif()
```
briaguya0: is there a specific reason to switch this from being `REQUIRED`? i'd prefer to stick to required and continue to do the "install newer" thing in CI for distros that don't have it. i also assume most devs are on distros that ship SDL3.

src/fast/backends/gfx_sdl2.cpp
```diff
@@ -497,21 +522,11 @@ bool GfxWindowBackendSDL2::GetMouseState(uint32_t btn) {
...
-    // Revisit on SDL3
-    auto mouse = SDL_GetWindowMouseRect(mWnd);
-    if (capture) {
-        int w, h;
-        SDL_GetWindowSize(mWnd, &w, &h);
-        mCursorClip = { (w / 2) - 1, (h / 2) - 1, 2, 2 };
-    }
-    SDL_SetWindowMouseRect(mWnd, capture ? &mCursorClip : NULL);
+    SDL_SetWindowRelativeMouseMode(mWnd, capture);
 }
```
briaguya0, line 511: this said "revisit on SDL3," did you test it? it's not clear if we still need this or not and i don't want it to disappear without discussion.

src/fast/backends/gfx_sdl.cpp
briaguya0, line 1: might be worth renaming this to `gfx_sdl.cpp` (or `gfx_sdl3.cpp`)

src/fast/backends/gfx_sdl2.cpp
briaguya0: we should probably change this to not be `GfxWindowBackendSDL2` (and instead have either just `GfxWindowBackendSDL` or use `GfxWindowBackendSDL3`)

include/ship/controller/controldevice/controller/mapping/factories/RumbleMappingFactory.h
```diff
@@ -46,8 +46,8 @@ class RumbleMappingFactory {
 
     /**
      * @brief Creates default SDL rumble mappings for a specific physical device type.
-     * @param physicalDeviceType The type of physical device (e.g., SDL2 gamepad, Nintendo controller).
```
briaguya0, line 49: this was already wrong, but this change makes it *more* wrong. `physticalDeviceType`s are

https://github.com/Kenix3/libultraship/blob/a57f6b5ec8dd1c4a02bf2b8481e20f2441135129/include/ship/controller/physicaldevice/PhysicalDeviceType.h#L11-L16

i think at most that would be used as a guard to early return and not try to make mappings if the type wasn't `SDLGamepad`, but i don't think that guard is actually implemented in there.

include/ship/controller/physicaldevice/SDLAddRemoveDeviceEventHandler.h
```diff
@@ -9,12 +9,12 @@ class ControlDeck;
 class Window;
 
 /**
- * @brief Hidden GuiWindow that listens for SDL controller add/remove events.
```
briaguya0, line 12: curious about the remap event thing here. was that required for the move to SDL3? if not i feel like it should either be a separate PR to main (i think those events exist in SDL2) and then this would be porting that functionality, or it can just be a follow-up.

src/fast/backends/gfx_sdl2.cpp
```diff
@@ -224,19 +224,34 @@ void GfxWindowBackendSDL2::SetFullscreenImpl(bool on, bool call_callback) {
...
+        const SDL_DisplayMode* mode = nullptr;
+
+        if (!windowedFullscreen) {
+            const SDL_DisplayMode* desktopMode = SDL_GetDesktopDisplayMode(display_in_use);
+            if (desktopMode == nullptr ||
+                !SDL_GetClosestFullscreenDisplayMode(display_in_use, desktopMode->w, desktopMode->h,
+                                                     desktopMode->refresh_rate, true, &fullscreenMode)) {
+                SPDLOG_ERROR("Failed to find an exclusive fullscreen mode.");
+                SPDLOG_ERROR(SDL_GetError());
+                return;
```
briaguya0: i think these early returns are changing behavior a bit.

> convo with claude about it
> 
> ok, so it's not fully clear what's going on with the behavior change and the early returns. before we'd have either gotten mode from GetDesktopDisplayMode and used it in SetWindowDisplayMode, or failed to get it in GetDesktopDisplayMode but still tried to call SetWindowFullscreen (which doesn't use mode)          
>
> ● You've got it exactly right. Let me nail down the last piece — what each version does when the mode is absent.
> ● Exactly right, and the doc confirms both halves.
>
>  SDL2 — the mode was advisory. SDL_SetWindowDisplayMode's param doc (SDL_video.h:686): passing NULL / never calling it means "use the window's dimensions and the desktop's format and refresh rate." So on main, if GetDesktopDisplayMode failed you just skipped the refinement, and SDL_SetWindowFullscreen(mWnd, 
>  FULLSCREEN | FULLSCREEN_DESKTOP) still ran and still worked. As you say — that call never took the mode. Failure was nearly a no-op.
>
>  SDL3 — the mode is the mechanism, but "no mode" is still a valid state. SDL_GetWindowFullscreenMode (:984-985) returns "a pointer to the exclusive fullscreen mode to use or NULL for borderless fullscreen desktop mode." So a window that has never had a fullscreen mode set is already in the borderless state.
>
>  That's the piece that makes the early returns look wrong rather than merely different: SDL3 doesn't require you to bail. Not calling SetWindowFullscreenMode, or calling it with nullptr, both land you in borderless — the exact graceful degradation main had. The PR chose to return instead.
>
>  Both returns are avoidable for the same reason:
>
>  - First return (can't find an exclusive mode): leave mode = nullptr, fall through → borderless.
>  - Second return (SetWindowFullscreenMode failed): the window keeps its previous mode, almost certainly the NULL default → falling through still gives borderless.
>
>  What returning actually costs you: SDL_SetWindowFullscreen never runs, mFullScreen never updates, and mOnFullscreenChanged never fires. So the user hits the fullscreen toggle and nothing happens — just a log line. One mitigating detail: since mFullScreen stays unchanged, the if (mFullScreen == on) return; guard at the top doesn't latch, so a second press retries rather than getting permanently stuck.
>
>  Still, "fullscreen silently does nothing on machines where exclusive-mode enumeration fails" is a worse outcome than main's "you get borderless," and SDL3 gives you the fallback for free.

src/ship/audio/SDLAudioPlayer.cpp
```diff
@@ -10,53 +10,44 @@ SDLAudioPlayer::~SDLAudioPlayer() {
...
     SPDLOG_INFO("SDL Audio initialized: {} channels, {} Hz", mNumChannels, this->GetSampleRate());
 
-    SDL_PauseAudioDevice(mDevice, 0);
+    SDL_ResumeAudioDevice(SDL_GetAudioStreamDevice(mStream));
     return true;
 }
 
 int SDLAudioPlayer::Buffered() {
-    return SDL_GetQueuedAudioSize(mDevice) / (sizeof(int16_t) * mNumChannels);
+    return mStream == nullptr ? 0 : SDL_GetAudioStreamQueued(mStream) / (sizeof(int16_t) * mNumChannels);
```
briaguya0, line 46: i had a big comment about deciding between using available and queued here

https://github.com/Kenix3/libultraship/blob/2465c44d810d43422b16fb1c233a14314958a7a9/src/audio/SDLAudioPlayer.cpp#L32-L46

happy to stick with queued but want to make sure it's discussed/considered

src/ship/controller/controldevice/controller/Controller.cpp
```diff
@@ -5,9 +5,9 @@
 #include "ship/config/ConsoleVariable.h"
 #include "ship/window/Window.h"
 #if __APPLE__
-#include <SDL_events.h>
+#include <SDL3/SDL_events.h>
 #else
-#include <SDL2/SDL_events.h>
+#include <SDL3/SDL_events.h>
 #endif
```
briaguya0: this whole block can just be
```cpp
#include <SDL3/SDL_events.h>
```
now

src/ship/controller/physicaldevice/SDLAddRemoveDeviceEventHandler.cpp
```diff
@@ -26,26 +26,29 @@ void SDLAddRemoveDeviceEventHandler::UpdateElement() {
...
     }
 
-    // The connected controller set changed, so re-point the ImGui gamepad
-    // backend at it (keeps menu navigation working across hotplug).
+    while (SDL_PeepEvents(&event, 1, SDL_GETEVENT, SDL_EVENT_GAMEPAD_REMAPPED, SDL_EVENT_GAMEPAD_REMAPPED) > 0) {
+        if (mControlDeck) {
+            mControlDeck->GetConnectedPhysicalDeviceManager()->RefreshConnectedSDLGamepads();
+            changed = true;
+        }
+    }
```
briaguya0: mentioned earlier but curious about this one. when are we getting remapped events and why do we need to refresh with them? want to make sure this is at least documented as it goes a bit beyond migration

briaguya0 (commented):
overall this is looking great! left a few comments in there

src/ship/audio/SDLAudioPlayer.cpp
mckinlee: I kept `SDL_GetAudioStreamQueued`. LUS uses this value as producer-side backpressure for data submitted with `SDL_PutAudioStreamData`, matching the role `SDL_GetQueuedAudioSize` had in the SDL2 implementation. `SDL_GetAudioStreamAvailable` reports converted output that an application can retrieve with `SDL_GetAudioStreamData`; that is not the side this playback-only stream reads. I added a short comment at the call site to make the choice explicit.

References:

- https://wiki.libsdl.org/SDL3/README-migration#sdl_audioh
- https://wiki.libsdl.org/SDL3/SDL_GetAudioStreamQueued
- https://wiki.libsdl.org/SDL3/SDL_GetAudioStreamAvailable

src/fast/backends/gfx_sdl2.cpp
mckinlee: Failure to find or apply an exclusive mode now logs the error and falls through to the normal fullscreen call with a null mode, which gives the borderless fallback. The state update and callback are no longer skipped by those failures.

include/ship/controller/physicaldevice/SDLAddRemoveDeviceEventHandler.h
mckinlee: I added remap handling while working through controller hotplug. Because SDL3 exposes `SDL_EVENT_GAMEPAD_REMAPPED` beside add/remove, I extended the existing event path to rescan an already-open controller after a mapping change and refresh ImGui controller navigation. The intent was to keep the cached mapping and gamepad name current without requiring a disconnect. It was a proactive behavior improvement, not required for SDL3, and SDL2 has the equivalent `SDL_CONTROLLERDEVICEREMAPPED` event. However, I removed that because it should be a seperate PR, if at all... but I don't see a reason to atm.

.github/workflows/tidy-format-validation.yml
mckinlee: I restored the checkout depth and clang-tidy diff command from `main`, so PR 1191 now changes only the SDL package installed by that workflow. I moved the complete-history correction into separate draft PR [1195](https://github.com/Kenix3/libultraship/pull/1195); its build, test, and tidy matrix passes.

cmake/dependencies/linux.cmake
mckinlee: No particular reason other than preference, but totally fine. I restored `find_package(SDL3 REQUIRED CONFIG COMPONENTS SDL3)`, removed the FetchContent fallback, and kept the newer SDL installation in CI.

src/fast/backends/gfx_sdl2.cpp
mckinlee: I did not run a reliable interactive mouse-capture test. I checked the SDL3 API contract: relative mouse mode now constrains the cursor and `SDL_HINT_MOUSE_RELATIVE_MODE_CENTER` defaults to enabled, while `SDL_SetWindowMouseRect` is for requesting a specific confinement rectangle. That supports removing the old 2×2 workaround, and the now-unused `mCursorClip` member is gone. Lmk how you wanna proceed.

include/ship/controller/controldevice/controller/mapping/factories/RumbleMappingFactory.h
mckinlee: Ah, yeh.. you right. That value is LUS `PhysicalDeviceType`, not an SDL gamepad subtype. The factory now returns no rumble mapping unless the type is `PhysicalDeviceType::SDLGamepad`, and the focused test covers keyboard, mouse, and SDL gamepad inputs.

src/ship/controller/physicaldevice/SDLAddRemoveDeviceEventHandler.cpp
mckinlee: The remapped event is emitted when SDL changes the mapping associated with an already connected gamepad, such as after mappings are added or reloaded. I originally refreshed the connected gamepad cache so its name and mapping state would track that change, then refreshed ImGui's gamepad list. You're right,, so I removed the remap handling and its tests rather than documenting it into this PR.

tests/connected_physical_device_manager_tests.cpp
```diff
@@ -0,0 +1,60 @@
...
+    manager.RefreshConnectedSDLGamepads();
+    const auto secondScan = manager.GetConnectedSDLGamepadsForPort(0);
+    ASSERT_TRUE(secondScan.contains(mInstanceId));
+    EXPECT_EQ(secondScan.at(mInstanceId), firstHandle);
+
+    const SDL_JoystickID detachedInstanceId = mInstanceId;
+    ASSERT_TRUE(SDL_DetachVirtualJoystick(detachedInstanceId));
+    mInstanceId = 0;
+    manager.RefreshConnectedSDLGamepads();
+    EXPECT_FALSE(manager.GetConnectedSDLGamepadsForPort(0).contains(detachedInstanceId));
```
briaguya0, line 56:
```suggestion
    EXPECT_FALSE(manager.GetConnectedSDLGamepadsForPort(0).contains(detachedInstanceId));
    EXPECT_EQ(SDL_GetGamepadFromID(detachedInstanceId), nullptr);
```

the null gamepad check was removed in https://github.com/Kenix3/libultraship/commit/2707c0758c628a67443467754b7b6923b9b349a6

if i'm understanding https://github.com/Kenix3/libultraship/issues/1192 correctly: `SDL_DetachVirtualJoystick` doesn't lead to `SDL_CloseGamepad` getting called because of _SDL_ code, but because of _our_ code, so the null gamepad check verifies we aren't leaking

include/ship/controller/controldevice/controller/mapping/factories/RumbleMappingFactory.h
```diff
@@ -46,11 +46,12 @@ class RumbleMappingFactory {
...
-     * @param portIndex          The controller port index (0-based).
+     * @param physicalDeviceType Type of physical device receiving the mapping.
+     * @param portIndex Controller port
+     * index (0-based).
```
briaguya0:
```suggestion
     * @param portIndex Controller port index (0-based).
```

include/fast/Fast3dGui.h
```diff
@@ -77,8 +77,9 @@ class Fast3dGui : public Ship::Gui {
     /**
      * @brief Forwards a platform window event to the active ImGui backend.
      *
-     * Only Fast3D backends (gfx_sdl2, gfx_dxgi) construct and dispatch WindowEvents.
-     * Callers retrieve the Gui via Window::GetGui() and dynamic_cast to Fast3dGui.
+     * Only Fast3D backends (gfx_sdl, gfx_dxgi) construct and dispatch WindowEvents.
+     * Callers retrieve the Gui via
+     * Window::GetGui() and dynamic_cast to Fast3dGui.
```
briaguya0:
```suggestion
     * Only Fast3D backends (gfx_sdl, gfx_dxgi) construct and dispatch WindowEvents.
     * Callers retrieve the Gui via Window::GetGui() and dynamic_cast to Fast3dGui.
```

include/fast/WindowEvent.h
```diff
@@ -10,9 +10,9 @@ namespace Fast {
 /**
  * @brief Platform-agnostic wrapper around a window system event.
  *
- * Constructed by the Fast3D backends (gfx_sdl2, gfx_dxgi) and passed to
- * Fast3dGui::HandleWindowEvents() so that the ImGui backend can process input
- * events without the ship layer depending on any Fast3D or platform-specific type.
+ * Constructed by the Fast3D backends (gfx_sdl, gfx_dxgi) and passed to
+ * Fast3dGui::HandleWindowEvents() so that the
+ * ImGui backend can process input events without the ship layer depending on any Fast3D or platform-specific type.
```
briaguya0:
```suggestion
 * Constructed by the Fast3D backends (gfx_sdl, gfx_dxgi) and passed to
 * Fast3dGui::HandleWindowEvents() so that the ImGui backend can process input
 * events without the ship layer depending on any Fast3D or platform-specific type.
```

src/fast/backends/gfx_sdl2.cpp
briaguya0: worth checking with @Malkierian as they did
* https://github.com/Kenix3/libultraship/pull/946

but i don't think i'm coming to the same conclusion after reading https://wiki.libsdl.org/SDL3/SDL_SetWindowRelativeMouseMode

https://wiki.libsdl.org/SDL3/SDL_SetWindowRelativeMouseMode#remarks states (emphasis mine)
> While the window has focus and relative mouse mode is enabled, the cursor is hidden, the mouse position is constrained ***to the window***

> If you'd like to keep the mouse position fixed while in relative mode you can use [SDL_SetWindowMouseRect](https://wiki.libsdl.org/SDL3/SDL_SetWindowMouseRect)().

https://github.com/Kenix3/libultraship/pull/946 specifically mentions

> The default SDL capture clip rect appeared to just be the size of the window.

as for `SDL_HINT_MOUSE_RELATIVE_MODE_CENTER`, https://wiki.libsdl.org/SDL2/SDL_HINT_MOUSE_RELATIVE_MODE_CENTER mentions that _already_ being default-on in SDL2

> By default SDL will constrain the mouse to the center of the window

so i'm thinking we should probably keep this workaround

.github/workflows/tidy-format-validation.yml
briaguya0: i'm open to just adding an exclude for `MobileImpl.h` like we do for `AppleFolderManager.mm` so tidy passes on this PR while we figure out how we want to handle https://github.com/Kenix3/libultraship/pull/1195

briaguya0 (commented):
this is looking very close to ready to land imo. few more comments, mostly little cleanup things, one continued discussion, one clang-tidy thought

include/fast/WindowEvent.h
```diff
@@ -10,9 +10,11 @@ namespace Fast {
...
  * @brief Platform-agnostic wrapper around a window system event.
  *
- * Constructed by the Fast3D backends (gfx_sdl, gfx_dxgi) and passed to
- * Fast3dGui::HandleWindowEvents() so that the
- * ImGui backend can process input events without the ship layer depending on any Fast3D or platform-specific type.
+ * Constructed by the Fast3D backends (gfx_sdl,
+ * gfx_dxgi) and passed to
+ * Fast3dGui::HandleWindowEvents() so that the ImGui backend can process input
+ * events
+ * without the ship layer depending on any Fast3D or platform-specific type.
```
briaguya0: to be clear, my issue here was that your PR is formatting these doc comment blocks very strangely, and since the words aren't actually changing it's causing noise in the diff. getting them back to what they were before this PR is the goal.

```suggestion
 * Constructed by the Fast3D backends (gfx_sdl, gfx_dxgi) and passed to
 * Fast3dGui::HandleWindowEvents() so that the ImGui backend can process input
 * events without the ship layer depending on any Fast3D or platform-specific type.
```

include/fast/Fast3dGui.h
```diff
@@ -77,9 +77,10 @@ class Fast3dGui : public Ship::Gui {
     /**
      * @brief Forwards a platform window event to the active ImGui backend.
      *
-     * Only Fast3D backends (gfx_sdl, gfx_dxgi) construct and dispatch WindowEvents.
-     * Callers retrieve the Gui via
-     * Window::GetGui() and dynamic_cast to Fast3dGui.
+     * Only Fast3D backends (gfx_sdl,
+     * gfx_dxgi) construct and dispatch WindowEvents.
+     * Callers retrieve the Gui via Window::GetGui() and
+     * dynamic_cast to Fast3dGui.
```
briaguya0: to be clear, my issue here was that your PR is formatting these doc comment blocks very strangely, and since the words aren't actually changing it's causing noise in the diff. getting them back to what they were before this PR is the goal.

```suggestion
     * Only Fast3D backends (gfx_sdl, gfx_dxgi) construct and dispatch WindowEvents.
     * Callers retrieve the Gui via Window::GetGui() and dynamic_cast to Fast3dGui.
```

include/ship/controller/controldevice/controller/mapping/factories/RumbleMappingFactory.h
```diff
@@ -46,11 +46,13 @@ class RumbleMappingFactory {
...
-     * @param portIndex Controller port
-     * index (0-based).
+     * @param physicalDeviceType
+     * Type of physical device receiving the mapping.
+     * @param portIndex Controller port index (0-based).
+     *
      * @param consoleVariable    ConsoleVariable for persisting mappings.
-     * @param controlDeck        ControlDeck for physical device access.
+     * @param controlDeck        ControlDeck
+     * for physical device access.
```
briaguya0: This one does have actual text changes worth keeping, but the formatting is all messed up now. We shouldn't need to change the formatting here.

```suggestion
     * @param physicalDeviceType Type of physical device receiving the mapping.
     * @param portIndex Controller port index (0-based).
     * @param consoleVariable    ConsoleVariable for persisting mappings.
     * @param controlDeck        ControlDeck for physical device access.
```

briaguya0 (commented):
looks great! on to final nit territory, only comments left are about the doc comment formatting being odd/causing diff noise

include/ship/controller/controldevice/controller/mapping/factories/RumbleMappingFactory.h
briaguya0: resolving because i have an updated version in the newest review

include/fast/Fast3dGui.h
briaguya0: resolving because i have an updated version in the newest review

include/fast/WindowEvent.h
briaguya0: resolving because i have an updated version in the newest review

include/ship/controller/controldevice/controller/mapping/factories/RumbleMappingFactory.h
```diff
@@ -45,12 +45,11 @@ class RumbleMappingFactory {
                                   std::shared_ptr<ControlDeck> controlDeck);
 
     /**
-     * @brief Creates default SDL rumble mappings for a specific physical device type.
-     * @param physicalDeviceType The type of physical device (e.g., SDL2 gamepad, Nintendo controller).
-     * @param portIndex          The controller port index (0-based).
+     * @param physicalDeviceType Type of physical device receiving the mapping.
```
briaguya0: looks like my previous suggestion led to the `@brief` getting dropped, my bad

```suggestion
     * @brief Creates default SDL rumble mappings for a specific physical device type.
     * @param portIndex Controller port index (0-based).
```

include/ship/controller/controldevice/controller/mapping/factories/RumbleMappingFactory.h
```diff
@@ -46,11 +46,11 @@ class RumbleMappingFactory {
 
     /**
      * @brief Creates default SDL rumble mappings for a specific physical device type.
-     * @param physicalDeviceType The type of physical device (e.g., SDL2 gamepad, Nintendo controller).
-     * @param portIndex          The controller port index (0-based).
+     * @param portIndex Controller port index (0-based).
+     * @param portIndex Controller port index (0-based).
```
briaguya0, line 50:
```suggestion
     * @brief Creates default SDL rumble mappings for a specific physical device type.
     * @param physicalDeviceType Type of physical device receiving the mapping.
     * @param portIndex Controller port index (0-based).
```

briaguya0 (approved):
:shipit:

MegaMech (approved):

Kenix3 (approved):

PR #1193, opened 2026-08-01
XML Support for segmented addresses in SPVertex

SlimyGooblin:
Needed to support XML imports for skinLimb meshes

Not sure how I managed to delete my branch, forth time's the charm I guess..

### DIFF
--- a/src/fast/resource/factory/DisplayListFactory.cpp
+++ b/src/fast/resource/factory/DisplayListFactory.cpp
@@ -490,16 +490,21 @@ ResourceFactoryXMLDisplayListV0::ReadResource(std::shared_ptr<Ship::File> file,
             std::string fName = child->Attribute("Path");
             // fName = ">" + fName;
 
-            char* str = (char*)malloc(fName.size() + 1);
-            dl->Strings.push_back(str);
-            strcpy((char*)str, fName.data());
+            if (fName[0] == '>' && fName[1] == '0' && (fName[2] == 'x' || fName[2] == 'X')) {
+                uint32_t seg = std::stoul(fName.substr(1), nullptr, 16);
+                g = { gsSPVertex(seg | 1, child->IntAttribute("Count"), child->IntAttribute("VertexOffset")) };
+            } else {
+                char* str = (char*)malloc(fName.size() + 1);
+                dl->Strings.push_back(str);
+                strcpy((char*)str, fName.data());
 
-            g = GsSpVertexOtR2P1(str);
+                g = GsSpVertexOtR2P1(str);
 
-            dl->Instructions.push_back(g);
+                dl->Instructions.push_back(g);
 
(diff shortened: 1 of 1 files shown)


src/fast/resource/factory/DisplayListFactory.cpp
```diff
@@ -490,16 +490,21 @@ ResourceFactoryXMLDisplayListV0::ReadResource(std::shared_ptr<Ship::File> file,
...
-            char* str = (char*)malloc(fName.size() + 1);
-            dl->Strings.push_back(str);
-            strcpy((char*)str, fName.data());
+            if (fName[0] == '>' && fName[1] == '0' && (fName[2] == 'x' || fName[2] == 'X')) {
+                uint32_t seg = std::stoul(fName.substr(1), nullptr, 16);
+                g = { gsSPVertex(seg | 1, child->IntAttribute("Count"), child->IntAttribute("VertexOffset")) };
+            } else {
+                char* str = (char*)malloc(fName.size() + 1);
+                dl->Strings.push_back(str);
+                strcpy((char*)str, fName.data());
```
Kenix3: We should definutely use a `strncpy` here where it is bounded. I know that the size of the buffer is guaranteed to be OK in this case, but it is good code hygiene to apply security in each place.
                `strncpy((char*)str, fName.data(), fName.size() + 1);`

SlimyGooblin: fixed with: db4568a3c99abd4d23ca691ae99d2896afedcf6b

src/fast/resource/factory/DisplayListFactory.cpp
```diff
@@ -490,16 +490,21 @@ ResourceFactoryXMLDisplayListV0::ReadResource(std::shared_ptr<Ship::File> file,
...
-            char* str = (char*)malloc(fName.size() + 1);
-            dl->Strings.push_back(str);
-            strcpy((char*)str, fName.data());
+            if (fName[0] == '>' && fName[1] == '0' && (fName[2] == 'x' || fName[2] == 'X')) {
+                uint32_t seg = std::stoul(fName.substr(1), nullptr, 16);
+                g = { gsSPVertex(seg | 1, child->IntAttribute("Count"), child->IntAttribute("VertexOffset")) };
+            } else {
+                char* str = (char*)malloc(fName.size() + 1);
+                dl->Strings.push_back(str);
+                strncpy((char*)str, fName.data(), fName.size() + 1);
```
Kenix3, line 499: Hey, back again. Here's an even better suggestion.
```suggestion
                snprintf(str,  fName.size(), "%s", fName.data());
```

SlimyGooblin: 6c925d8

Sorry, would've done this earlier but I ended up getting sick last week

PR #1194, opened 2026-08-01
Updates Versioning documentation to cover epoch semver.

Kenix3:


briaguya0 (commented):
this doesn't quite explain why epochs are needed. main being unstable is common for lots of projects that don't use epochs. what epochs buy us is a way to differentiate between "this is a breaking change because the signature of one function in the API changed" and "almost none of the places you're using LUS in your port will continue to work without changes"

Kenix3:
> this doesn't quite explain why epochs are needed. main being unstable is common for lots of projects that don't use epochs. what epochs buy us is a way to differentiate between "this is a breaking change because the signature of one function in the API changed" and "almost none of the places you're using LUS in your port will continue to work without changes"

```
When LUS goes through major refactors, `main` may be unstable while work progresses in a new epoch. The previous epoch then enters maintenance mode so fixes and selected features can continue to land there until the new epoch stabilizes.
```
I disagree, I think this explicitly covers the concern.

briaguya0:
> When LUS goes through major refactors, `main` may be unstable while work progresses in a new epoch. The previous epoch then enters maintenance mode so fixes and selected features can continue to land there until the new epoch stabilizes.

the main missing puzzle piece here is that it doesn't mention "without blocking breaking changes from landing in the previous epoch" - that's the big difference.

with `X.Y.Z` we could have
> When LUS goes through major refactors, `main` may be unstable while work progresses in a new major version. The previous major version then enters maintenance mode so fixes and selected features can continue to land there until the new major version stabilizes.

but in that case we'd be blocked from landing any breaking changes as those would require a new major

Malkierian:
I think I agree with bria here. The wording, from my perspective, doesn't really give any difference functionally between major and epoch, at least as far as LUS's implementation is concerned. TBF, though, looking at the link provided in the new docs, I didn't fully understand there, either.

docs/INDEX.md
```diff
@@ -27,7 +29,11 @@ Please see the [Contributing](contributing) page for more information.
 
 ## Versioning
 
-We use semantic versioning. We have defined the API as: every C linkage function, variable, struct, class, public class method, or enum included from `libultraship.h`.
+We use [epoch semantic versioning](https://antfu.me/posts/epoch-semver). We have defined the API as: every C linkage function, variable, struct, class, public class method, or enum included from `libultraship.h`.
+
+When LUS goes through major refactors, `main` may be unstable while work progresses in a new epoch. The previous epoch then enters maintenance mode so fixes and selected features can continue to land there until the new epoch stabilizes.
```
briaguya0: > `main` may be unstable while work progresses in a new epoch

this feels bad to me, mostly because "unstable" can mean different things in different contexts. for some readers it may imply "broken" (which kinda feels like the intended meaning here), but in _versioning_ terms it just means "can get breaking API changes".

i think it'd be more clear to spell it out. during major refactors `main` may not be in a production-ready state, and there may not be a clear migration path for ports.

docs/VERSIONING.md
```diff
@@ -0,0 +1,57 @@
...
+## Why epochs
+
+LUS is a library undergoing active architectural improvements. From time to time, we perform large refactors that touch core systems - context initialization, component lifecycle, resource management, or event dispatch.
+
+### The problem with traditional semver
+
+With standard semantic versioning, a breaking change forces a major version bump (for example, `1.x.x` to `2.0.0`). This creates a dilemma:
+
+- **Option A**: Keep breaking changes on `main`, making it unstable for months while the refactor lands incrementally.
+  - **Result**: Port maintainers cannot safely pull fixes or features during the refactor period.
```
briaguya0: this doesn't clearly communicate the problem. this assumes no breaking changes landing on `main` during the refactor period are triggering releases. it's really "land breaking changes on `main` and hold off tagging a new major release until some undefined moving target criteria is met"

the real issue with "just refactor on main" as a strategy is that the commits landing on `main` mid-refactor aren't ready for anyone to use (not just that it's hard to migrate to, but that it likely doesn't make _sense_ to migrate to because they might change again and they might not be fully tested)

"breaking changes" doesn't communicate this well. changing one param in an API sig is a breaking change, but it's easy for ports to understand and migrate to. reworking the entire component system and having commits halfway through is a whole different thing.

docs/VERSIONING.md
```diff
@@ -0,0 +1,57 @@
...
+
+- **Option B**: Hold all refactor work on a long-lived feature branch.
+  - **Result**: The branch diverges significantly, merge conflicts accumulate, and collaboration becomes difficult.
```
briaguya0: > The branch diverges significantly, merge conflicts accumulate, and collaboration becomes difficult.

this will _still_ happen with epochs, the difference is `main` will be the "long-lived feature branch" and the previous epoch will also get active development.

the problem epochs solve is the inability to put out _any_ breaking changes on the pre-refactor version of the code while waiting _years_ for a refactor to complete.

docs/VERSIONING.md
```diff
@@ -0,0 +1,57 @@
...
+
+- **Option C**: Release frequent major versions (`1.0` to `2.0` to `3.0` in quick succession).
+  - **Result**: Version numbers become less meaningful, and users face constant migration churn.
```
briaguya0: this also misses the point. the issue isn't version numbers lacking meaning, it's that breaking changes landing mid-refactor aren't things we should be trying to support. for major refactors we need a place to land breaking changes without worrying about releasing/supporting those breaking changes, and a way to say "this is now the refactored version of the codebase we are going to support."

docs/VERSIONING.md
```diff
@@ -0,0 +1,57 @@
...
+- Clearly signals architectural dividing lines
+
+### When to expect a new epoch
+
+A new epoch typically begins when:
+- Core architectural patterns change (for example, context/component model redesign)
+- Multiple subsystems require coordinated breaking changes
+- The refactor is expected to span several months of incremental work
+
+Minor breaking changes within an epoch still follow semantic versioning rules (bump `major` within that epoch).
```
briaguya0: > Minor breaking changes

i'd say "small" or "targeted" or something other than "minor" here as "minor" in the context of versioning is tightly-coupled to "non-breaking."

docs/VERSIONING.md
```diff
@@ -0,0 +1,57 @@
...
+- **During a major refactor**: `main` moves to a new epoch (for example, `2.x.x.x`), where breaking changes land incrementally.
+- **The previous epoch** (for example, `1.x.x.x`) continues as a maintained branch:
+  - Critical fixes backported
+  - Selected features may be cherry-picked
+  - Stable baseline for ports not yet ready to migrate
+
+This approach:
+- Keeps `main` moving forward without long-lived feature branches
+- Provides a stable baseline for ports during transition periods
+- Avoids rapid major version churn
```
briaguya0: i want to steer away from this. there's nothing wrong with `2.23.0.1` or something (meaning 23 breaking changes landed in the `2.x.x.x` epoch) assuming each of the 23 breaking changes landed with a clear migration path for ports and with the expectation that the code is production-ready.

docs/VERSIONING.md
```diff
@@ -0,0 +1,57 @@
...
+
+This replaces the previous epoch rule that treated every C linkage function, variable, struct, class, public class method, or enum included from `libultraship.h` as part of the versioned API.
+
+## Why epochs
+
+LUS is a library undergoing active architectural improvements. From time to time, we perform large refactors that touch core systems - context initialization, component lifecycle, resource management, or event dispatch.
+
+### The problem with traditional semver
+
+With standard semantic versioning, a breaking change forces a major version bump (for example, `1.x.x` to `2.0.0`). This creates a dilemma:
```
briaguya0: the issues are more:
* how do ports know if 4 -> 5 means one arg changed or the whole codebase works completely differently?
* when a refactor is in-progress, how can a port land a small breaking change in LUS?
  * if we only have a branch supporting the most recent released major version, and the refactor is going to land as the _next_ major version, then only non-breaking changes can land during the entire refactor process. this makes things very difficult as there are often times when the best way to solve a problem requires a small breaking change,

briaguya0 (commented):
this definitely feels better, i still want to make sure we hit all the right notes though. left some comments.

docs/INDEX.md
briaguya0: addressed by https://github.com/Kenix3/libultraship/pull/1194/commits/5f5a7b15320288e023d12297900e7ae4208165de

docs/VERSIONING.md
briaguya0: addressed by https://github.com/Kenix3/libultraship/pull/1194/commits/5f5a7b15320288e023d12297900e7ae4208165de

docs/VERSIONING.md
briaguya0: addressed by https://github.com/Kenix3/libultraship/pull/1194/commits/5f5a7b15320288e023d12297900e7ae4208165de

docs/VERSIONING.md
briaguya0: addressed by https://github.com/Kenix3/libultraship/pull/1194/commits/5f5a7b15320288e023d12297900e7ae4208165de

docs/VERSIONING.md
briaguya0: addressed by https://github.com/Kenix3/libultraship/pull/1194/commits/5f5a7b15320288e023d12297900e7ae4208165de

docs/VERSIONING.md
briaguya0: addressed by https://github.com/Kenix3/libultraship/pull/1194/commits/5f5a7b15320288e023d12297900e7ae4208165de

docs/VERSIONING.md
briaguya0: resolved by https://github.com/Kenix3/libultraship/pull/1194/commits/5f5a7b15320288e023d12297900e7ae4208165de

briaguya0 (approved):
:shipit:

PR #1195, opened 2026-08-01
Check complete pull requests with clang-tidy

mckinlee:
## Summary

- run clang-tidy against the complete pull-request range
- keep changed headers in analysis by forcing C++20 parsing
- exclude only the C translation unit `src/ship/utils/glob.c`

## Testing

- parsed the workflow with PyYAML and verified its diff-selection contract
- verified the SDL3 pull-request range selects 15 changed headers, including `MobileImpl.h`, while excluding C inputs
- analyzed `MobileImpl.h` through the LLVM 14 diff driver and a real compile database with the C++20 overrides

### DIFF
--- a/.github/workflows/tidy-format-validation.yml
+++ b/.github/workflows/tidy-format-validation.yml
@@ -9,7 +9,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          fetch-depth: 2
+          fetch-depth: 0
       - name: Install dependencies
         run: |
           sudo apt-get update
@@ -53,7 +53,7 @@ jobs:
           mkdir clang-tidy-result
       - name: Analyze
         run: |
-          git diff -U0 HEAD^ -- 'src' 'include' ':!include/libultraship/color.h' ':!include/ship/utils/color.h' ':!include/libultraship/libultra/*' ':!src/libultraship/libultra/*' ':!src/fast/*' ':!include/fast/*' ':!src/ship/utils/AppleFolderManager.mm' | clang-tidy-diff -p1 -path build -export-fixes clang-tidy-result/fixes.yml
+          git diff -U0 "${{ github.event.pull_request.base.sha }}"...HEAD -- 'src' 'include' ':!include/libultraship/color.h' ':!include/ship/utils/color.h' ':!include/libultraship/libultra/*' ':!src/libultraship/libultra/*' ':!src/fast/*' ':!include/fast/*' ':!src/ship/utils/AppleFolderManager.mm' | clang-tidy-diff -p1 -path build -regex='.*\.(cpp|cc|cxx|mm)$' -export-fixes clang-tidy-result/fixes.yml
       - name: Save PR metadata
         run: |
           echo ${{ github.event.number }} > clang-tidy-result/pr-id.txt
--- a/.github/workflows/tidy-format-validation.yml
+++ b/.github/workflows/tidy-format-validation.yml
@@ -53,7 +53,7 @@ jobs:
           mkdir clang-tidy-result
       - name: Analyze
         run: |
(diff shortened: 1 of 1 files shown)


briaguya0:
> [!NOTE]
> This comment was generated by Claude, driven by @briaguya0.

I went digging into the `MobileImpl.h` failure on #1191 and reproduced it locally, and I think the `-regex` change has a side effect worth knowing about before this lands.

## `-regex` removes header linting entirely, not just for header-only PRs

`clang-tidy-diff.py` runs **one clang-tidy invocation per changed file**, and the `-line-filter` for each invocation names only that one file ([`clang-tidy-diff.py:395-402`](https://github.com/llvm/llvm-project/blob/main/clang-tools-extra/clang-tidy/tool/clang-tidy-diff.py)):

```python
for name in lines_by_file:
    line_filter_json = json.dumps(
        [{"name": name, "lines": lines_by_file[name]}], separators=(",", ":")
    )
    ...
    command.append("-line-filter=" + line_filter_json)
```

So a header is never covered via the `.cpp` that includes it — the line filter for that `.cpp` run lists only the `.cpp`, and diagnostics in the header get dropped. Confirmed with a minimal repro:

| line-filter contains | header warning |
|---|---|
| `main.cpp` and `mylib.h` | reported |
| `main.cpp` only | suppressed — *"1 due to line filter"* |

The only path by which a changed header gets linted today is being its own direct input. `-regex='.*\.(cpp\|cc\|cxx\|mm)$'` removes headers from `lines_by_file`, so they get no invocation of their own *and* no line-filter entry anywhere else. Net effect is that no header is linted in any PR, not just header-only ones.

## Root cause

Headers are never in `compile_commands.json` (0 of 171 entries here). clang-tidy interpolates a command from the nearest known TU by path and **inherits that TU's language and standard**. The DB is where it goes wrong — of 171 entries, **162 carry no `-std` at all**, 4 are `-std=gnu11`, 5 are `-std=gnu++20`.

For `include/ship/port/mobile/MobileImpl.h` the match lands on a C file, so it's parsed as C. Stripping the 5 `.c` entries from a copy of the DB makes the error vanish:

| compilation database | result |
|---|---|
| full `build/` (171 entries) | `error: unknown type name 'namespace'` ×2 |
| all 5 `.c` entries removed (166) | clean |

## Blast radius is wider than this one header

Sweeping all 222 headers with `clang-tidy -p build`, 13 fail standalone. Seven are already covered by the existing pathspec exclusions (`:!include/fast/*`, `:!include/libultraship/libultra/*`). **Six are in scope:**

| header | error |
|---|---|
| `include/ship/port/mobile/MobileImpl.h` | `unknown type name 'namespace'` — parsed as C |
| `include/libultraship/classes.h` | ↓ |
| `include/libultraship/libultraship.h` | ↓ |
| `include/ship/events/CoreEvents.h` | ↓ |
| `include/ship/events/Events.h` | `Events.h:167: no member named 'contains'` — pre-C++20 std |
| `include/ship/events/EventSystem.h` | ↓ |

Two flavors of the same bug: wrong language, or wrong standard. So a narrow pathspec exclusion isn't great either — it'd be six today and would grow.

## Suggested alternative

Forcing both on the clang-tidy side fixes the root cause and keeps header coverage:

| flag | `MobileImpl.h` | `Events.h` |
|---|---|---|
| `-extra-arg=-std=gnu++20` | ✗ | ✓ |
| `-extra-arg=-xc++` | ✓ | ✗ |
| **both** | ✓ | ✓ |

`clang-tidy-diff.py` already forwards `-extra-arg` (`:218`, `:385`). Three sampled `.cpp` files stay clean with the flags on.

One catch: `src/ship/utils/glob.c` goes 0 → 10 errors under `-xc++`, and it's in scope. It's the only real C file under `src`/`include` (the other four are vendored in `build/_deps`), so it'd need `':!src/ship/utils/glob.c'` added to the pathspec — one exclusion, for a file C++ checks can't meaningfully lint anyway.

Which gives:

| approach | headers linted | cost |
|---|---|---|
| `-regex` (this PR) | none, ever | all header linting silently disappears |
| pathspec exclusions | all but 6 | whack-a-mole, grows over time |
| `-extra-arg` + exclude `glob.c` | all | one C file dropped from tidy |

## Caveats

- I ran clang-tidy **22.1.8** (Fedora); CI runs **14**. The mechanism should hold, but the exact failing set may differ.
- I configured without `-DENABLE_SCRIPTING=ON`, which CI uses, so the `scripting/` headers weren't tested against their real TUs.
- Everything above is about the `-regex` half of this PR. The `fetch-depth: 0` + `base.sha...HEAD` half is a separate concern, and splitting the two would let whichever fix you land go in without waiting on the scope discussion.

Issue #1152, opened 2026-07-17
port-maintenance cherry-pick tracking issue


briaguya0:
Tracks commits that have landed on `port-maintenance` but not yet on `main`.

For each entry, the **port-maintenance** column links the commit and PR as merged into `port-maintenance`. Where a change has already been cherry-picked onto `main`, the **main** column links the equivalent commit and PR there.

## Needs cherry-pick to `main`

- [ ] **Unify the pyramidlike test** — [`62e973a`](https://github.com/Kenix3/libultraship/commit/62e973aeb4a53ad4d22bb91e2d9373ecdfcd246c) · [#1239](https://github.com/Kenix3/libultraship/pull/1239)
  - https://github.com/Kenix3/libultraship/pull/1278
- [ ] **Use atomic write for `Config::Save`** — [`d4f9b84`](https://github.com/Kenix3/libultraship/commit/d4f9b849dfd6c1c04ee54b0d28ff5929ba02261b) · [#1183](https://github.com/Kenix3/libultraship/pull/1183)
  - https://github.com/Kenix3/libultraship/pull/1277
- [ ] **Fix Windows crash handler producing unusable tracebacks** — [`cdba319`](https://github.com/Kenix3/libultraship/commit/cdba319b8668fd2882d3a5e083e5bb98cbbee2be) · [#1190](https://github.com/Kenix3/libultraship/pull/1190)
  - https://github.com/Kenix3/libultraship/pull/1276
- [ ] **Fixes/safety for shader push/pop** — [`7f9b86a`](https://github.com/Kenix3/libultraship/commit/7f9b86a593c526fc42261d7fe197100cecf57178) · [#1247](https://github.com/Kenix3/libultraship/pull/1247)

## Needs port to `main`

- [ ] **Untangle various parts from the Context destructor** — [`1d98291`](https://github.com/Kenix3/libultraship/commit/1d9829140444268827a1369095bcc93518ff3b4d) · [#1103](https://github.com/Kenix3/libultraship/pull/1103)
  - https://github.com/Kenix3/libultraship/issues/1196
- [ ] **Add Palette4bpp Support for Texture Loading into ImGui** — [`2bfbde3`](https://github.com/Kenix3/libultraship/commit/2bfbde3a72c119f8073ad762ec6be131dff5df66) · [#1157](https://github.com/Kenix3/libultraship/pull/1157)
  - https://github.com/Kenix3/libultraship/issues/1201
[...]

PR #1197, opened 2026-08-02
Fix warnings (cherry-pick #1126)

briaguya0:
Cherry-picks #1126 (`2cc049d`) from `port-maintenance` onto `main`. Authored by @Pepe20129; applies with no conflicts.

Both warnings are still present on `main` verbatim, and both are in headers, so they propagate into port builds:

- `include/fast/lus_gbi.h:137` — `GIMMCMD` is defined with no redefinition guard, which warns in ports whose own `gbi.h` already defines it.
- `include/ship/utils/binarytools/BitConverter.h:244,249` — `ByteSwap` passes `size_t` expressions (`pos * 2`, `pos * 4`) to `ToUInt16BE`/`ToUInt32BE`, which take `int32_t offset` (`:90`, `:116`). Now explicitly cast.

Part of #1152.

### DIFF
--- a/include/fast/lus_gbi.h
+++ b/include/fast/lus_gbi.h
@@ -134,6 +134,9 @@ constexpr int8_t RDP_G_TRI1_WIDE = OPCODE(0x49);
 
 /* macros for command parsing: */
 #define GDMACMD(x) (x)
+#ifdef GIMMCMD
+#undef GIMMCMD
+#endif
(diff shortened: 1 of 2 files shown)


Pepe20129 (approved):
Whoops forgot to PR this to main after it landed in port-maintenance

serprex (approved):

Kenix3 (approved):

PR #1212, opened 2026-08-02
Fix wasapi hang

### LINKED ISSUE #1211: SoH & 2Ship Hanging on boot (Audio Related)
Something with the audio changes recently on LUS seem to have affected both zelda ports. It's unclear if this is an LUS issue or if this was a bad LUS bump on the port side, but both have similar behavior.

Booting up a clean install SoH & 2Ship on Windows works fine for the first boot, then if you close and re-open it, it will just hang on boot, until you either reboot your computer or hand edit the JSON to use the SDL audio driver instead of Windows Audio Session API.

My hunch is that something is not tearing down properly for the audio session API, which is preventing the further reboots from working, but I don't know much about audio nor do I have a windows dev env to fix this on.

serprex:
Destructor needs to DoClose to stop stream & release client

Fixes #1211

### DIFF
--- a/include/ship/audio/AudioPlayer.h
+++ b/include/ship/audio/AudioPlayer.h
@@ -38,7 +38,7 @@ class AudioPlayer {
      */
     AudioPlayer(AudioSettings settings) : mAudioSettings(settings) {
     }
-    ~AudioPlayer();
+    virtual ~AudioPlayer();
 
     /**
      * @brief Calls DoInit() and sets the initialised flag on success.
--- a/include/ship/audio/WasapiAudioPlayer.h
+++ b/include/ship/audio/WasapiAudioPlayer.h
@@ -30,9 +30,15 @@ class WasapiAudioPlayer : public AudioPlayer, public IMMNotificationClient {
     }
 
     /**
+     * @brief Closes the stream and unregisters the endpoint notification callback.
+     */
+    ~WasapiAudioPlayer();
+
+    /**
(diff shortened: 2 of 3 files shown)


serprex:
need someone on Windows to test

include/ship/audio/AudioPlayer.h
```diff
@@ -38,7 +38,7 @@ class AudioPlayer {
      */
     AudioPlayer(AudioSettings settings) : mAudioSettings(settings) {
     }
-    ~AudioPlayer();
+    virtual ~AudioPlayer();
```
MegaMech, line 41: are you sure you want virtual here?

MegaMech (changes requested):

MegaMech:
Tested on Windows --> no issues found

include/ship/audio/AudioPlayer.h
serprex: https://github.com/Kenix3/libultraship/blob/a57f6b5ec8dd1c4a02bf2b8481e20f2441135129/src/ship/audio/SDLAudioPlayer.cpp#L6

other audio backends are already overriding destructor, so yes

MegaMech: looked it up.

You add virtual to the base class deconstructor. But not to the inherit classes or something like that.

lgtm

MegaMech:
Note that this was tested on `main` not port-maintenance.

PR #1215, opened 2026-08-03
[main 2.0.0] Remove Default ResourceManager nullptr from Archive.h

MegaMech:
Creating Archive class without a Resource Manager crashes the game and it's difficult to track the crash down. I recommend finding a solution.

I think if you require it then the user won't be able to compile and will find the issue....

However, perhaps there's a scenario where you don't need the resource manager. Perhaps an overload is better? But in that case I could see the same user issue still occurring...

Perhaps instead we add warning if an archive did not have its resource manager set? Maybe it should use `ShipCompat::GetResourceManager()`?


Kenix3:
I'm thinking instead of forcing the user to supply a ResourceManager, the Archive should be able to operate with separate code paths when the ResourceManager is unavailable.

MegaMech:
Under what scenario would the resource manager be unavailable? Unless you want to give this one a go, I would need more context?

Because I think with your idea, we would need to check if the resource manager is unavailable and then fail if it's an archive that cannot work without a resource manager. I don't really understand how we would do that except right from archive construction?

PR #1187, opened 2026-07-31
Refactors ResourceInitData and ResourceIdentifier together.

Kenix3:
Resolves #762

Issue #1160, opened 2026-07-22
Deprecate Binary Resource Header


Kenix3:
- Remove the binary header support complete.
- Keep the JSON meta file support.
- When a `ResourceInitData` is not passed in _and_ a JSON meta file is not found, use these defaults:
  - Read `ByteOrder` from the `Resource.InitData.Archive` instead of the `Resource.InitData` directly.
  - Read `IsCustom` from the `Resource.InitData.Archive` instead of the `Resource.InitData`
  - Read `Type` from the `Resource.Identifier.Path` as the file extension. (Today this is `Resource.InitData.Path`, but we're working on changes to move that data)
  - Read `ResourceVersion` from the `Resource.Identifier.Path`. Either as a double extension, or just separate extensions per version.
  - `Format` can be automatically determined as well, but this is the one we didn't discuss and is still a question of how to handle. I guess we can attempt to parse it as XML, and if that fails assume it's binary.
- The `Archive` can have a metadata file (like we have for version) to describe the `Archive`'s ByteOrder and `IsCustom` state. I think if we can't find that metadata file in the Archive, we can assume `ByteOrder` is system byte order and let it explode. We can assume is `IsCustom` is false.

PR #1218, opened 2026-08-03
Resolve .meta aliases by archive priority (port #1168 to main)

### LINKED ISSUE #984: `.meta` files no longer work

### LINKED ISSUE #1165: `.meta` alias should fall back to the real same-path asset when its target is missing

### LINKED ISSUE #1208: Port #1168 to main: resolve .meta aliases by archive priority
Sub-issue of #1152.

[#1168](https://github.com/Kenix3/libultraship/pull/1168) ("Resolve .meta aliases by archive priority", [`9a1a2bd`](https://github.com/Kenix3/libultraship/commit/9a1a2bd9356d3de18f54ef220a413da18d16779b)) reworks `.meta` alias resolution so a request for `X` loads the highest-priority provider among the real asset at `X` and the target of any `X.meta`, ranked by the archive each **asset** lives in (last-loaded wins; ties go to the alias).

It fixes #984 and resolves #1165 on `port-maintenance`. Both bugs should be re-confirmed against `main` before porting — `main`'s meta handling is not the same code (see below), so it may not exhibit them identically.

## Why it cannot be cherry-picked

`main`'s tip commit #1187 ("Refactors ResourceInitData and ResourceIdentifier together") reworked exactly this code. `git cherry-pick 9a1a2bd9` gives **2 conflicts** across 7 files.

The split is clean, and worth knowing before starting:

### Applies with no conflict — the archive-priority plumbing

- `Archive::GetPriority` / `Archive::SetPriority` (`src/ship/resource/archive/Archive.cpp`, `include/.../Archive.h`)
- `ArchiveManager::GetFilePriority` (`src/ship/resource/archive/ArchiveManager.cpp`, `include/.../ArchiveManager.h`)
- The `Archive::IndexFile` fix — stop stripping the `.meta` suffix so a real `foo` and its `foo.meta` stay distinct entries. **This is the root cause of #984**, and it lands cleanly.
- `include/ship/resource/ResourceLoader.h`

### Conflicts — the resolution logic

`src/ship/resource/ResourceLoader.cpp` and `src/ship/resource/ResourceManager.cpp`, both for the same reason:

| | `port-maintenance` | `main` |
| --- | --- | --- |
| `ResourceLoader::LoadResource` | `std::string filePath` | `const ResourceIdentifier&` |
| `ResourceManager` lookup | `identifier.Path` | `identifier.GetPath()` / `IsPath()` / `GetPathHash()` |
| meta init data | plain path | `metadataIdentifier` with `SetOwner` / `SetParent` propagation |

**`main` already has its own `.meta` handling** inside `LoadResource`: it loads `filePath + ".meta"`, builds a `metadataIdentifier` from it, copies the requesting identifier's owner and parent onto it, and reloads the file through that identifier. #1168 replaces that with `ResolveMetaAlias` plus a priority comparison.

So this is not a missing feature dropping into a blank slate — it is two different implementations of meta resolution that have to be reconciled. The port has to keep `main`'s owner/parent propagation working while adding priority-based provider selection, and `port-maintenance`'s version has no equivalent of that propagation to copy from.

`main` does keep a `std::string filePath` compatibility overload (`include/ship/resource/ResourceLoader.h:73`), but the identifier form is the primary one and is what `ResourceManager` calls.

## Other required changes

- `ResolveMetaAlias` uses `Context::GetRawInstance()->GetResourceManager()`. That API is from #1103, which we are not porting (see #1196). `ResourceLoader` on `main` already holds `mResourceManager`.
- `ResourceManager::LoadResourceProcess`'s new guard calls `mArchiveManager->HasFile(identifier.Path + ".meta")`. `main` has `HasFile` in both path and hash forms (`include/ship/resource/archive/ArchiveManager.h:135,142`), so this has a home — but it needs the identifier accessors, and a hash-identified resource has no path to append `.meta` to. Decide what that case should do.

## The commit bundles four separable changes

Each is independently reviewable and could land as its own commit:

1. **Priority-based alias resolution** — the headline change: `ResolveMetaAlias`, `GetFilePriority`, the `IndexFile` fix, the `LoadResourceProcess` guard.
2. **`SetBufferOffset`** — a `.meta` alias can point at a binary resource that still carries its OTR header. The header path self-positions past it, but init data supplied by a caller or a `.meta` does not, so the reader started at byte 0 and read the header as body (garbage limb counts, `vector::reserve` throws). Detects a still-present header and advances `File::BufferOffset`, gated on `legacyInitData` so headerless buffers stay at 0. **Explicitly transitional — retires with the header itself under #1160.**
3. **`isCustom` from meta JSON** — a resource resolved through an alias built its init data from JSON and never set `IsCustom`, so it always reported false. Read it from the meta, defaulting false so aliasing a vanilla asset stays non-custom.
4. **Evict alias resources on directory unload** — an alias resource is cached under its base path, which has no archive file of its own, so a directory unload never evicts it while the real files it depends on are. It then survives stale, pointing at freed dependencies, and is never reloaded because the next request finds it cached. Fix: when unloading a `.meta` sidecar, also evict the resource at its base path.

Item 2 in particular is worth deciding on separately given #1160 will remove the header anyway.

## Verification note

The original was verified against the SoH meta-alias-testing kit — all single and layered cases matching expected results. That verification was against `port-maintenance`'s string-path model. Because the port has to work in terms of `ResourceIdentifier` including owner/parent propagation, **the kit should be re-run against the ported version** rather than assuming the result carries over.

briaguya0:
Ports #1168 ("Resolve `.meta` aliases by archive priority") onto `main`, per #1208.

Not a cherry-pick. `main`'s `.meta` handling is different code — #1187 reworked `ResourceInitData`
and `ResourceIdentifier` together — so `git cherry-pick 9a1a2bd9` conflicts across
`ResourceLoader.cpp` and `ResourceManager.cpp`. This is a reimplementation against the identifier
model.

## What changes

A request for `X` loads whichever of {the file at `X`} and {the target its `.meta` names} sits in
the higher-priority archive, ranked by the archive holding each **asset** rather than the one
holding the `.meta`. Ties go to the `.meta`, since an archive shipping both is being specific about
which it means.

That replaces "the `.meta` always wins", which broke two ways:

- a `.meta` whose target was absent nulled out the file to load and failed outright instead of
  falling back (#1165)
- an archive shipping only a `.meta` claimed a path it had no file for, shadowing real assets in
  every archive below it (#984)

## Commits

| | | |
|---|---|---|
| 1 | Index `.meta` files under their literal name | #1208 item 1 — root cause of #984 |
| 2 | Track archive load-order priority | item 1 |
| 3 | Resolve `.meta` aliases by archive priority | item 1 |
| 4 | Skip the OTR header for binary `.meta` alias targets | item 2 |
| 5 | Let a `.meta` declare `isCustom` | item 3 |
| 6 | Evict aliased resources when their `.meta` is unloaded | item 4 |
| 7 | Index the path a `.meta` implies | supersedes 6, fixes hash-identified aliases |
| 8 | Add tests for `.meta` alias resolution | |
| 9 | Resolve `.meta` aliases in ResourceManager | review feedback |
| 10 | Pass the file hash into AddFile | review feedback |
| 11 | Stop constraining a `.meta` target to the requesting archive | |

Commits 1–3 must land together — 1 alone regresses the target-absent cases.

## Answering the review

Both comments asked for these APIs to be expressed in terms of `ResourceIdentifier`. Tracing the
parameters showed they were symptoms of the logic sitting at the wrong layer, so commit 9 moves it
rather than re-typing a signature.

`ResourceLoader` documents itself as taking a File and its `ResourceInitData` and handing them to a
factory. It was also deciding *which* file to load, reaching back through `mResourceManager` to find
a sidecar, rank it, and load whichever won. And the decision was split: `ResourceManager` decided a
`.meta` might exist, the loader decided whether it wins, and the seam between them travelled as a
null `fileToLoad` — one parameter meaning "here is the pre-loaded file", "here is a hint that may be
replaced", and "nothing is here, go look for a `.meta`" depending on the caller.

`ResolveMetaAlias` now lives on `ResourceManager`, next to the file loading it depends on.
`LoadResourceProcess` resolves the alias, loads whatever won, and hands the loader a file and init
data that agree. `HasMetaAlias` is gone — it existed only to stop the early bail from firing when a
path had nothing but a sidecar, and the bail now simply sits after resolution.

`GetFilePriority` takes a `ResourceIdentifier` (commit 9). It deliberately ignores
`identifier.GetParent()`: scoping the ranking to one archive contradicts the reach-back rule, and
nothing in the tree ever loads with a parent set, so a parent-aware branch would encode a semantic
decision nobody has made. Happy to revisit if that's what was meant.

## Two things found while doing this

**A `.meta` implies a path, which nothing was indexing** (commit 7). `ArchiveManager` keeps two
tables: `mHashes` maps a hash to a path, `mFileToArchive` maps a hash to the archive serving it.
Commit 1 stopped indexing sidecars under their base name, which fixed #984 — a meta-only archive was
claiming a path it had no file for — but it also dropped the `mHashes` entry, which was harmless and
useful. Without it nothing mapped `CRC64(foo)` back to `foo`, so a request identified by hash had no
path to append `.meta` to and could never resolve an alias, and `ListFiles` never mentioned `foo`, so
a directory unload could not reach a resource cached under it.

Commit 7 records `foo` in `mHashes` for every `foo.meta` and deliberately not in `mFileToArchive`.
That closes the hash case, and it retires commit 6: the suffix strip existed because nothing was
listed at the base path for the unload walk to reach, and now something is.

This matters for the motivating use case — an archive shipping only `gLinkHumanSkel.meta` with no
real `gLinkHumanSkel` anywhere is #1165's cross-game scenario, and ports resolve display lists by
CRC64.

**Parent was being copied onto the alias target** (commit 11). An identifier's parent means "load
this from that archive" — `LoadFileProcess` hands the path straight to it instead of going through
the virtual file system. Copying it onto the target claimed the target lives in the same archive as
the request, which the `.meta` says nothing about; a target is resolved across every mounted archive,
the same way `GetFilePriority` ranks it. Owner still carries over, so a resource reached through an
alias stays tagged for whoever asked for it.

## Decisions worth review

**Hash-identified resources can resolve a `.meta`.** #1208 leaves this one open. A `.meta` is keyed
by the path it sits beside, and `main`'s `ResourceIdentifier` may hold a hash rather than a path, so
resolving one requires turning the hash back into a path first. Supporting it is the reason commit 7
exists.

They are a live path, not a hypothetical — `ResourceManager::LoadResource(uint64_t crc)` braces a
hash identifier straight through, and ports reach it via `ResourceGetDataByCrc`. Excluding them would
also be internally inconsistent: `ResourceLoader` already resolves hash→path, so aliasing would work
for a hash request when the archive happened to ship a real file too, and silently not when it
shipped only the `.meta`.

**Decided against supporting alias chaining.** A `.meta` carries `ResourceInitData` — `type`,
`format`, `version`, `isCustom` — not just a target path. Allowing `X.meta → Y.meta → Z` would mean
deciding *which* of those `.meta` files the init data is read from, and there is no obviously correct
answer. Getting it wrong mis-parses the resource silently rather than failing.

So a `.meta` target has to be a real file. Where `Y` exists only as a `Y.meta`, resolution gives up
and whatever sits at `X` loads instead. This is not a regression — `port-maintenance` behaves
identically.

**Item 2 (`SetBufferOffset`) is explicitly transitional** and retires with the header under #1160,
as #1208 notes.

## Verification

**Unit tests** — `tests/archive_resource_meta_resolution_tests.cpp`, 17 cases over in-memory
archives mounted in a known order. Covers the ten scenarios from the manual kit plus the things it
cannot reach: `isCustom` in both directions, a `.meta` with no `"path"` key supplying init data for a
headerless resource, requests by CRC64, an alias target that is indexed but cannot be served, and
eviction on directory unload including under `alt/`.

They were written against the pre-review code and pass unchanged through commits 9–11, which is what
makes the refactor checkable rather than merely plausible. Two mutations confirm they are not
vacuous: inverting the priority comparison fails 9 of them including the control, and restoring
`IndexFile`'s suffix strip fails 9 including the no-shadowing case.

**Manual kit**, with full instructions and the resolution model:
**https://github.com/briaguya0/Ghostship/tree/meta-alias-testing/meta-alias-testing**

Ten cases, each run in its own process with a restart between, confirmed visually on the title
screen. The vehicle is a baserom texture aliased to a synthetic replacement, so vanilla reads as
legible copyright text rather than an ambiguous colour.

| # | case | before | after | |
|---|---|---|---|---|
| 1 | real + meta + target | green | **magenta** | flip |
| 2 | real + meta, no target | green | green | held |
| 3 | meta + target, no real | corrupted | **magenta** | flip |
| 4 | meta only | corrupted | **vanilla** | flip |
| 5 | real, no meta | green | green | held |
| 6 | target carrying an OTR header | corrupted | **magenta** | flip |
| 7 | flagship: meta < real < target | green | **magenta** | flip |
| 8 | flagship, target dropped | green | green | held |
| 9 | L1: real ranked above target | green | green | control |
| 10 | reach-back: target below the meta | corrupted | **magenta** | flip |

Case 7 → magenta with the target present and case 8 → green without it, from the same unmodified
`.meta`, is #1165. Cases 3, 4 and 10 are #984 — case 4 logged 91 failed loads in a five-second
baseline run and now logs zero. Case 9 is the control against an inverted comparison: same `.meta`,
same target present, but with the real asset ranked above it, so it must stay green.

The kit was re-run after commits 7–11 with identical results.

**Not covered by either.** Commit 11 has no test behind it: every unit test and every kit case has
`parent == nullptr`, so the line it removes was a no-op in all of them. It is justified by reading
`LoadFileProcess`, not by observation. The same blind spot applies to `GetFilePriority` ignoring
parent.

## Unrelated findings

Both turned up while capturing the baseline, neither is caused by this change, and neither is fixed
here:

1. Failed resource loads retry every frame despite the `NotFound` cache entry, because the early
   return in `LoadResourceProcess` only fires for alt-asset paths.
2. Every archive is opened and indexed twice per launch, from a single `LoadResourceFiles()` call
   site.

### DIFF
--- a/include/ship/resource/ResourceLoader.h
+++ b/include/ship/resource/ResourceLoader.h
@@ -93,6 +93,15 @@ class ResourceLoader {
      */
     uint32_t GetResourceType(const std::string& type);
 
+    /**
+     * @brief Reads and parses the resource header from the meta-file to produce ResourceInitData.
+     * @param filePath       Virtual path (for error messages).
+     * @param metaFileToLoad File containing the header (may be a separate ".meta" sidecar).
+     * @return Populated ResourceInitData, or nullptr if the header is invalid.
+     */
+    std::shared_ptr<ResourceInitData> ReadResourceInitData(const std::string& filePath,
+                                                           std::shared_ptr<File> metaFileToLoad);
+
   protected:
     /** @brief Registers the built-in factories (Blob, JSON, Shader). Called during construction. */
     void RegisterGlobalResourceFactories();
@@ -108,26 +117,6 @@ class ResourceLoader {
      * @return Matching factory, or nullptr if none is registered.
      */
     std::shared_ptr<ResourceFactory> GetFactory(uint32_t format, std::string typeName, uint32_t version);
-
-    /**
-     * @brief Reads and parses the resource header from the meta-file to produce ResourceInitData.
-     * @param filePath       Virtual path (for error messages).
-     * @param metaFileToLoad File containing the header (may be a separate ".meta" sidecar).
-     * @return Populated ResourceInitData, or nullptr if the header is invalid.
-     */
-    std::shared_ptr<ResourceInitData> ReadResourceInitData(const std::string& filePath,
-                                                           std::shared_ptr<File> metaFileToLoad);
-
-    /**
-     * @brief Resolves a `.meta` at filePath, deciding whether it or filePath itself should load.
-     * @param identifier The requesting identifier; its owner and parent carry over to the result.
-     * @param filePath   Virtual path being loaded.
-     * @param fileToLoad In/out: the file at filePath, null when only a `.meta` is there. Set to
-     *                   the `.meta`'s target when the target is the one chosen.
-     * @return Init data parsed from the `.meta`, or nullptr to keep loading filePath as-is.
-     */
-    std::shared_ptr<ResourceInitData> ResolveMetaAlias(const ResourceIdentifier& identifier,
-                                                       const std::string& filePath, std::shared_ptr<File>& fileToLoad);
 
     /** @brief Creates a ResourceInitData with default/zeroed fields. */
     static std::shared_ptr<ResourceInitData> CreateDefaultResourceInitData();
--- a/include/ship/resource/ResourceLoader.h
+++ b/include/ship/resource/ResourceLoader.h
@@ -118,6 +118,17 @@ class ResourceLoader {
     std::shared_ptr<ResourceInitData> ReadResourceInitData(const std::string& filePath,
                                                            std::shared_ptr<File> metaFileToLoad);
 
+    /**
+     * @brief Resolves a `.meta` at filePath, deciding whether it or filePath itself should load.
+     * @param identifier The requesting identifier; its owner and parent carry over to the result.
+     * @param filePath   Virtual path being loaded.
+     * @param fileToLoad In/out: the file at filePath, null when only a `.meta` is there. Set to
+     *                   the `.meta`'s target when the target is the one chosen.
+     * @return Init data parsed from the `.meta`, or nullptr to keep loading filePath as-is.
+     */
+    std::shared_ptr<ResourceInitData> ResolveMetaAlias(const ResourceIdentifier& identifier,
+                                                       const std::string& filePath, std::shared_ptr<File>& fileToLoad);
+
     /** @brief Creates a ResourceInitData with default/zeroed fields. */
     static std::shared_ptr<ResourceInitData> CreateDefaultResourceInitData();
 
--- a/include/ship/resource/archive/ArchiveManager.h
+++ b/include/ship/resource/archive/ArchiveManager.h
@@ -230,6 +230,13 @@ class ArchiveManager : public Component {
     /** @brief Rebuilds the hash-to-path and path-to-archive lookup tables from the current archive list. */
     void ResetVirtualFileSystem();
 
+    /**
+     * @brief Records a file in the hash-to-path and path-to-archive lookup tables.
+     * @param filePath Virtual path of the file.
+     * @param archive  Archive serving the file.
+     */
+    void AddFile(const std::string& filePath, const std::shared_ptr<Archive>& archive);
+
   private:
     std::vector<std::shared_ptr<Archive>> mArchives;
     std::vector<uint32_t> mGameVersions;
--- a/include/ship/resource/archive/ArchiveManager.h
+++ b/include/ship/resource/archive/ArchiveManager.h
@@ -233,10 +233,11 @@ class ArchiveManager : public Component {
 
     /**
      * @brief Records a file in the hash-to-path and path-to-archive lookup tables.
+     * @param hash     CRC64 of the virtual path.
      * @param filePath Virtual path of the file.
      * @param archive  Archive serving the file.
      */
-    void AddFile(const std::string& filePath, const std::shared_ptr<Archive>& archive);
+    void AddFile(uint64_t hash, const std::string& filePath, const std::shared_ptr<Archive>& archive);
 
   private:
     std::vector<std::shared_ptr<Archive>> mArchives;
--- a/include/ship/resource/archive/ArchiveManager.h
+++ b/include/ship/resource/archive/ArchiveManager.h
@@ -237,7 +237,7 @@ class ArchiveManager : public Component {
      * @param filePath Virtual path of the file.
      * @param archive  Archive serving the file.
      */
-    void AddFile(uint64_t hash, const std::string& filePath, const std::shared_ptr<Archive>& archive);
+    void AddFileToVfs(uint64_t hash, const std::string& filePath, const std::shared_ptr<Archive>& archive);
 
   private:
     std::vector<std::shared_ptr<Archive>> mArchives;
--- a/include/ship/resource/archive/ArchiveManager.h
+++ b/include/ship/resource/archive/ArchiveManager.h
@@ -9,6 +9,7 @@
 #include <stdint.h>
 #include <functional>
 #include "ship/resource/File.h"
+#include "ship/resource/ResourceIdentifier.h"
 #include "ship/core/Component.h"
 #ifdef ENABLE_SCRIPTING
 #include "ship/security/Keystore.h"
@@ -151,10 +152,10 @@ class ArchiveManager : public Component {
     /**
      * @brief Load-order priority of the archive that owns the given file (higher = higher
      * priority; the last-loaded archive wins). Used to rank a real asset against an alias target.
-     * @param filePath Virtual path of the file.
+     * @param identifier Identifier of the file.
      * @return Priority index, or -1 if no loaded archive has the file.
      */
-    int32_t GetFilePriority(const std::string& filePath);
+    int32_t GetFilePriority(const ResourceIdentifier& identifier);
 
     /**
      * @brief Lists virtual paths of all files matching the given search mask across all archives.
--- a/include/ship/resource/archive/ArchiveManager.h
+++ b/include/ship/resource/archive/ArchiveManager.h
@@ -149,6 +149,14 @@ class ArchiveManager : public Component {
     std::shared_ptr<Archive> GetArchiveFromFile(const std::string& filePath);
 
     /**
+     * @brief Load-order priority of the archive that owns the given file (higher = higher
+     * priority; the last-loaded archive wins). Used to rank a real asset against an alias target.
+     * @param filePath Virtual path of the file.
+     * @return Priority index, or -1 if no loaded archive has the file.
+     */
+    int32_t GetFilePriority(const std::string& filePath);
+
+    /**
      * @brief Lists virtual paths of all files matching the given search mask across all archives.
      * @param searchMask Glob pattern (empty = list everything).
      * @return Sorted, deduplicated list of matching paths.
--- a/include/ship/resource/ResourceManager.h
+++ b/include/ship/resource/ResourceManager.h
@@ -418,12 +418,14 @@ class ResourceManager : public Component {
     std::shared_ptr<IResource> GetCachedResource(std::variant<ResourceLoadError, std::shared_ptr<IResource>> cacheLine);
 
     /**
-     * @brief Whether a `.meta` sits at this identifier's path, naming a target to load instead.
+     * @brief Resolves the `.meta` beside this identifier, if its target should load instead.
      *
-     * A `.meta` is keyed by the path it sits at, so a hash-identified resource has one only when
-     * the hash resolves back to a path.
+     * The real asset at the identifier and the `.meta`'s target are both ranked by the archive
+     * holding them, and the higher one loads. A tie goes to the `.meta`.
+     *
+     * @return Init data whose Identifier names the file to load, or nullptr if no `.meta` wins.
      */
-    bool HasMetaAlias(const ResourceIdentifier& identifier);
+    std::shared_ptr<ResourceInitData> ResolveMetaAlias(const ResourceIdentifier& identifier);
 
   private:
     std::unordered_map<ResourceIdentifier, std::variant<ResourceLoadError, std::shared_ptr<IResource>>,
--- a/include/ship/resource/ResourceManager.h
+++ b/include/ship/resource/ResourceManager.h
@@ -417,6 +417,14 @@ class ResourceManager : public Component {
 
     std::shared_ptr<IResource> GetCachedResource(std::variant<ResourceLoadError, std::shared_ptr<IResource>> cacheLine);
 
+    /**
+     * @brief Whether a `.meta` sits at this identifier's path, naming a target to load instead.
+     *
+     * A `.meta` is keyed by the path it sits at, so a hash-identified resource has one only when
+     * the hash resolves back to a path.
+     */
+    bool HasMetaAlias(const ResourceIdentifier& identifier);
+
   private:
     std::unordered_map<ResourceIdentifier, std::variant<ResourceLoadError, std::shared_ptr<IResource>>,
                        ResourceIdentifierHash>
--- a/include/ship/resource/archive/Archive.h
+++ b/include/ship/resource/archive/Archive.h
@@ -159,6 +159,15 @@ class Archive : public std::enable_shared_from_this<Archive> {
     bool IsInitialized();
 
     /**
+     * @brief Load-order priority assigned by the ArchiveManager when the archive is mounted.
+     * Higher means loaded later. -1 if unset.
+     */
+    int32_t GetPriority();
+
+    /** @brief Sets the load-order priority. Called by the ArchiveManager. */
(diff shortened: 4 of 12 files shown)


include/ship/resource/ResourceLoader.h
```diff
@@ -118,6 +118,17 @@ class ResourceLoader {
...
+    /**
+     * @brief Resolves a `.meta` at filePath, deciding whether it or filePath itself should load.
+     * @param identifier The requesting identifier; its owner and parent carry over to the result.
+     * @param filePath   Virtual path being loaded.
+     * @param fileToLoad In/out: the file at filePath, null when only a `.meta` is there. Set to
+     *                   the `.meta`'s target when the target is the one chosen.
+     * @return Init data parsed from the `.meta`, or nullptr to keep loading filePath as-is.
+     */
+    std::shared_ptr<ResourceInitData> ResolveMetaAlias(const ResourceIdentifier& identifier,
+                                                       const std::string& filePath, std::shared_ptr<File>& fileToLoad);
```
Kenix3: What's the point of a separate path being passed in when that should be passed via resource identifier now? I'm also not sure if I agree that this function should load a file into the shared ptr reference. Shouldn't this just be resolving the actual identifier of the file to load?

include/ship/resource/archive/ArchiveManager.h
```diff
@@ -148,6 +148,14 @@ class ArchiveManager : public Component {
      */
     std::shared_ptr<Archive> GetArchiveFromFile(const std::string& filePath);
 
+    /**
+     * @brief Load-order priority of the archive that owns the given file (higher = higher
+     * priority; the last-loaded archive wins). Used to rank a real asset against an alias target.
+     * @param filePath Virtual path of the file.
+     * @return Priority index, or -1 if no loaded archive has the file.
+     */
+    int32_t GetFilePriority(const std::string& filePath);
```
Kenix3: This should probably receive a ResourceIdentifier

include/ship/resource/archive/ArchiveManager.h
```diff
@@ -222,6 +231,14 @@ class ArchiveManager : public Component {
     /** @brief Rebuilds the hash-to-path and path-to-archive lookup tables from the current archive list. */
     void ResetVirtualFileSystem();
 
+    /**
+     * @brief Records a file in the hash-to-path and path-to-archive lookup tables.
+     * @param hash     CRC64 of the virtual path.
+     * @param filePath Virtual path of the file.
+     * @param archive  Archive serving the file.
+     */
+    void AddFile(uint64_t hash, const std::string& filePath, const std::shared_ptr<Archive>& archive);
```
Kenix3: We need a better name for this. Maybe AddFileToVfs?

include/ship/resource/ResourceLoader.h
```diff
@@ -93,6 +93,15 @@ class ResourceLoader {
      */
     uint32_t GetResourceType(const std::string& type);
 
+    /**
+     * @brief Reads and parses the resource header from the meta-file to produce ResourceInitData.
+     * @param filePath       Virtual path (for error messages).
+     * @param metaFileToLoad File containing the header (may be a separate ".meta" sidecar).
+     * @return Populated ResourceInitData, or nullptr if the header is invalid.
+     */
+    std::shared_ptr<ResourceInitData> ReadResourceInitData(const std::string& filePath,
```
Kenix3, line 102: Doesn't file path already live on the File? Probably best for the implementation to read from that and simplify the signature.

It appears this is the existing signature, so I'm good with punting.

briaguya0: yeah, i just needed to move it from protected to public. i didn't want to change the method itself.

include/ship/resource/archive/ArchiveManager.h
briaguya0: i was following the pattern of `AddArchive` (which isn't named `AddArchiveToVfs`), but i can change it

briaguya0: addressed in https://github.com/Kenix3/libultraship/pull/1218/changes/99513aaf6cb9985babaa7cae9009fcd9c61de972

Kenix3 (approved):

PR #1220, opened 2026-08-03
[main 2.0.0] Add Palette4bpp Support for ImGui

### LINKED ISSUE #1201: Port #1157 to main: Palette4bpp ImGui texture loading (uses GetRawInstance, plus 3 fixes needed)
Sub-issue of #1152.

[#1157](https://github.com/Kenix3/libultraship/pull/1157) ("Add Palette4bpp Support for Texture Loading into ImGui", [`2bfbde3`](https://github.com/Kenix3/libultraship/commit/2bfbde3a72c119f8073ad762ec6be131dff5df66), by @Caladius) adds CI4 + TLUT decoding to `Fast3dGui::LoadGuiTexture` so 4bpp indexed textures can be shown in ImGui. The feature is wanted on `main`, but the commit cannot be cherry-picked.

`git cherry-pick 2bfbde3a` produces 3 conflicts in `src/fast/Fast3dGui.cpp`, and more importantly the result would not compile — see blocker 1.

### 1. `Context::GetRawInstance()` does not exist on `main` — hard blocker

The new palette load reads:

```cpp
auto paletteRes = std::static_pointer_cast<Fast::Texture>(
    Ship::Context::GetRawInstance()->GetResourceManager()->LoadResource(palettePath));
```

`GetRawInstance()` is the `port-maintenance` API introduced by #1103, which we are deliberately **not** porting (see #1196). `main` injects the dependency instead — `mResourceManager`, set in `Fast3dGui::Init` (`src/fast/Fast3dGui.cpp:59`) and used at `:582` and `:734`.

**Fix:** `mResourceManager->LoadResource(palettePath, true)`.

Note the `true`. The image itself is loaded with `LoadResource(path, true)` (`loadExact`), but the palette is loaded without it, so alt-asset resolution applies to the palette and not the indices. With an HD texture pack enabled the palette can resolve to a replacement resource while the index data stays vanilla. The two should match.

### 2. The RAW-path guard change is an out-of-bounds read — drop this hunk

The commit relaxes the early rejection inside the `TEX_FLAG_LOAD_AS_RAW` branch:

```diff
-        if (res.Type == Fast::TextureType::Palette4bpp || res.Type == Fast::TextureType::Palette8bpp) {
+        if (res.Type == Fast::TextureType::Palette8bpp) {
```

A `Palette4bpp` texture carrying `TEX_FLAG_LOAD_AS_RAW` then reaches:

```cpp
texBuffer.assign(res.ImageData, res.ImageData + (res.Width * res.Height * 4));
```

A 4bpp image only holds `W*H/2` bytes, so this reads **8x** what is allocated — for a 32x32 CI4, 4096 bytes out of a 512-byte buffer.

The reference implementation does the opposite. `Interpreter::ImportTextureRaw` (`src/fast/interpreter.cpp:1101-1107`) short-circuits before any RGBA32 handling:

```cpp
switch (type) {
    case Fast::TextureType::Palette4bpp:
        ImportTextureCi4(tile, importReplacement);
        return;
    case Fast::TextureType::Palette8bpp:
        ImportTextureCi8(tile, importReplacement);
        return;
```

In fast3d proper, RAW + CI4 has never meant "raw RGBA32" — it means "4bpp indexed, decode with the TLUT".

This hunk is also **not needed for the feature**: the new decode case lives in the `else` branch, which the RAW path never reaches. It is actively counterproductive — with the guard removed, a RAW-flagged CI4 skips the new decoder and takes the over-reading `assign` instead.

**Fix:** keep the original `Palette4bpp || Palette8bpp` rejection. If RAW + CI4 should work, route it into the new decoder the way `ImportTextureRaw` does, rather than into the RGBA32 assign.

### 3. Palette read is unbounded

Indices are 4-bit, so `paletteRes->ImageData[idx * 2 + 0 .. +1]` reads up to byte 31 unconditionally. Nothing checks `Fast::Texture::ImageDataSize` (`include/fast/resource/type/Texture.h:39`), which is available.

Partial TLUTs are real, not hypothetical — `Interpreter::GfxDpLoadTlut` (`src/fast/interpreter.cpp:2399-2405`) computes `entryCount = high_index + 1` and carries clamping logic precisely because of them. An 8-byte TLUT resource yields a 24-byte heap over-read past the end of the resource buffer.

`ImportTextureCi4` cannot hit this: it reads from `mRdp->palettes[]`, which point into the fixed 256-byte `palette_staging[]` buffers (`src/fast/interpreter.cpp:925-929`). A short TLUT there gives stale colors, never an OOB. The GUI path reads the resource buffer directly and has no such backstop.

**Fix:** clamp against `paletteRes->ImageDataSize` — skip or transparent-fill out-of-range indices, and warn.

### 4. `std::string` constructed from `nullptr` — UB

`src/libultraship/window/gui/GfxDebuggerWindow.cpp` in this commit:

```cpp
gui->LoadGuiTexture(name, *metadata.resource, nullptr, ImVec4{ 1.0f, 1.0f, 1.0f, 1.0f });
```

`nullptr` binds to `const std::string& palettePath`. Upstream already caught this and fixed it to `""` in [`aca5863`](https://github.com/Kenix3/libultraship/commit/aca5863c63f5bcc7d74452da6d7f6e03f5c1aaa6) (#1172, "fix a gfx debugger crash").

That one-line fix was **deliberately excluded** from #1198, because #1157 is not on `main` and `LoadGuiTexture` there still takes three arguments (`GfxDebuggerWindow.cpp:657`). It must land as part of this port.

## Hardening, not blocking

- `std::static_pointer_cast<Fast::Texture>` is an unchecked downcast, and `paletteRes->Type` is never validated to be `RGBA16bpp`. Pointing `palettePath` at a non-Texture resource, or at an RGBA32/IA16 image, silently decodes garbage — the `!paletteRes->ImageData` check will not catch it.
- The decoder hardcodes palette entries 0-15. `ImportTextureCi4` selects a bank: `palette = mRdp->palettes[palIdx / 8] + (palIdx % 8) * 16 * 2` (`src/fast/interpreter.cpp:929`). Fine for a dedicated 16-entry TLUT resource, but a shared 256-entry TLUT with a non-zero intended bank renders wrong colors with no diagnostic. Consider an optional bank/offset parameter, or warn when `ImageDataSize > 32`.
- The loop assumes rows are packed at `W/2` bytes with no padding. `ImportTextureCi4` uses an explicit stride (`src/fast/interpreter.cpp:970-971`). Any 4bpp asset with padded rows will shear. Not a live bug for real N64 textures, but a documented limitation.

## Verified correct — no change needed

Reviewed against the sibling `RGBA16bpp` case and against `ImportTextureCi4`:

- **RGBA5551 unpack** is byte-identical to the existing `RGBA16bpp` case and matches `ImportTextureCi4` channel by channel, including the `* 0xFF / 0x1F` scaling (`SCALE_5_8`, `src/fast/interpreter.cpp:51`). No endianness or channel-order discrepancy.
- **Nibble order** — high nibble first — matches `ImportTextureCi4` (`src/fast/interpreter.cpp:970`).
- **Odd `Width * Height`** pushes one extra pixel, but `UploadTexture` consumes exactly `W*H*4` and ignores it. Same shape as the existing `Grayscale4bpp` and `GrayscaleAlpha4bpp` cases.
- **Error paths** return before `api->NewTexture()` and before the `mGuiTextures[name]` write, so no renderer texture leaks and no partial cache entry is inserted.

## Note on the signature change

The new overloads insert `palettePath` *before* `tint`:

```cpp
void LoadGuiTexture(const std::string& name, const std::string& path,
                    const std::string& palettePath = "",
                    const ImVec4& tint = ImVec4(1.0f, 1.0f, 1.0f, 1.0f));
```

This is source-breaking for downstream callers using the 3-arg `LoadGuiTexture(name, path, tint)` form. The break is a loud compile error (`ImVec4` does not convert to `std::string`) rather than a silent misbinding, so it is safe — but it needs coordinating with SoH and any other consumer.

## Scope

Between the `mResourceManager` swap, dropping the RAW hunk, adding the bounds check, and folding in the #1172 one-liner, the result is enough of a rewrite that it should land as its own commit rather than carrying a `(cherry picked from ...)` trailer. Original authorship credit to @Caladius in the message body.

MegaMech:
Ported from https://github.com/Kenix3/libultraship/pull/1157
Thanks Caladius!

Tested on SpaghettiKart!

Resolves https://github.com/Kenix3/libultraship/issues/1201
According to the issue the palette creates a bigger buffer than needed.

### DIFF
--- a/include/fast/Fast3dGui.h
+++ b/include/fast/Fast3dGui.h
@@ -98,7 +98,7 @@ class Fast3dGui : public Ship::Gui {
      * @param path Virtual resource path of the source image.
      * @param tint RGBA tint multiplied over the image (use ImVec4(1,1,1,1) for no tint).
      */
-    void LoadGuiTexture(const std::string& name, const std::string& path, const std::string& palettePath = "",
+    void LoadGuiTexture(const std::string& name, const Fast::Texture& tex, const std::string& palettePath = "",
                         const ImVec4& tint = ImVec4(1.0f, 1.0f, 1.0f, 1.0f));
 
     /**
--- a/include/fast/Fast3dGui.h
+++ b/include/fast/Fast3dGui.h
@@ -98,7 +98,8 @@ class Fast3dGui : public Ship::Gui {
      * @param path Virtual resource path of the source image.
      * @param tint RGBA tint multiplied over the image (use ImVec4(1,1,1,1) for no tint).
      */
-    void LoadGuiTexture(const std::string& name, const std::string& path, const ImVec4& tint);
+    void LoadGuiTexture(const std::string& name, const std::string& path, const std::string& palettePath = "",
+                        const ImVec4& tint(1.0f, 1.0f, 1.0f, 1.0f));
 
     /**
      * @brief Returns true if a texture with the given name is already cached.
--- a/include/fast/Fast3dGui.h
+++ b/include/fast/Fast3dGui.h
@@ -99,7 +99,7 @@ class Fast3dGui : public Ship::Gui {
      * @param tint RGBA tint multiplied over the image (use ImVec4(1,1,1,1) for no tint).
      */
     void LoadGuiTexture(const std::string& name, const std::string& path, const std::string& palettePath = "",
-                        const ImVec4& tint(1.0f, 1.0f, 1.0f, 1.0f));
+                        const ImVec4& tint = ImVec4(1.0f, 1.0f, 1.0f, 1.0f));
 
     /**
      * @brief Returns true if a texture with the given name is already cached.
--- a/include/fast/Fast3dGui.h
+++ b/include/fast/Fast3dGui.h
@@ -89,6 +89,15 @@ class Fast3dGui : public Ship::Gui {
      * @param path Virtual resource path of the source image.
      * @param tint RGBA tint multiplied over the image (use ImVec4(1,1,1,1) for no tint).
      */
+    void LoadGuiTexture(const std::string& name, const std::string& path, const std::string& palettePath = "",
+                        const ImVec4& tint = ImVec4(1.0f, 1.0f, 1.0f, 1.0f));
+
+    /**
+     * @brief Loads an image from an archive path and caches it under the given name.
+     * @param name Path/texture name used to reference the texture in GetTextureByName().
+     * @param path Virtual resource path of the source image.
+     * @param tint RGBA tint multiplied over the image (use ImVec4(1,1,1,1) for no tint).
+     */
     void LoadGuiTexture(const std::string& name, const std::string& path, const ImVec4& tint);
 
     /**
--- a/src/fast/Fast3dGui.cpp
+++ b/src/fast/Fast3dGui.cpp
@@ -773,10 +773,11 @@ void Fast3dGui::LoadGuiTexture(const std::string& name, const Fast::Texture& tex
     mGuiTextures[name] = asset;
 }
 
-void Fast3dGui::LoadGuiTexture(const std::string& name, const std::string& path, const ImVec4& tint) {
+void Fast3dGui::LoadGuiTexture(const std::string& name, const std::string& path,
+                               const std::string& palettePath, const ImVec4& tint) {
     const auto res = static_cast<Fast::Texture*>(mResourceManager->LoadResource(path, true).get());
 
-    LoadGuiTexture(name, *res, tint);
+    LoadGuiTexture(name, *res, palettePath, tint);
 }
 
 void Fast3dGui::UnloadTexture(const std::string& name) {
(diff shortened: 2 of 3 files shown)


include/fast/Fast3dGui.h
```diff
@@ -89,7 +89,17 @@ class Fast3dGui : public Ship::Gui {
      * @param path Virtual resource path of the source image.
      * @param tint RGBA tint multiplied over the image (use ImVec4(1,1,1,1) for no tint).
      */
-    void LoadGuiTexture(const std::string& name, const std::string& path, const ImVec4& tint);
+    void LoadGuiTexture(const std::string& name, const std::string& path, const std::string& palettePath = "",
```
Kenix3, line 92: Need to explain the new parameters in the doxygen.

PR #1221, opened 2026-08-03
[main 2.0.0] Add ImGui FileBrowserWindow as a fallback file Dialogue

MegaMech:
https://github.com/Kenix3/libultraship/pull/1139
Thanks JeodC

### HOW TO IMPLEMENT
You must add your own `gui->AddGuiWindow()` command. It should create same place as before. But you must hold onto your own pointer now in your Engine.cpp

Todo:
- [x] Port
- [x] Compile
- [ ] Moving the OS to ControlDeck Crashes SpaghettiKart
image

Resolves https://github.com/Kenix3/libultraship/pull/1139

### DIFF
--- a/include/ship/window/gui/FileBrowserWindow.h
+++ b/include/ship/window/gui/FileBrowserWindow.h
@@ -1,0 +1,138 @@
+#pragma once
+
+#include <filesystem>
+#include <functional>
+#include <optional>
+#include <string>
+#include <vector>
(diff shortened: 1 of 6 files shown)


Malkierian:
Have to imagine the crash is because of a port-side change needed to accommodate that move.

MegaMech:
Agreed

PR #1222, opened 2026-08-03
[main 2.0.0] Add opt-in Memoization of Texture Path Resolution

### LINKED ISSUE #1206: Port #1175 to main: opt-in memoization of texture path resolution
Sub-issue of #1152.

[#1175](https://github.com/Kenix3/libultraship/pull/1175) ("Add opt-in memoization of texture path resolution", [`2917d0f`](https://github.com/Kenix3/libultraship/commit/2917d0f4fe62c579174561dcd34f327c9410bb72), by @JeodC) memoizes OTR texture-path resolution on `G_SETTIMG`, keyed by display-list pointer.

Profiling Lighthouse in heavy scenes showed path resolution as the largest avoidable cost in the display list walk — ~10 ms/frame in `LoadResourceProcess` covering string construction, hash, cache mutex and map lookup, all re-resolving the same handful of textures the scene binds every frame.

This is a small port, not a clean cherry-pick. Everything below is mechanical with a known target — no design decisions needed beyond the question at the end.

## What blocks the cherry-pick

`git cherry-pick 2917d0f4` produces exactly **one** conflict, in `src/fast/interpreter.cpp`:

```
<<<<<<< HEAD
    std::static_pointer_cast<Fast::Texture>(sResourceManager->LoadResourceProcess(imgData));
=======
    std::static_pointer_cast<Fast::Texture>(gfx->ResolveResourceCached(imgData));
>>>>>>> 2917d0f4
```

Both sides want the same call; `main` just reaches the resource manager differently. **Take the incoming side verbatim** — routing through `gfx->ResolveResourceCached` is the point of the change.

## The other change needed

The new `Interpreter::ResolveResourceCached` calls the `port-maintenance`-only API twice:

```cpp
Ship::Context::GetRawInstance()->GetResourceManager()->LoadResourceProcess(path)
```

`GetRawInstance()` came from #1103, which we are not porting (see #1196). `main` has two replacements:

- `Interpreter::mResourceManager` — member, `include/fast/interpreter.h:536` ("Cached ResourceManager, set in Init()")
- `sResourceManager` — file-scope static, `src/fast/interpreter.cpp:126`, used by the free-function command handlers

`ResolveResourceCached` is an `Interpreter` member, so **`mResourceManager` is correct**, matching existing member-function usage at `src/fast/interpreter.cpp:3012` and `:3049`.

## Everything else applies clean

- `mResolvedResourceCache` / `mResolvedResourceCacheEnabled` members (`include/fast/interpreter.h`)
- `SetResolvedResourceCacheEnabled(bool)` opt-in setter
- the `mResolvedResourceCache.clear()` hook added to `TextureCacheClear()` — `main`'s version of that function (`src/fast/interpreter.cpp:445-455`) is otherwise identical to `port-maintenance`'s

## Notes for whoever picks this up

**Off by default.** `mResolvedResourceCacheEnabled = false`, and the disabled path is a straight pass-through, so this is a genuine no-op for existing ports until one opts in.

**The precondition is real.** The memo is keyed on the *display-list path pointer*, not the string contents. The header comment says it is "Safe for ports whose display lists carry stable path pointers" — a port must actually satisfy that before enabling it, or it will hand back the wrong resource for a reused address. Worth keeping that comment verbatim in the port.

**Only hits are memoized**, deliberately. The original comment explains why:

> The resource manager caches its own misses, so re-asking for one is cheap, and `CacheExternalResource` can turn a path that missed into a valid resource at runtime. A memoized null would outlive the resource itself.

Note that `CacheExternalResource` is not on `main` yet — it arrives with #1118, currently in #1205's sibling PR #1198. The reasoning still holds without it (the "resource manager caches its own misses" half), but the comment references an API that may not exist at the time this lands. Either sequence it after #1198 or adjust the comment.

## Open question

Worth a maintainer's view before this lands: is pointer-keyed memoization the shape we want on `main`, given the stable-path-pointer precondition is invisible at the call site and silently wrong if violated? The original PR notes that if multiple ports adopt it successfully, the opt-in could be removed entirely — which would make that precondition load-bearing for everyone.

MegaMech:
https://github.com/Kenix3/libultraship/pull/1175
Thanks JeodC

Todo:
- [x] Port
- [x] Compile

Resolves https://github.com/Kenix3/libultraship/issues/1206

### DIFF
--- a/src/fast/interpreter.cpp
+++ b/src/fast/interpreter.cpp
@@ -458,13 +458,13 @@ std::shared_ptr<Ship::IResource> Interpreter::ResolveResourceCached(const char* 
         return nullptr;
     }
     if (!mResolvedResourceCacheEnabled) {
-        return Ship::Context::GetRawInstance()->GetResourceManager()->LoadResourceProcess(path);
+        return mResourceManager->LoadResourceProcess(path);
     }
     auto it = mResolvedResourceCache.find(path);
     if (it != mResolvedResourceCache.end()) {
         return it->second;
     }
-    auto res = Ship::Context::GetRawInstance()->GetResourceManager()->LoadResourceProcess(path);
+    auto res = mResourceManager->LoadResourceProcess(path);
     // Only memoize a hit. The resource manager caches its own misses, so re-asking
     // for one is cheap, and CacheExternalResource can turn a path that missed into a
     // valid resource at runtime. A memoized null would outlive the resource itself.
--- a/src/fast/interpreter.cpp
+++ b/src/fast/interpreter.cpp
@@ -442,12 +442,45 @@ ColorCombiner* Interpreter::LookupOrCreateColorCombiner(const ColorCombinerKey& 
     return &mPrevCombiner->second;
 }
 
+void Interpreter::SetResolvedResourceCacheEnabled(bool enabled) {
+    mResolvedResourceCacheEnabled = enabled;
+    if (!enabled) {
+        mResolvedResourceCache.clear();
+    }
+}
+
+/**
+ * Texture binds resolve the same paths every frame; skip the resource
+ * manager's string/hash/mutex work by memoizing on the pointer.
+ */
+std::shared_ptr<Ship::IResource> Interpreter::ResolveResourceCached(const char* path) {
+    if (path == nullptr) {
+        return nullptr;
+    }
+    if (!mResolvedResourceCacheEnabled) {
+        return Ship::Context::GetRawInstance()->GetResourceManager()->LoadResourceProcess(path);
+    }
+    auto it = mResolvedResourceCache.find(path);
+    if (it != mResolvedResourceCache.end()) {
+        return it->second;
+    }
+    auto res = Ship::Context::GetRawInstance()->GetResourceManager()->LoadResourceProcess(path);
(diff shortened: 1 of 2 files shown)


src/fast/interpreter.cpp
```diff
@@ -3862,7 +3895,7 @@ bool gfx_set_timg_handler_rdp(F3DGfx** cmd0) {
     if ((i & 1) != 1) {
         if (gfx_check_image_signature(imgData) == 1) {
             std::shared_ptr<Fast::Texture> tex =
-                std::static_pointer_cast<Fast::Texture>(sResourceManager->LoadResourceProcess(imgData));
+                std::static_pointer_cast<Fast::Texture>(gfx->ResolveResourceCached(imgData));
```
MegaMech, line 3898: Is it really opt-in if we completely override the old function call?

Kenix3: Yeah it does not appear to be opt-in. There is no logic within `ResolveResourceCached` that will just call into the ResourceManager if the memoization is not opted in.

Malkierian: It's opt-in because the new function returns the old call early if memoization is not enabled.

PR #1224, opened 2026-08-03
[main 2.0.0] Untangle Various Parts from the Context Destructor

MegaMech:
From https://github.com/Kenix3/libultraship/pull/1103

This PR seems mostly moot now. However, I've tried to moving anything that seemed useful over.

Unless this issue still persists...
https://github.com/Kenix3/libultraship/issues/1102
I don't think any of this PR is needed?

I don't know where the below gets created anymore.
image  

Perhaps here? `Engine.cpp`

image

It seems to be here I think: `Logger.cpp`
image  
 

Please advise.

### DIFF
--- a/include/fast/Fast3dGui.h
+++ b/include/fast/Fast3dGui.h
@@ -1,6 +1,7 @@
 #pragma once
 #include <SDL2/SDL.h>
 #include "ship/window/gui/Gui.h"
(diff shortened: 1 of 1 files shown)


MegaMech:
It's been deemed this PR is no longer needed. Closing.

Thanks!

PR #1226, opened 2026-08-03
port-maintenance: SDL3 audio backend

bassdr:
This is #1125 targetting port-maintenance

This brings SDL3 as an optional dependency to allow using SDL3 as an audio backend. This is working much better on Linux than the only SDL2 audio backend available.

As discussed in https://github.com/Kenix3/libultraship/pull/1105 and https://github.com/Kenix3/libultraship/pull/1107, currently the SDL audio backend has quite a lot of audio artifacts. On Windows and macOS, the dedicated WASAPI and CoreAudio backends make it less important, as these backends are of good quality (fewer audio artifacts), but Linux only has the SDL2 backend available.
Since fixing SDL2 backend or merging my PipeWire backend (Linux modern audio api) does not seem to have a lot of interest, it was discussed an SDL3 backend would be a better fit for the project.

This PR does not need to migrate the whole project to SDL3, and is designed to work on all platforms (only tested on Linux so far), while we are waiting for a full migration.

To achieve SDL2 + SDL3 loading side-by-side, SDL3 is loaded dynamically (rather than linked) on purpose: SDL2 and SDL3 export overlapping SDL_* symbol names, so linking both into one binary makes the shared names (e.g. SDL_InitSubSystem) bind to SDL2 — breaking SDL3 init. Loading SDL3 with local visibility (RTLD_LOCAL) keeps its symbols isolated, and also makes it a soft runtime dependency.

### DIFF
--- a/include/fast/backends/gfx_direct3d_common.h
+++ b/include/fast/backends/gfx_direct3d_common.h
@@ -3,6 +3,8 @@
 #if defined(ENABLE_DX11) || defined(ENABLE_DX12)
 
 #ifdef __cplusplus
+#include <windows.h>
+#include <wrl/client.h>
 #include "../interpreter.h"
 #include <cstdint>
 #include <string>
--- a/include/fast/backends/gfx_dxgi.h
+++ b/include/fast/backends/gfx_dxgi.h
@@ -1,6 +1,9 @@
 #pragma once
 #if defined(ENABLE_DX11) || defined(ENABLE_DX12)
 
+#include <windows.h>
+#include <wrl/client.h>
+
 #include "gfx_rendering_api.h"
 #include "ship/utils/HResultException.h"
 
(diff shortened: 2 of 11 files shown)


serprex:
I think we've decided to leave port-maintenance on SDL2

bassdr:
This is not a port to sdl3, this is an alternate audio backend that loads sdl3 dynamically so we can use the sdl3 audio layer alongside sdl2 for everything else.

bassdr:
Closing this in favour of #1251.

The reason I went to SDL3 was that its audio layer behaved better than SDL2's on Linux. Having since taken that backend apart, I think the part that actually helped was the pull model and the ring buffer, not SDL3 itself - and both of those can be had on SDL2, on this branch, in two files with no new dependency, no runtime loader and no second backend to select. That is #1251.

If measurement later shows SDL3's audio layer wins on its own merits beyond that, this can come back with evidence attached, which is a much better argument than the one I opened it with. For now it is a large PR whose premise I can no longer defend.

Thanks @briaguya0 for the digging in the thread here - the conclusion that the ports all bound their own buffering is what made removing the fill guard safe, and that carries straight over.

PR #1227, opened 2026-08-03
Make ConsoleVariable thread-safe (port-maintenance)

bassdr:
## Problem

`ConsoleVariable` stores its CVars in a `std::unordered_map` (`mVariables`) with no synchronisation, but the map is accessed from more than one thread. A consumer that reads CVars from the audio thread while the game thread mutates the store will eventually race.

The failure reproduces reliably when the game thread applies a large batch of CVar writes at once (e.g. loading a settings preset) while the audio thread is reading a CVar per note:

```
Signal: 11  INVALID ACCESS TO STORAGE
  Ship::ConsoleVariable::GetInteger(char const*, int)
  CVarGetInteger
  Audio_InitNoteSub
  Audio_ProcessNotes
  ...
  OTRAudio_Thread()
terminate called after throwing an instance of 'std::bad_alloc'
```

The batch of inserts rehashes the map and frees the old bucket array while the audio thread is mid-`find()`, so the lookup walks freed memory. The resulting heap corruption then surfaces as `std::bad_alloc` on the next allocation.

## Fix

Guard every public accessor of `mVariables` with a `std::recursive_mutex`. It is recursive because the public methods call one another (`Register*` -> `Get` + `Set*`, `Load` -> `LoadFromPath` -> `Set*`), so a single non-recursive lock at each entry point would self-deadlock. The two protected helpers (`LoadFromPath`, `LoadLegacy`) are only reached from `Load` and stay under its lock.

Reads on the audio thread take an uncontended lock (a couple of atomics); a large write burst briefly serialises with them instead of crashing.

### DIFF
--- a/include/libultraship/bridge/consolevariablebridge.h
+++ b/include/libultraship/bridge/consolevariablebridge.h
@@ -12,7 +12,7 @@
  * @brief Returns a shared pointer to the CVar with the given name, or nullptr if it does not exist.
  * @param name CVar name (e.g. "gMyOption").
  */
-std::shared_ptr<Ship::CVar> CVarGet(const char* name);
+Ship::CVar* CVarGet(const char* name);
 
 extern "C" {
 #endif
--- a/include/ship/config/ConsoleVariable.h
+++ b/include/ship/config/ConsoleVariable.h
@@ -4,6 +4,7 @@
 #include <nlohmann/json.hpp>
 #include <stdint.h>
 #include <memory>
+#include <mutex>
 #include <unordered_map>
 #include <string>
 #include <string_view>
@@ -209,5 +210,7 @@ class ConsoleVariable {
         }
     };
     std::unordered_map<std::string, std::shared_ptr<CVar>, TransparentStringHash, TransparentStringEqual> mVariables;
+    // Guards mVariables against concurrent access from the audio thread and the game thread.
+    std::recursive_mutex mVariablesMutex;
 };
 } // namespace Ship
--- a/include/ship/config/ConsoleVariable.h
+++ b/include/ship/config/ConsoleVariable.h
@@ -4,14 +4,24 @@
 #include <nlohmann/json.hpp>
 #include <stdint.h>
 #include <memory>
+#include <mutex>
 #include <unordered_map>
+#include <unordered_set>
 #include <string>
 #include <string_view>
+#include <atomic>
+#include <deque>
+#include <vector>
 
 namespace Ship {
 
-/** @brief Discriminator tag for the active field of the CVar union. */
-typedef enum class ConsoleVariableType { Integer, Float, String, Color, Color24 } ConsoleVariableType;
+/**
+ * @brief Discriminator tag for the active field of the CVar union.
+ *
+ * None means the entry carries no value: it is either freshly created and not yet
+ * filled in, or it has been cleared. Lookups treat it as if the name did not exist.
+ */
+typedef enum class ConsoleVariableType { None, Integer, Float, String, Color, Color24 } ConsoleVariableType;
 
 /**
  * @brief A single console variable (CVar) holding a typed value.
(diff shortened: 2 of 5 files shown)


garrettjoecox:
Seeing as some of the ports are calling these thousands of times per second, this feels like it has pretty severe performance implications. This needs thorough playtesting in a port to make sure we're not regressing

bassdr:
> Seeing as some of the ports are calling these thousands of times per second, this feels like it has pretty severe performance implications. This needs thorough playtesting in a port to make sure we're not regressing

You are right, and I tested quickly and it does add overhead on SoH. I did it this way because avoiding the lock requires a redesign where the ownership of the CVar becomes clear. I am drafting it right now.

bassdr:
@garrettjoecox playtested in Shipwright on three variants and rewrote the implementation. Short version: you were right that the lock costs something on the read path, but **I could not measure an in-game regression for any variant, including the mutex one.**

The original mutex version is preserved at [`backup/cvar-thread-safety-mutex-port-maintenance`](https://github.com/bassdr/libultraship/tree/backup/cvar-thread-safety-mutex-port-maintenance) if you would rather take that - a 24-line diff against the same base, and it does fix the crash.

### What it does now

Readers take no lock. Entries live in a deque that only grows, so a `CVar*` handed out stays valid; the name index is replaced rather than mutated, and a writer introducing a new name publishes a fresh map atomically. Superseded maps are retained, because a lock-free reader may still be inside one. Writers serialise on a mutex readers never touch. Values are written before the type that describes them behind a release fence, and a new `None` type marks unfilled or cleared entries - which also makes clearing an O(1) retype rather than an index rebuild.

It also closes pre-existing damage in the same file: `SetString` freed a buffer a concurrent reader could still hold, `free()` was applied to entries whose union held no string, self-assignment was a use-after-free, and `LoadLegacy` leaked a `strdup`. String values are now interned in a set that is never pruned, so a `const char*` from `GetString` stays valid for the lifetime of the store.

### Microbenchmark

Real `ConsoleVariable.cpp` from three revisions, median of 3 runs, same flags as the game binary (clang 22.1.8, `-march=znver3 -flto=thin -O3`), Ryzen 5 5600G. Nanoseconds per read, lower is better:

| | 1 reader | 2 | 4 | 8 |
|---|---|---|---|---|
| no synchronisation | 23.8 | 28.1 | 15.8 | 9.2 |
| mutex | 36.4 | 184 | 154 | 204 |
| this branch | **20.0** | **10.1** | **5.1** | **3.9** |

The mutex costs about +53% per read uncontended; this version comes out ~16% faster than having no synchronisation at all, because dropping the `shared_ptr` removes an atomic refcount pair per lookup.

### Playtest

Three `soh.elf` binaries differing only in this file, run back to back from one sandbox with the same config, save and scene set, frame limit 360, vsync off. **All three sat at roughly 260 FPS in the Market; I could not tell them apart.** The lock is effectively uncontended in practice, so what is left is ~16 ns against a 3.85 ms frame. The multi-reader columns are a stress shape, not a prediction, and the 8-thread row oversubscribes a 6-core machine - the single-reader column is the honest one.

So the case for this version over the backup branch is correctness and structure, not framerate: the `GetString` use-after-free is gone by construction rather than by being careful. If you prefer the smaller diff, the backup branch is there and I will not argue for this one on performance grounds.

One trade-off worth stating: interned strings are never freed. That is bounded by the distinct values a session assigns, but a few CVars (`ExcludedLocations`, `EnabledTricks`, the enabled-mods list) intern a fresh multi-kilobyte joined list on every UI toggle. Recycling them needs hazard pointers or an epoch scheme, which felt like more machinery than this warrants - happy to revisit if you disagree.

Benchmark source available if you want to reproduce it.

bassdr:
@garrettjoecox @Kenix3 main took the plain mutex in #1253, so I opened #1269 to bring the same 24 lines here. That fixes the crash.

Rescoping this PR to the lock-free readers as an optimisation on top of #1269.

garrettjoecox:
> @garrettjoecox @Kenix3 main took the plain mutex in #1253, so I opened #1269 to bring the same 24 lines here. That fixes the crash.
> 
> Rescoping this PR to the lock-free readers as an optimisation on top of #1269.

that doesn't seem great, seems like maybe we should aim to revert #1253?

bassdr:
> > @garrettjoecox @Kenix3 main took the plain mutex in #1253, so I opened #1269 to bring the same 24 lines here. That fixes the crash.
> > Rescoping this PR to the lock-free readers as an optimisation on top of #1269.
> 
> that doesn't seem great, seems like maybe we should aim to revert #1253?

As long as we get the fix IMO, both PR fixes a crash, this one with an optim as a bonus. IMO I would do this in this order:
1. Merge #1269
2. Merge #1227 after a rebase
3. If there are no regression on SoH, cherry-pick #1227 in `main`

The good news: `main` does not get the crash anymore
The bad news: `main` is slower (probably imperceptible) and `port-maintenance` (soh) still has the crash

PR #1228, opened 2026-08-04
fix(fast3d): accept tagged Android ARM64 resource pointers

pacoa-kdbg:
## Summary
- Mask the ARM64 top byte before Fast3D validates a candidate `__OTR__` resource-path pointer.
- Android allocators can tag heap pointers; rejecting them made Fast3D upload resource-path bytes as texture data.

## Validation
- AI-generated implementation.
- Built, installed, and device-tested on an AYANEO Pocket S Mini (Adreno 740, Android 13).
- User subsequently performed manual testing on the same device.

This is intentionally a one-line Fast3D pointer-validation change.

### DIFF
--- a/src/fast/interpreter.cpp
+++ b/src/fast/interpreter.cpp
@@ -5348,7 +5348,9 @@ int32_t gfx_check_image_signature(const char* imgData) {
 #if UINTPTR_MAX > 0xFFFFFFFFu
     // On 64-bit: filter kernel/sentinel addresses. Upper bound covers all
     // user-space layouts (x86_64 47-bit canonical, ARM64 48-bit VA, etc.).
-    if (i > 0x0000FFFFFFFFFFFFull) {
+    // ARM64 Android heap pointers can carry a top-byte memory tag; the CPU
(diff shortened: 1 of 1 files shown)


JeodC:
This looks like a simpler version of #1214. Is this deprecated in favor of that, then, or are they collaborative?

pacoa-kdbg:
@JeodC yes, it is in functionality, just what I needed for the 'port-maintenance' branch that lighthouse uses to get it working on android.


This should be closed if considered appropriate to just use the other PR and wait for that to get into this branch

JeodC:
I wouldn't know what's appropriate and what's not here. @izzy2lost builds for multiple different Android devices I believe, so I'll ask them to weigh in. We are also in the midst of migrating things from `port-maintenance` over to `main`, so pull requests to this branch may be frozen for all I know.

If that happens to be the case, we will have to wait for Lighthouse to migrate to LUS 2.0.

izzy2lost:
Yes this is the same change I need. Sorry I messed up my PR. Don't do mine do this one. Mine isn't pointing to the right branch. I'll delete mine. Please accept this one when you get a chance.

PR #1229, opened 2026-08-06
Restore component-owned state

mckinlee:
## Summary
- preserve game-specific button names passed to ControlDeck
- release ResourceManager-owned loader and archive components when the manager leaves the component tree
- safely skip ImGui backend shutdown when no context was initialized
- cover the component-state regressions with focused tests

## Testing
Windows Debug libultraship_tests build
- ctest --test-dir build-validation -C Debug --output-on-failure --timeout 120 (621/621)
- Township SDL3 focused regression tests (11/11)

### DIFF
--- a/include/ship/resource/ResourceManager.h
+++ b/include/ship/resource/ResourceManager.h
@@ -407,6 +407,7 @@ class ResourceManager : public Component {
      * - "validHashes"  (array of uint32): acceptable game-version hashes; empty = all accepted.
      */
     void OnInit(const nlohmann::json& initArgs = nlohmann::json::object()) override;
+    void OnRemoved(bool forced) override;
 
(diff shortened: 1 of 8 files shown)


mckinlee:
FYI, the ControlDeck portion of this PR is not strictly necessary for Township yet. I’m still exploring whether input should ultimately be owned by LUS or Rainfall. It allows Township to provide GC style button names and have LUS retain and expose them instead of discarding them. A follow up would still be needed for the Input Editor to use those names and display a GC layout.

mckinlee:
Closing this in favor of consolidating Township's libultraship changes on one current-main support branch. These component-lifecycle fixes are preserved and can be included in a cumulative upstream proposal when the Township integration is ready for maintainer review.

PR #1234, opened 2026-08-06
use spdlog::fmt_lib over fmt, opens up SPDLOG_USE_STD_FORMAT in future

serprex:
requires spdlog 1.10

### DIFF
--- a/cmake/dependencies/android.cmake
+++ b/cmake/dependencies/android.cmake
@@ -40,7 +40,7 @@ if (NOT ${tinyxml2_FOUND})
 endif()
 
 #=================== spdlog ===================
-find_package(spdlog QUIET)
+find_package(spdlog 1.10 QUIET)
 if (NOT ${spdlog_FOUND})
     FetchContent_Declare(
         spdlog
--- a/cmake/dependencies/ios.cmake
+++ b/cmake/dependencies/ios.cmake
@@ -40,7 +40,7 @@ if (NOT ${tinyxml2_FOUND})
 endif()
 
 #=================== spdlog ===================
-find_package(spdlog QUIET)
+find_package(spdlog 1.10 QUIET)
 if (NOT ${spdlog_FOUND})
     FetchContent_Declare(
         spdlog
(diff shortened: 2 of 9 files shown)


cmake/dependencies/android.cmake
```diff
@@ -40,7 +40,7 @@ if (NOT ${tinyxml2_FOUND})
 endif()
 
 #=================== spdlog ===================
-find_package(spdlog QUIET)
+find_package(spdlog 1.10 QUIET)
```
briaguya0, line 43: i'm a little hesitant to start using the version arg here, looking at the cmake docs for it

https://cmake.org/cmake/help/latest/command/find_package.html

> The <version> argument requests a version with which the package found should be compatible. There are two possible forms in which it may be specified:
>
>    A single version with the format major[.minor[.patch[.tweak]]], where each component is a numeric value.
>
>    A version range with the format versionMin...[<]versionMax where versionMin and versionMax have the same format and constraints on components being integers as the single version. By default, both end points are included. By specifying <, the upper end point will be excluded. Version ranges are only supported with CMake 3.19 or later.
>
> >Note
> >
> > With the exception of CPS packages, version support is currently provided only on a package-by-package basis. When a version range is specified but the package is only designed to expect a single version, the package will ignore the upper end point of the range and only take the single version at the lower end of the range into account. Non-CPS packages that do support version ranges do so in a manner that is determined by the individual package. See the [Version Selection](https://cmake.org/cmake/help/latest/command/find_package.html#version-selection) section below for details and important caveats.

the parts that stand out to me are

> version support is currently provided only on a package-by-package basis

and

> with which the package found should be compatible

and from the "version selection" section

> When the <version> argument is given, Config mode will only find a version of the package that claims compatibility with the requested version

so i'd want to dig into what "claims compatibility" means here as (at least from what i've read of the docs) it doesn't seem cmake does `npm`-style semver checking.

briaguya0 (commented):
overall in favor of this, one concern about the implementation

cmake/dependencies/android.cmake
serprex: I can dig into details here when I'm back next week

serprex: 1.10 compatible with all 1.x >=1.10, will correctly reject spdlog v2: https://github.com/gabime/spdlog/blob/f5f173a1a57d0e2e0115f2ed71ee7ea316516853/CMakeLists.txt#L421C1-L421C94

PR #1235, opened 2026-08-07
fast3d: fix swapped F3DEX lookat axes breaking env mapping

JeodC:
Before (LUS on the right):
image

After:
Screenshot 2026-08-07 163932

`G_MV_LOOKATX` and `G_MV_LOOKATY` were stored into mRsp->lookat in reverse. Mistakenly introduced in #598 which added the `F3DEX1` case, copying the idiom from the `F3DEX_G_MV_L0` case immediately below it: `(index - BASE) / 2`. That works for lights because `G_MV_L0..L7` ascend. It does not work for lookat: gbi.h declares `G_MV_LOOKATY 0x82` before `G_MV_LOOKATX 0x84`. #727 added the `SpReset` defaults, written as `lookat[0] = {0,127,0} / lookat[1] = {127,0,0}` to agree with the already-inverted storage. This made the inversion internally self-consistent, which is why it survived.

### DIFF
--- a/src/fast/interpreter.cpp
+++ b/src/fast/interpreter.cpp
@@ -4944,11 +4944,14 @@ void Interpreter::SpReset() {
     mRsp->modelview_matrix_stack_size = 1;
     mRsp->current_num_lights = 2;
     mRsp->lights_changed = true;
-    mRsp->lookat[0].dir[0] = 127;
-    mRsp->lookat[0].dir[1] = 0;
+    // Transposed from the SDK basis on purpose: lookat[0] is LOOKATX and drives S.
+    // Only games that never send a lookat see these, and MK64 matches output with them.
+    // TODO: Use ucode DMEM defaults from hardware.
+    mRsp->lookat[0].dir[0] = 0;
+    mRsp->lookat[0].dir[1] = 127;
     mRsp->lookat[0].dir[2] = 0;
-    mRsp->lookat[1].dir[0] = 0;
-    mRsp->lookat[1].dir[1] = 127;
+    mRsp->lookat[1].dir[0] = 127;
+    mRsp->lookat[1].dir[1] = 0;
     mRsp->lookat[1].dir[2] = 0;
     CalculateNormalDir(&mRsp->lookat[0], mRsp->current_lookat_coeffs[0]);
     CalculateNormalDir(&mRsp->lookat[1], mRsp->current_lookat_coeffs[1]);
--- a/src/fast/interpreter.cpp
+++ b/src/fast/interpreter.cpp
@@ -2370,7 +2370,7 @@ void Interpreter::GfxSpMovememF3d(uint8_t index, uint8_t offset, const void* dat
             break;
         case F3DEX_G_MV_LOOKATY:
         case F3DEX_G_MV_LOOKATX:
-            memcpy(mRsp->lookat + (index - F3DEX_G_MV_LOOKATY) / 2, data, sizeof(F3DLight_t));
+            memcpy(mRsp->lookat + (F3DEX_G_MV_LOOKATX - index) / 2, data, sizeof(F3DLight_t));
             break;
         case F3DEX_G_MV_L0:
         case F3DEX_G_MV_L1:
@@ -4942,11 +4942,11 @@ void Interpreter::SpReset() {
     mRsp->modelview_matrix_stack_size = 1;
     mRsp->current_num_lights = 2;
     mRsp->lights_changed = true;
-    mRsp->lookat[0].dir[0] = 0;
-    mRsp->lookat[0].dir[1] = 127;
+    mRsp->lookat[0].dir[0] = 127;
+    mRsp->lookat[0].dir[1] = 0;
     mRsp->lookat[0].dir[2] = 0;
-    mRsp->lookat[1].dir[0] = 127;
-    mRsp->lookat[1].dir[1] = 0;
+    mRsp->lookat[1].dir[0] = 0;
+    mRsp->lookat[1].dir[1] = 127;
     mRsp->lookat[1].dir[2] = 0;
     CalculateNormalDir(&mRsp->lookat[0], mRsp->current_lookat_coeffs[0]);
     CalculateNormalDir(&mRsp->lookat[1], mRsp->current_lookat_coeffs[1]);
--- a/src/fast/interpreter.cpp
+++ b/src/fast/interpreter.cpp
@@ -2368,9 +2368,11 @@ void Interpreter::GfxSpMovememF3d(uint8_t index, uint8_t offset, const void* dat
         case F3DEX_G_MV_VIEWPORT:
             CalcAndSetViewport((const F3DVp_t*)data);
             break;
-        case F3DEX_G_MV_LOOKATY:
         case F3DEX_G_MV_LOOKATX:
-            memcpy(mRsp->lookat + (F3DEX_G_MV_LOOKATX - index) / 2, data, sizeof(F3DLight_t));
+            memcpy(mRsp->lookat + 0, data, sizeof(F3DLight_t));
+            break;
+        case F3DEX_G_MV_LOOKATY:
+            memcpy(mRsp->lookat + 1, data, sizeof(F3DLight_t));
             break;
         case F3DEX_G_MV_L0:
         case F3DEX_G_MV_L1:


Malkierian (approved):

MegaMech:
Can you test on SpaghettiKart?

JeodC:
> Can you test on SpaghettiKart?

Any spots to test in particular?

Malkierian:
Are F3DEX_G_MV_LOOKATX and F3DEX_G_MV_LOOKATY supposed to be handled in the same case? I can't shake the feeling this should instead be separating the cases rather than just replacing what's there.

JeodC:
> Are F3DEX_G_MV_LOOKATX and F3DEX_G_MV_LOOKATY supposed to be handled in the same case? I can't shake the feeling this should instead be separating the cases rather than just replacing what's there.

Now that you mention it, I suppose it could. Would be less confusing for future reference as well.

src/fast/interpreter.cpp
```diff
@@ -2368,9 +2368,11 @@ void Interpreter::GfxSpMovememF3d(uint8_t index, uint8_t offset, const void* dat
         case F3DEX_G_MV_VIEWPORT:
             CalcAndSetViewport((const F3DVp_t*)data);
             break;
-        case F3DEX_G_MV_LOOKATY:
         case F3DEX_G_MV_LOOKATX:
-            memcpy(mRsp->lookat + (index - F3DEX_G_MV_LOOKATY) / 2, data, sizeof(F3DLight_t));
+            memcpy(mRsp->lookat + 0, data, sizeof(F3DLight_t));
```
briaguya0, line 2372: curious about the `/ 2` getting dropped here. having 2 separate cases is definitely easier to follow than using the enum value in the calculation, just want to make sure it's still lining up properly

MegaMech:
> Any spots to test in particular?

The intro logo (the spinning logo), just drive around a track.

I don't think the game uses the macro. But I'd really prefer if it was at least given a quick test. Given that it changes lookAt farther down which I'm not sure what exactly that part does there.

src/fast/interpreter.cpp
JeodC: The `/ 2` was the bug, not incidental. `F3DEX_G_MV_LOOKATY = 0x82`, `F3DEX_G_MV_LOOKATX = 0x84`, so `(index - LOOKATY) / 2` sent Y to `lookat[0]` and X to `lookat[1]`, which is backwards, since `lookat[0]` feeds `dotx` and `lookat[1]` feeds `doty`. The F3DEX2 path already maps `X` t 0 and `Y` to 1 via `offset / 24`; this just makes F3DEX agree. The `/ 2` only worked because the opcodes are 2 apart.

Malkierian:
Alright, did some testing on the spinning logo. The results are that the old way is more accurate to emulator for MK. Here's emu, SK before the changes, and SK after the changes:
mkemu
skold
sknew
As you can see, the light is pointing down right in both emu and SK before, whereas in SK after, it's pointing up left. So, this may be a BK-specific difference we have to apply in another way.

Malkierian:
Also, I don't think the case handling was bugged, it was just unclear in comparison to how SpReset was being handled later. I tried changing only the case handling on SK and the lighting was still consistent with old and emu.

Malkierian:
Also, for what it's worth, at least on SK, swapping the +0 and +1 on the lookat cases changes nothing that I could see. The only thing that actually changes anything is the SpReset stuff.

JeodC:
From what I can find in SK, it never sets a `lookat` at all: no `gSPLookAt` anywhere in game code, and `src/os/guLookAtRef.c` is dead code (`guLookAtReflect` is never called, and the file isn't even in CMakeLists). The case split can't do anything there, and the only thing that moves the logo is the `SpReset` half. Every texgen surface in MK64 is running on whatever LUS defaults to.

BK does send `lookat` through `gSPLookAt`, and the slot order isn't really in question: texgen reads `lookat_coeffs[0]` as U, and the F3DEX2 path already puts `G_MVO_LOOKATX` (offset 0) in slot 0. The old F3DEX1 math was using DMEM order (Y at 0x82, X at 0x84) on an array that's X first.

What we could do here is keep the case split and drop the `SpReset` change. Those defaults only affect games that never set a lookat, so picking values for that needs its own evidence. FWIW, leaving the old values in with the fixed cases just reads them as `X = (0,127,0), Y = (127,0,0)`, so SK keeps rendering the way it does now. BK uploads its `lookat` before every env-mapped model, so the defaults never survive into a texgen draw there and the fix in the screenshots is unaffected.

JeodC:
> From what I can find in SK, it never sets a `lookat` at all: no `gSPLookAt` anywhere in game code, and `src/os/guLookAtRef.c` is dead code (`guLookAtReflect` is never called, and the file isn't even in CMakeLists). The case split can't do anything there, and the only thing that moves the logo is the `SpReset` half. Every texgen surface in MK64 is running on whatever LUS defaults to.
> 
> BK does send `lookat` through `gSPLookAt`, and the slot order isn't really in question: texgen reads `lookat_coeffs[0]` as U, and the F3DEX2 path already puts `G_MVO_LOOKATX` (offset 0) in slot 0. The old F3DEX1 math was using DMEM order (Y at 0x82, X at 0x84) on an array that's X first.
> 
> What we could do here is keep the case split and drop the `SpReset` change. Those defaults only affect games that never set a lookat, so picking values for that needs its own evidence. FWIW, leaving the old values in with the fixed cases just reads them as `X = (0,127,0), Y = (127,0,0)`, so SK keeps rendering the way it does now. BK uploads its `lookat` before every env-mapped model, so the defaults never survive into a texgen draw there and the fix in the screenshots is unaffected.

While it's on my mind, reverting the `SpReset` half keeps SK where it is, but it doesn't explain anything. LUS has no concept of "no lookat set yet" - it invents a basis at reset, and any F3DEX game that never uploads one just renders off that invention. Nobody has checked it against what the ucode actually leaves in DMEM, so the old values and mine are both guesses.

But I suspect the defaults aren't even the interesting part. If the F3DEX ucode data image has zeros in the lookat slots, which is what I'd expect for an uninitialized DMEM buffer, then a game that genuinely never uploads a lookat gets `dotx = doty = 0` on console. That's every texgen vertex sampling one texel, a flat wash, not the directional highlight in the emu screenshot above. Which makes me think MK64 *does* send a lookat on hardware and something between the rom and `GfxSpMovememF3d` is dropping it. Torch is a candidate - its movemem handler in `DisplayListFactory.cpp` is marked "TODO: Fix this opcode" and only resolves the pointer as a light asset.

Easy to check: log every `F3DEX_G_MV_LOOKATX/Y` hit in `GfxSpMovememF3d` and play the intro. No hits means the upload is being lost in extraction and the `SpReset` values are papering over a SK bug. Hits means the defaults never applied to that scene and the logo change came from something else.

MegaMech:
The logo uses G_TEXTURE_GEN

And a reflect map

JeodC:
> The logo uses G_TEXTURE_GEN
> 
> And a reflect map

That's what makes it interesting! Texgen is on, so the model consumes a lookat, but nothing in SK ever produces one. Either the retail game had a call site that didn't survive the decomp, or the `lookat` rides along in the rom display list and we're losing it in extraction.

MegaMech:
That light command tho is a sub command and it looks like lookat is also a subcommand. Are they sharing the same sub command?

Perhaps on hardware the light command does more than 1 function?

I just know move mem is a bloody nightmare

JeodC:
Now you're gonna go "but the decomp is at 100%", but look. `guLookAtReflect` isn't in the rom. `mk64.ld` links `src/os/guLookAtF.o` but `guLookAtRef.o` isn't in the linker script at all, it only shows up in `safe_gcc.mk`. The file matches as libultra source, but the game never calls it and never computes a `lookat` at runtime.

If the logo has a reflect map, then, the `lookat` has to be baked in rom data and pushed by a static `gsSPLookAtX/Y` in the DL. That works for this scene since the camera is fixed and the model is what spins. That's the case torch handles worst: the movemem path resolves the pointer through `GetStringByAddr` and wants the struct declared as an asset, but `data_segment2.yml` has nothing but gfx and vtx entries around `D_02007F60`. If that upload is being dropped in extraction, SK falls back to the LUS defaults.

> That light command tho is a sub command and it looks like lookat is also a subcommand. Are they sharing the same sub command?
> 
> Perhaps on hardware the light command does more than 1 function?
> 
> I just know move mem is a bloody nightmare

In F3DEX2 they are the same subcommand, yes. Not in F3DEX.

JeodC:
So, I've exonerated Torch: the rom data matches what Torch extracts. That leaves us with one head scratcher for later.
https://github.com/Kenix3/libultraship/pull/1235/commits/1b6561c45b224620240cc412adaadab43f967ec3 ought to fix both BK and SK, so good to go?

PR #1236, opened 2026-08-09
Custom Tex Scrolling

Jameriquiah:
adds a custom xml command for user defined tex scroll settings. i tried to keep the usability as simple as possible, so u just define directional speed for both axes and thats it.

Example command that would be present in a displaylist:
`	<ScrollTexture Tile="0" SpeedX="4" SpeedY="-3" />`

tile is the texture slot u want to scroll, can either be 0 or 1, X is up, Y is down, positive numbers scroll up/right, negative scrolls down/left


PurpleHato:
John Fast just added the feature that everyone wanted :shipit:

serprex (approved):

Jameriquiah:
conflicts resolved

PurpleHato (approved):
LGTM

PR #1248, opened 2026-08-18
Verify and retry the stb_image.h download

bassdr:
`file(DOWNLOAD)` without a `STATUS` or hash argument treats an HTTP error page and a truncated body as success. When the raw endpoint hiccups, `stb_image.h` is written empty or half-written, the configure step reports nothing, and the build dies several minutes later on every use of stb:

```
GuiTexture.cpp(9,9): error C3861: 'stbi_image_free': identifier not found
GuiTextureFactory.cpp(20,54): error C2146: syntax error: missing '>' before identifier 'stbi_uc'
```

This has now cost re-runs on unrelated PRs (most recently #1253, twice), and it is confusing every time because the reported errors point at our source rather than at the download.

The fix verifies the file against a known SHA-256 and retries before giving up, so a transient failure no longer needs a human to notice and re-run. A file that is already present and hashes correctly is skipped, which also stops the redundant re-download that happened on every single configure.

### Verified

Against a local server injecting each failure mode:

| scenario | before | after |
| --- | --- | --- |
| healthy download | works | works |
| reconfigure twice | re-downloads every time | 0 re-downloads |
| HTTP 404 | 0-byte header, `libstb.a` links, fails later in the build | fails at configure, names the URL |
| truncated body | 5000-byte header, fails later with `unterminated comment` | fails at configure, names the URL |
| 404, 404, then healthy | broken build | recovers inside the same configure |
| connection refused | 0-byte header, fails later in the build | fails at configure, names the URL |
| corrupted cached file | used as-is | re-downloaded and repaired |

Also confirmed the committed section against the real URL (downloads, hash matches, stb compiles) and that an existing build tree reconfigures without re-fetching.

### DIFF
--- a/cmake/dependencies/common.cmake
+++ b/cmake/dependencies/common.cmake
@@ -52,7 +52,40 @@ endif()
 
 #=================== STB ===================
 set(STB_DIR ${CMAKE_BINARY_DIR}/_deps/stb)
-file(DOWNLOAD "https://github.com/nothings/stb/raw/0bc88af4de5fb022db643c2d8e549a0927749354/stb_image.h" "${STB_DIR}/stb_image.h")
+set(stb_url "https://github.com/nothings/stb/raw/0bc88af4de5fb022db643c2d8e549a0927749354/stb_image.h")
+set(stb_expected_hash c54b15a689e6a1f32c75e2ec23afa442e3e0e37e894b73c1974d08679b20dd5c)
+
+# file(DOWNLOAD) does not report error pages or truncated bodies, so verify the hash and retry.
+set(stb_verified FALSE)
+if(EXISTS "${STB_DIR}/stb_image.h")
+    file(SHA256 "${STB_DIR}/stb_image.h" stb_actual_hash)
+    if(stb_actual_hash STREQUAL "${stb_expected_hash}")
+        set(stb_verified TRUE)
+    endif()
+endif()
+
+foreach(stb_attempt RANGE 1 4)
+    if(stb_verified)
+        break()
+    endif()
+    if(stb_attempt GREATER 1)
+        message(STATUS "Retrying stb_image.h download (attempt ${stb_attempt})")
+        execute_process(COMMAND ${CMAKE_COMMAND} -E sleep 3)
+    endif()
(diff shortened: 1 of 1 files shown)


briaguya0 (commented):
adding expected hash is reasonable but the 404 was obviously due to github downtime today. the title of this PR shouldn't be "now 404s" as https://github.com/nothings/stb/raw/0bc88af4de5fb022db643c2d8e549a0927749354/stb_image.h is loading without issues now

please update the title and description of this PR

edit: updated referenced URL in comment as `/raw` redirects to `raw.` so when i copied the url from my browser at first  i got the redirected one instead of the original

serprex (approved):

bassdr:
You are right, and thank you for the correction - I checked and `/raw` is a 302 to `raw.githubusercontent.com` that CMake follows without trouble, same bytes and same hash. The URL was never the problem, yesterday's failure was the outage.

So I have dropped the URL change: the diff is now only `EXPECTED_HASH`, and the title and description no longer mention it. What is left is the part that stands on its own - a fetch that returns nothing currently leaves an empty header and a green configure, and the build fails later with `unknown type name 'stbi_uc'`. Outages happen; the argument is only that they should stop the configure instead of poisoning the build directory.

@serprex your approval was on the earlier diff, so it may be worth a second look - the change is strictly smaller now.

bassdr:
Hit this again today on #1253, which touches nothing but `ConsoleVariable`:

```
GuiTexture.cpp(9,9): error C3861: 'stbi_image_free': identifier not found
GuiTextureFactory.cpp(20,54): error C2146: syntax error: missing '>' before identifier 'stbi_uc'
GuiTextureFactory.cpp(24,9): error C3861: 'stbi_failure_reason': identifier not found
```

[build-windows-arm64](https://github.com/Kenix3/libultraship/actions/runs/32747030776/job/97494907636) failed that way while `build-windows-x64` on the same commit passed, so the fetch simply failed on one runner and the bare `file(DOWNLOAD)` wrote a partial `stb_image.h` without complaining. The errors point at our source rather than at the download, which is the part that costs the time - I went and read the diff before noticing every `stbi_*` symbol was missing at once.

`EXPECTED_HASH` turns that into a configure-time failure naming the file.

PR #1250, opened 2026-08-19
audio/SDL: pull model with a ring buffer and underrun masking

bassdr:
`SDLAudioPlayer` pushes: `DoPlay()` queues onto the device straight from the game thread. Total latency is our queue plus SDL's stream plus the device buffer - three buffers, only one of which we can see, so there is nothing to tune. It also drops the buffer outright once `Buffered()` passes 6000 frames, turning a slow frame into an audible cut.

This opens the stream with an `SDL_AudioStreamCallback` instead, so SDL asks for samples when the device needs them and `DoPlay()` only writes into a ring buffer we own.

**Not a new design for this repo.** `CoreAudioAudioPlayer` already works exactly this way, and exists because the SDL path was not good enough on macOS. This brings SDL to the same architecture with that implementation's rough edges addressed:

| | `CoreAudioAudioPlayer` | here |
| --- | --- | --- |
| lock in the audio callback | blocking `pthread_mutex_lock` | `try_lock`, silence for that burst if contended |
| short ring | partial fill, zero-pad the rest | whole burst or nothing |
| underrun | silence | last burst replayed faded to zero |
| allocation on the audio thread | none | none (scratch sized in `DoInit()`) |

### Why it is worth doing

- **Latency.** Owning the buffer puts the whole budget in one place, so `DesiredBuffered` means what it says instead of being one term of three.
- **Underruns become inaudible rather than merely rarer.** You can substitute samples only if you fill the buffer; `SDL_PutAudioStreamData` gives no such hook.
- **A step toward fewer backends.** WASAPI and CoreAudio exist because the SDL path could not compete, and SDL routes to both underneath anyway. CoreAudio first, since it is the one this now matches.

### Scope

Two files, `SDLAudioPlayer` the only class touched. No new backend, no CMake changes, no interface changes. Two commits: the pull model, then the underrun fade.

Replaces #1107 and #1125, both closed. #1107 removed the same `Buffered() < 6000` drop and nothing else; #1125 added a whole second SDL3 backend, which stopped making sense once #1191 put libultraship on SDL3.

#1251 is the same change against `port-maintenance`, in SDL2 form. That branch sees far more play, so it is the better place to find out whether this helps.

### Testing

The ring buffer, callback and fade come from #1125, which I have been running on Linux (PipeWire) for months; this is that logic moved onto `SDLAudioPlayer` and trimmed of what the SDL3 migration made redundant. In this exact form it has not been through a play session, and never on Windows or macOS - the SDL path is selectable but not the default there, so testing would be welcome, particularly latency against the native backends.

### DIFF
--- a/include/ship/audio/SDLAudioPlayer.h
+++ b/include/ship/audio/SDLAudioPlayer.h
@@ -12,7 +12,9 @@ namespace Ship {
  *
  * SDLAudioPlayer opens the platform output with `SDL_OpenAudioDeviceStream` and lets SDL pull from it:
  * SDL owns the audio thread and calls AudioStreamCallback() whenever the device needs more samples.
- * DoPlay() only writes into a ring buffer, which decouples the game thread's timing from the device's.
+ * DoPlay() only writes into a ring buffer, which decouples the game thread's timing from the device's
+ * and keeps the whole latency budget in one place. If the ring runs dry the callback replays the last
+ * burst faded out, which covers a load spike far less audibly than a hole in the stream.
  * It supports stereo and 6-channel output according to the configured AudioChannelsSetting.
  * This backend is available on all platforms that Ship supports.
  */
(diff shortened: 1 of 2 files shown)


bassdr:
Opened #1251, the same change against `port-maintenance` in SDL2 form.

Whichever lands first, the other is a small translation: the ring buffer, the fade and the locking are identical, only the SDL calls differ.

serprex (approved):

bassdr:
@Kenix3 approved by @serprex on Aug 19. #1251 (the port-maintenance twin) is held until this lands.

PR #1251, opened 2026-08-19
audio/SDL: pull model with a ring buffer and underrun masking (port-maintenance)

bassdr:
`SDLAudioPlayer` pushes: `DoPlay()` calls `SDL_QueueAudio` straight from the game thread, so the device is fed on the game's schedule rather than its own. Total latency is our queue plus SDL's plus the device buffer - three buffers, one of them visible, so `DesiredBuffered` does not mean what it says and there is nothing to tune. And once `Buffered()` passes 6000 frames `DoPlay()` throws the buffer away, turning a slow frame into an audible cut.

This passes `SDL_OpenAudioDevice` a callback instead - the mode SDL2's audio API is built around, `SDL_QueueAudio` being the later convenience path:

- SDL owns the audio thread and asks for samples when the device needs them.
- `DoPlay()` only writes into a ring buffer sized at twice `DesiredBuffered`, so the whole latency budget sits in one place we control.
- Writes are whole-chunk-or-nothing; a partial write leaves a gap mid-chunk, which clicks.
- On underrun the callback replays the last burst faded linearly to zero instead of handing SDL silence. Only the first burst fades - repeating it would turn a stall into a stutter.
- The callback never blocks: it `try_lock`s and plays silence for that one burst if the game thread holds the ring. SDL2 gives the destination pointer and an exact length, so nothing is assembled on the audio thread.

**Not a new design for this repo.** `CoreAudioAudioPlayer` already works exactly this way, and exists because the SDL path was not good enough on macOS. This brings SDL to the same architecture with that implementation's rough edges addressed:

| | `CoreAudioAudioPlayer` | here |
| --- | --- | --- |
| lock in the audio callback | blocking `pthread_mutex_lock` | `try_lock`, silence for that burst if contended |
| short ring | partial fill, zero-pad the rest | whole burst or nothing |
| underrun | silence | last burst replayed faded to zero |

### Why it is worth doing

- **Latency on Linux.** SDL2 is the only audio path Linux has here, and owning the buffer is what lets the target actually be a target.
- **Underruns become inaudible rather than merely rarer.** You can substitute samples only if you fill the buffer; `SDL_QueueAudio` offers no such hook.
- **Fewer backends, eventually.** The bespoke WASAPI and CoreAudio backends exist because the SDL path could not compete, and SDL routes to both underneath anyway.

### Scope and relationship to the other PRs

Two files, `SDLAudioPlayer` the only class touched. No new backend, no CMake changes, no interface changes.

- Replaces #1189 and #1226, both closed. #1189 removed the same 6000-frame drop and also doubled `want.samples`; that pulls the other way once this lands, since the ring already absorbs the jitter the doubling compensated for, so `want.samples` stays at `GetSampleLength()`.
- #1250 is the same change against `main`, in SDL3 form. Whichever merges first, the other is a small translation - the ring, the fade and the locking are identical, only the SDL calls differ.

### Testing

The ring buffer, callback and fade come from the SDL3 backend in #1125 (closed), which I have been running on Linux for months. In this exact form - SDL2, this branch - it has not been through a play session, and never on Windows or macOS, where the SDL path is selectable but not the default. Testing there would be welcome, particularly latency against the native backends.

### DIFF
--- a/src/ship/audio/SDLAudioPlayer.cpp
+++ b/src/ship/audio/SDLAudioPlayer.cpp
@@ -1,6 +1,10 @@
 #include "ship/audio/SDLAudioPlayer.h"
 #include <spdlog/spdlog.h>
 
+#include <algorithm>
+#include <bit>
+#include <cstring>
+
 namespace Ship {
 
 SDLAudioPlayer::~SDLAudioPlayer() {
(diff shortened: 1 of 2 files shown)


src/ship/audio/SDLAudioPlayer.cpp
```diff
@@ -9,54 +12,209 @@ SDLAudioPlayer::~SDLAudioPlayer() {
...
 
+static constexpr uint32_t NextPow2(uint32_t v) {
+    v--;
+    v |= v >> 1;
+    v |= v >> 2;
+    v |= v >> 4;
+    v |= v >> 8;
+    v |= v >> 16;
+    return v + 1;
+}
```
serprex: `std::bit_ceil`

serprex:
I'd be more open to merging this after #1250 merged to `main`, as that avoids adding to backlog of PRs that haven't merged to main we need to track

src/ship/audio/SDLAudioPlayer.cpp
bassdr: Right, C++20 is on and this is exactly it. Swapped for `std::bit_ceil` on both branches, the hand-rolled version is gone.

Values are identical over the whole range that matters - checked all of [1, 2^31] rather than assumed. The one behaviour difference is at 0, in `std::bit_ceil`'s favour: the old code returned 0 there, which would have left `mRingMask` at 0xFFFFFFFF over an empty buffer, where `bit_ceil` gives 1.

PR #1252, opened 2026-08-20
Keep the draw batch when a texture load does not change the binding

buddingmonkey:
This fix is a necessary step for optimization on low power or high performance demanding platforms. I found it when building a stereoscopic XR port of a decomp. It can yield a significant reduction in draw calls.

`gDPLoadTextureBlock` expands to `G_SETTIMG` + `G_SETTILE` + `G_LOADBLOCK` + `G_SETTILE` +
`G_SETTILESIZE`, and `GfxDpSetTile`, `GfxDpLoadBlock` and `GfxDpSetTileSize` each set
`mRdp->textures_changed[]` unconditionally. Nothing compares the new tile against the texture
already bound, so the next triangle flushes the batch whenever a tile command ran rather than when
the binding moved. `ImportTexture` then finds the texture in the cache and uploads nothing — the
`Flush()` in front of it is the only thing that cost anything.

N64 display lists re-issue the same load per object, so this is the common case, not a rare one:
sprites, particles, fonts, HUD. A 60-particle emitter on one texture becomes 60 draw calls of two
triangles each, where the batch holds 256.

So compare before flushing, in the shape `LookupOrCreateColorCombiner` already uses.
`BuildTextureBinding` splits the cache-key construction out of `ImportTexture` and touches no GPU
state and no LRU; `TextureBindingUnchanged` builds the key for the slot and compares it against the
node the slot already holds. Keys are unique, so an equal key is the node a lookup would have
returned and the whole block is skipped — no flush, no `SelectTexture`, no LRU move. The early-out
is deliberately narrow: only an unmasked, non-alt-asset-blended tile whose address isn't a
registered framebuffer mirror. Everything else falls through to today's code. `SelectTextureFb`
binds outside the texture cache, so its two call sites now null `mRenderingState.mTextures[0]`.

The dirty flags are left alone. Guarding them in the RDP handlers would be cheaper still, but
`textures_changed` doubles as the general "re-select this slot" trigger, so suppressing it at the
source needs reasoning about consumers that comparing at the point of use doesn't.

Measured on [Lighthouse](https://github.com/HarbourMasters/Lighthouse) via Linux (Release, GL), draws per frame: 213 → 68 in gameplay, 290 → 94 in another scene,
1.8x fewer over a whole session with menus and logos included. A particle burst where each particle
carries its own frame went 902 → 890, which is the point — the guard only removes a flush that
changes nothing, so a scene that really does rebind per object keeps every draw call it has.

Written with AI assistance; the diff and its reasoning were reviewed by hand.

### DIFF
--- a/include/fast/interpreter.h
+++ b/include/fast/interpreter.h
@@ -210,6 +210,17 @@ struct TextureCacheMapIter {
     TextureCacheMap::iterator it;
 };
 
+// Everything a tile needs to name a texture, with no GPU state touched and no cache entry made.
+struct TextureBinding {
+    TextureCacheKey key;
+    const uint8_t* origAddr; // source address, after the empty-TMEM-slot fallback
+    const uint8_t* fbAddr;   // address the framebuffer-mirror lookup reads, null when replacing
+    uint32_t tmemIndex;
+    uint32_t texFlags;
+    uint8_t fmt, siz;
+    bool fellBackToOtherTmem;
+};
+
 struct RGBA {
     uint8_t r, g, b, a;
 };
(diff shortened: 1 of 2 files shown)


buddingmonkey:
Ran tidy and cleaned up the two failing lines. Should be ready for testing again

PR #1253, opened 2026-08-24
Make ConsoleVariable thread-safe

bassdr:
## Problem

`ConsoleVariable` stores its CVars in a `std::unordered_map` (`mVariables`) with no synchronisation, but the map is accessed from more than one thread. A consumer that reads CVars from the audio thread while the game thread mutates the store will eventually race.

The failure reproduces reliably when the game thread applies a large batch of CVar writes at once (e.g. loading a settings preset) while the audio thread is reading a CVar per note:

```
Signal: 11  INVALID ACCESS TO STORAGE
  Ship::ConsoleVariable::GetInteger(char const*, int)
  CVarGetInteger
  Audio_InitNoteSub
  Audio_ProcessNotes
  ...
  OTRAudio_Thread()
terminate called after throwing an instance of 'std::bad_alloc'
```

The batch of inserts rehashes the map and frees the old bucket array while the audio thread is mid-`find()`, so the lookup walks freed memory. The resulting heap corruption then surfaces as `std::bad_alloc` on the next allocation.

## Fix

Guard every public accessor of `mVariables` with a `std::recursive_mutex`. It is recursive because the public methods call one another (`Register*` -> `Get` + `Set*`, `Load` -> `LoadFromPath` -> `Set*`), so a single non-recursive lock at each entry point would self-deadlock. The two protected helpers (`LoadFromPath`, `LoadLegacy`) are only reached from `Load` and stay under its lock.

Reads on the audio thread take an uncontended lock (a couple of atomics); a large write burst briefly serialises with them instead of crashing.

`main` counterpart of #1227. The change is the same; the only difference is that the four methods reaching `Config` do so through the Component system's `mConfig` member here, rather than `Context::GetRawInstance()->GetConfig()`.

### DIFF
--- a/include/ship/config/ConsoleVariable.h
+++ b/include/ship/config/ConsoleVariable.h
@@ -5,6 +5,7 @@
 #include <nlohmann/json.hpp>
 #include <stdint.h>
 #include <memory>
+#include <mutex>
 #include <unordered_map>
 #include <string>
 #include <string_view>
@@ -217,5 +218,7 @@ class ConsoleVariable : public Component {
     };
     std::unordered_map<std::string, std::shared_ptr<CVar>, TransparentStringHash, TransparentStringEqual> mVariables;
     std::shared_ptr<Config> mConfig;
+    // Guards mVariables against concurrent access from the audio thread and the game thread.
+    std::recursive_mutex mVariablesMutex;
 };
 } // namespace Ship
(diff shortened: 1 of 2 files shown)


bassdr:
The `build-windows-arm64` red on the previous push was not this change: the `stb_image.h` download came back truncated, so the build failed with `error C3861: 'stbi_image_free': identifier not found` while `build-windows-x64` passed on the same commit.

#1248 is the fix for that flake -- it verifies the download with `EXPECTED_HASH`, so a bad fetch fails at configure time with a clear message instead of surfacing as a confusing compile error several minutes later. Until it merges, this is going to keep costing re-runs on unrelated PRs; this is the second time it has hit this one.

Re-pushed an identical tree here to re-run CI.

Kenix3 (approved):

PR #1260, opened 2026-08-28
Platform fixes: iOS build, web paths, dxgi, background throttling

KiritoDv:
A batch of small platform fixes:

- **dxgi**: define `WIN32_LEAN_AND_MEAN` before including Windows headers.
- **Web**: return `/storage` as the app directory under Emscripten.
- **iOS**: build fixes — don't compile the ImGui GL3 backend or advertise the OpenGL window backend on iOS, gate CoreAudio off (SDL audio is used), correct `GetAppBundlePath` to the actual bundle path, pin the Metal layer drawable size to the pixel viewport, libzip cmake pins for link-less `try_compile` false-positives, `GIT_SHALLOW` on FetchContent deps, and a `DISABLE_DLL_LOADER` option for platforms where `dlopen` is not allowed.
- **Unfocused throttling**: skip rendering (but keep simulating) while the window is minimized/hidden/occluded — acquiring a Metal drawable for a non-visible window stalls ~1s per frame, which looked like the game lagging in the background. Occlusion is read from SDL3's `SDL_WINDOW_OCCLUDED`. On macOS the process also opts out of App Nap so audio keeps running at full speed while unfocused.

Originally written against SDL2; re-ported to the SDL3 backend for this PR (the SDL2-era iOS point/pixel ImGui scaling workarounds were dropped — the ImGui SDL3 backend handles display scale itself).

Tested: compiles on macOS (Metal + GL). iOS/web behavior re-verification pending on device.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -139,6 +139,10 @@ if(ENABLE_SCRIPTING)
     target_compile_definitions(libultraship PUBLIC ENABLE_SCRIPTING)
 endif()
 
+if(DISABLE_DLL_LOADER)
+    target_compile_definitions(libultraship PUBLIC DISABLE_DLL_LOADER)
+endif()
+
 #=================== Compile Options & Defs ===================
 
 target_compile_definitions(libultraship PUBLIC
--- a/src/ship/core/Context.cpp
+++ b/src/ship/core/Context.cpp
@@ -323,6 +323,10 @@ std::string Context::GetAppDirectoryPath(const std::string& appName) {
     }
 #endif
 
+#ifdef __EMSCRIPTEN__
+    return "/storage";
+#endif
+
 #ifdef __IOS__
     const char* home = getenv("HOME");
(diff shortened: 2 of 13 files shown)


src/ship/core/Context.cpp
```diff
@@ -323,6 +326,10 @@ std::string Context::GetAppDirectoryPath(const std::string& appName) {
     }
 #endif
 
+#ifdef __EMSCRIPTEN__
```
Kenix3, line 329: LUS doesn't have a web target. We should probably add one (including build verifications) before adding code around it.

src/CMakeLists.txt
```diff
@@ -139,6 +139,10 @@ if(ENABLE_SCRIPTING)
     target_compile_definitions(libultraship PUBLIC ENABLE_SCRIPTING)
 endif()
 
+if(DISABLE_DLL_LOADER)
```
Kenix3, line 142: What does this do?

PR #1261, opened 2026-08-28
ScriptLoader: compiled-script cache, TCC error capture, optional TCC compiler

KiritoDv:
ScriptLoader quality-of-life improvements:

- **Compiled-script cache**: new `SetCacheDir()` — compiled mod binaries are cached keyed by a hash of the archive checksum + code version + build options, so unchanged mods skip TCC recompilation on subsequent runs. Disabled by default (empty cache dir).
- **Better compile errors**: TCC diagnostics are captured and surfaced in the thrown exception instead of a generic "TCC Error in file".
- **TCC runtime root**: `tcc_set_lib_path` is pointed at the runtime root (the library path whose parent has an `include/` sibling) so TCC finds its own headers and `libtcc1.a` in portable installs.
- **`ENABLE_TCC_COMPILER` option** (default ON): with it OFF, scripting still loads prebuilt per-platform binaries but TinyCC is neither built nor linked (`DISABLE_TCC_COMPILER`), for platforms where runtime codegen is unavailable or unwanted (e.g. iOS).

Implemented on top of the current Component-based ScriptLoader.

Tested: macOS build with `-DENABLE_SCRIPTING=ON`.

### DIFF
--- a/include/ship/scripting/ScriptLoader.h
+++ b/include/ship/scripting/ScriptLoader.h
@@ -10,6 +10,7 @@
 #include <string>
 #include <optional>
 #include <functional>
+#include <filesystem>
 #include <memory>
 
 namespace Ship {
@@ -98,6 +99,20 @@ class ScriptLoader : public Component {
     std::vector<std::string> GetLoadersInDependencyOrder() const;
 
     /**
+     * @brief Sets the directory used to cache compiled script binaries.
+     *
+     * When set, ScriptLoader will store compiled .so/.dll files in this
+     * directory keyed by a hash of the archive checksum, code version, and
+     * build options. On subsequent runs the cached binary is loaded directly,
+     * skipping recompilation when the mod has not changed.
+     *
+     * Defaults to empty (caching disabled).
+     *
+     * @param dir Filesystem path to the cache directory (created automatically).
+     */
+    void SetCacheDir(const std::filesystem::path& dir);
+
+    /**
      * @brief Sets the security level for script loading.
      * @param level The SafeLevel policy to enforce.
      */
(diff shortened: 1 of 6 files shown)


include/ship/scripting/ScriptLoader.h
```diff
@@ -113,6 +128,19 @@ class ScriptLoader : public Component {
     std::unordered_map<std::string, std::string> mCompileDefines;
     std::unordered_map<std::string, Scripting::LibraryLoader> mLoadedScripts;
     std::vector<std::shared_ptr<Archive>> mLoadedArchives;
+    std::filesystem::path mCacheDir; ///< Empty = caching disabled.
+
+    /**
+     * @brief Returns the cache file path for an archive, or empty if caching is disabled
+     *        or the archive has no checksum.
+     */
+    std::filesystem::path GetCachePath(const ArchiveManifest& manifest) const;
```
Kenix3, line 137: We should document the parameters of this function.

include/ship/scripting/ScriptLoader.h
```diff
@@ -113,6 +128,19 @@ class ScriptLoader : public Component {
...
+
+    /**
+     * @brief Copies @p srcPath into the cache directory using the manifest's cache key.
+     *        Does nothing if caching is disabled or the checksum is empty.
+     */
+    void StoreInCache(const ArchiveManifest& manifest, const std::string& srcPath) const;
```
Kenix3, line 143: We should document the parameters of this function.

src/ship/scripting/ScriptLoader.cpp
```diff
@@ -130,6 +205,12 @@ void ScriptLoader::Compile(const std::shared_ptr<Archive>& archive) {
 
         loader.WriteToTempFile(*data);
     } else if (!info.Main.empty()) {
+#ifdef DISABLE_TCC_COMPILER
+        SPDLOG_WARN("ScriptLoader: '{}' ships C sources but the TCC compiler is disabled; provide a prebuilt "
```
Kenix3, line 209: I think we should probably disable script loading altogether when scripting is turned off. We want the ability to disable this code altogether.

PR #1262, opened 2026-08-28
Renderer: move vertex pipeline to the GPU, implement mipmapping and two-tile LOD

KiritoDv:
First PR of the new-renderer series: moves the fast3d vertex pipeline to the GPU and implements proper mipmapping.

- **GPU pipeline**: matrix transformation, lighting, texgen, normals and the color combiner move from the CPU interpreter into the vertex/fragment shaders (GL, Metal, D3D11). The interpreter now streams raw vertex data and per-draw uniforms instead of pre-transformed triangles.
- **Mipmapping / LOD**: RT64-style mipmapping, two-tile LOD support, and multiple texture-selection/tmem fixes; sprite shading fixes.
- Shader templates gain shared include files (`fast3d_vs.glsli`, `filter.glsli`, `noise.glsli`, `palette.glsli`).
- `DrawAndRunGraphicsCommands` gains a `dlReplacements` parameter for display-list replacement during interpolation.

1. #1257 `port/gbi-opcodes`
2. **#1262 `port/renderer-gpu-pipeline` ← this PR**
3. #1263 `port/postprocessing`
5. #1265 `port/vulkan-backend`

Tested: compiles and links on macOS (Metal + GL) against current main. Runtime-verified extensively on the SDL2-era branch in Ghostship (SM64); re-verification on this SDL3 rebase is in progress.


bassdr:
Two walls in one large outdoor scene collide but never draw, on both the OpenGL and Vulkan
backends. Bisected to the vertex transform moving into the shader: restoring the CPU
transform in `GfxSpVertex` - writing clip coordinates and pointing at the identity palette
entry, the way the rect path does - is the only thing that makes them appear.

Ruled out by experiment: backface culling (forced off), depth compare (LESS vs LEQUAL),
alpha and discard (forced opaque fragments, no `discard`), texture packs, LTO and
optimisation level, matrix-history capacity (64 -> 256), and palette slot selection
(capped the mapping at one entry so every draw reads slot 0 - the walls stayed missing).

Ruled out by instrumentation: the palette contents are right. Comparing per vertex, at
draw time, the clip position the CPU would have computed at vertex-load time against what
`mTransform.mtx_palette[slot]` produces gives 0 mismatches in 21M vertices.

So the triangles are submitted (per-frame counts are normal), they carry the correct
matrix CPU-side, the slot they read does not matter, and they still never rasterise. The
remaining difference between the two paths is the vertex data itself: object coordinates
plus a palette matrix, versus finished clip coordinates. I have not been able to narrow it
further without a GPU capture.

Two unrelated things noticed in the same code:

- `mtx_slots[3]` in `GfxSpTri1` is left uninitialised when the two-attempt palette loop
  cannot satisfy a triangle, and is then used to index the uniform array. Unreachable at
  palette size 8, since a triangle would need vertices from 9 matrices.
- `GFX_MTX_PALETTE_SIZE` sizes `TransformUniforms` while the shaders hard-code
  `mtx_palette[32]`; lowering the constant alone renders black.

vulkan_missing2

PR #1263, opened 2026-08-28
Renderer: post-processing system, custom shaders and shader settings

KiritoDv:
Second PR of the new-renderer series: a full post-processing system.

- **Multipass framework**: archives can contribute post-process passes via `manifest.json`; passes render fullscreen through prism-processed shader templates with custom uniforms (32 slots, engine-provided frame/fb registers).
- **Custom shaders**: material shaders can be replaced per display-list path; `@setting(...)` declarations in shader templates are discovered at compile time and exposed as CVar-backed settings (float/int sliders, toggles, enums, color pickers).
- **ShaderSettingsWindow**: auto-generated UI for pass toggles and shader settings, namespaced by shader-pack name to avoid conflicts.
- **SRGB** becomes a regular post-process pass instead of a backend flag.
- Fixes along the way: viewport with postprocess steps, Metal postprocessing, texture interpolation, Metal texture stutter (manual framebuffer clear), full `G_CD_*` noise implementation, color dithering default off.

Requires a prism-processor bump (`aa83709`) for the new setting-declaration types.

1. #1257 `port/gbi-opcodes`
2. #1262 `port/renderer-gpu-pipeline`
3. **#1263 `port/postprocessing` ← this PR**
5. #1265 `port/vulkan-backend`

Tested: compiles on macOS (Metal + GL). Runtime-verified on the SDL2-era branch; re-verification on this rebase in progress.


src/fast/shaders/opengl/include/fast3d_vs.glsli
```diff
@@ -0,0 +1,171 @@
...
+    @{attr} float aMtxSlot;
+    @{update_floats(1)}
+
+    // Matrix palette: 4 vec4 rows per slot; clip = obj.x*r0 + obj.y*r1 + obj.z*r2 + obj.w*r3
+    uniform vec4 uMtxPalette[32];
+    uniform vec4 uYScale;
+
+    // Game-bindable custom uniform registers; [0]-[1] are engine built-ins
+    // (frame/time/delta, resolution). See CustomUniforms in gfx_rendering_api.h.
+    uniform vec4 uCustom[32];
```
JeodC, line 20: On GLES this uniform is `highp` here but `mediump` in the fragment stage, because `gfx_opengl.cpp` injects `precision mediump float;` into the fragment `GLSL_VERSION` and not the vertex one. As a result, the vertex stage falls back to the GLSL ES default of `highp`.

GLSL ES 3.00 requires uniforms shared between stages to match, so every program fails to link and a port renders black (audio and logic run fine), tested with Ghostship where this PR was already applied. Mesa enforces this with no relaxation at `#version >= 300`.

Changing this to `uniform highp vec4 uCustom[32];` matches the vertex default, and highp is needed for the frame/time and resolution registers to be usable.

src/fast/shaders/opengl/default.shader.glsl
```diff
@@ -1,297 +1,311 @@
...
+
+    @if(o_blend[0]) uniform sampler2D uTexBlend0;
+    @if(o_blend[1]) uniform sampler2D uTexBlend1;
+
+    uniform int frame_count;
+    uniform float noise_scale;
+
+    // Game-bindable custom uniform registers; [0]-[1] are engine built-ins
+    // (frame/time/delta, resolution). See CustomUniforms in gfx_rendering_api.h.
+    uniform vec4 uCustom[32];
```
JeodC, line 65: Same change needed here to match: `uniform highp vec4 uCustom[32];`. See the note on fast3d_vs.glsli:12.

src/fast/backends/gfx_opengl.cpp
JeodC, line 1: In `glLinkProgram`, `GL_LINK_STATUS` is never checked, so a link failure is completely silent and a port just renders black. The compile paths above check `GL_COMPILE_STATUS` and `abort()` on failure; the link path has no equivalent. Checking the status and logging `glGetProgramInfoLog` would have surfaced the issue immediately.

JeodC (changes requested):

src/fast/shaders/opengl/include/fast3d_vs.glsli
KiritoDv: Fixed in 4e34fecc. Declared `highp` explicitly, guarded with `@if(opengles)` because the desktop vertex stage is built as `#version 110`, where precision qualifiers are not valid. Thanks for tracking the link failure down to this.

src/fast/shaders/opengl/default.shader.glsl
KiritoDv: Fixed in 4e34fecc, same conditional form as the vertex stage.

src/fast/backends/gfx_opengl.cpp
KiritoDv: Fixed in 30b855b0. `GL_LINK_STATUS` is now checked and the program info log is logged through SPDLOG_ERROR before the abort, matching the compile paths.

PR #1264, opened 2026-08-28
Renderer: async texture streaming, auto-mipmaps and HD texture support

KiritoDv:
Third PR of the new-renderer series: texture streaming and generated mipmaps.

- **Async texture loading** (`gEnhancements.Graphics.AsyncTextureLoad`): HD/replacement textures decode off-thread while the vanilla texture renders; `AcquireDrawTexture` caches decoded archive textures so they are not re-decoded every frame (the main level-load stall).
- **Deferred texture frees** and a per-frame upload budget (`gEnhancements.Graphics.TextureUploadBudget`) to spread big 4K uploads across frames.
- **Auto-mipmap generation** for HD textures (box downsample, capped at 16x16), disabled for non-HD textures; palette-exhaustion fix.
- HD texture support ported from Jeod's implementation; texture-flash fix; vertices work on both `GBI_FLOATS` and regular builds.
- Tail fixes: Emscripten (GLSL ES precision, canvas sizing/resize), GLES3 (context attributes, `GL_MIRROR_CLAMP_TO_EDGE` fallback, Android window flags), Metal crash fix and `lequal` depth compare.

**Stacked PR** — review and merge bottom-up; until the layers below merge, each PR's diff includes them:
1. #1257 `port/gbi-opcodes`
2. #1262 `port/renderer-gpu-pipeline`
3. #1263 `port/postprocessing`
4. **#1264 `port/texture-streaming` ← this PR**
5. #1265 `port/vulkan-backend`

Tested: compiles on macOS (Metal + GL). Runtime-verified on the SDL2-era branch; re-verification on this rebase in progress.

PR #1265, opened 2026-08-28
Implement a Vulkan rendering backend

KiritoDv:
Final PR of the new-renderer series: a Vulkan rendering backend.

- Mirrors the Metal backend's architecture: per-framebuffer primary command buffers recorded during the frame, submitted in framebuffer order at EndFrame with the swapchain (fb 0) last, then present.
- Deliberate simplifications for correctness and portability (MoltenVK included): images live in `VK_IMAGE_LAYOUT_GENERAL`, render passes always LOAD/STORE with `vkCmdClearAttachments` clears, lazy per-(shader, state) pipelines, per-draw constants in a host-visible ring with dynamic offsets.
- Shaders compile at runtime via shaderc from prism templates (`shaders/vulkan/`); ImGui renders through `imgui_impl_vulkan`.
- Build wiring: `FindVulkan.cmake` (with Windows SDK auto-detection and shaderc_shared/combined fallbacks), `LUS_ENABLE_VULKAN`/`ENABLE_VULKAN`, backend registration behind `Vulkan_IsSupported()`, CI deps for apt/brew.

Written for SDL3 (`SDL_Vulkan_GetInstanceExtensions`/`SDL_Vulkan_CreateSurface` current signatures, `SDL_GetWindowSizeInPixels`) and constructor-injected dependencies like the other backends.

1. #1257 `port/gbi-opcodes`
2. #1262 `port/renderer-gpu-pipeline`
3. #1263 `port/postprocessing`
5. **#1265 `port/vulkan-backend` ← this PR**

Tested: compiles and links on macOS against MoltenVK + shaderc. Runtime-verified on the SDL2-era branch (Windows/Linux); re-verification on this rebase in progress.


Ammar-Sadaoui:
can i access depth buffer on this vulkan backend ?

quarrel07:
Heads up from Ghostship 3.0.0, which ships this backend via the converge line: linking Vulkan::Vulkan on Windows makes vulkan-1.dll a load-time import of the exe, and that DLL only exists on machines with a Vulkan-capable driver. A user on a PC without one can't launch the game at all, before the DX11/OpenGL config choice is ever read. This PR has the same target_link_libraries line, so main would inherit it.

I opened a small fix on the converge branch that marks vulkan-1.dll and shaderc_shared.dll as delay-load imports on MSVC: https://github.com/KiritoDv/libultraship/pull/9. Verified on a Ghostship CI build that both move from the normal import table to the delay-load table with no other changes. Same nine lines would apply here.

bassdr:
Heads-up on a collision with #1289, which bounds `mSegmentPointers[16]` in `interpreter.cpp`.

`G_MW_SEGMENT` indexes it with `offset / 4`, and `offset` arrives as `C0(0, 16)` — so `segNumber` reaches 16383 and the write lands up to ~128 KB past the array. `SegAddr` has the matching unmasked read (`w1 >> 24`, no bound). The RSP field is 4 bits; the fix wraps with `% MAX_SEGMENT_POINTERS` at both.

This PR has a hunk in `GfxSpMovewordF3d` and adds a new `SegAddr` caller for the float path, so whichever of the two merges second will need to carry it — and until then the new backend inherits the out-of-bounds read.

bassdr:
Ran this end-to-end on Linux (RADV, SDL3). Three issues stop the backend working, and I have opened KiritoDv/libultraship#12 against `port/vulkan-backend` with fixes for each rather than inline suggestions, since one of them moves state into `FrameSlot`.

- `Vulkan_IsSupported()` is never called, so `FAST3D_SDL_VULKAN` is never added to the available backends and the renderer cannot be selected at all.
- `GetPixelDepth`'s `!mFrameActive` early-out returns an empty map where the early-out below it fills zeros per coordinate. `Environment_GraphCallback` queries outside the frame, so this fires almost immediately.
- The vertex ring drops the overflowing draw and every draw after it in that frame; Hyrule Field renders with no Link and almost no HUD. Grown the same way `GrowStagingBuffer` does, via the slot's garbage list.

That combination is probably why it looked fine in review — the backend could not be reached from the menu. With the four commits applied it renders correctly and has been played for a while without artefacts.

bassdr:
Not caused by this PR, but it shows up through it: the window title reads `soh (Vulkan)` rather than `Ship of Harkinian (Vulkan)`.

The backend suffix is consistent — `gfx_sdl.cpp:403` and `gfx_dxgi.cpp:564` both format `"%s (%s)"`. The first `%s` is the problem. `Fast3dWindow.cpp:130` takes `GetContext()->GetShortName()`, which is the config-directory identifier, where `Context`'s constructor documents `name` as the human-readable one and `shortName` as the short identifier. Came in with #1185.

`GetContext()->GetName()` gives `Ship of Harkinian (Vulkan)`; verified on a local build, one-word change. Flagging it here only so the title is not read as something this backend introduced.

PR #1267, opened 2026-08-30
Fixes name for ShutdownHandler SDL quit event.

Kenix3:
The SDL_Quit event was renamed to SDL_Event_Quit in SDL3. A PR was merged which uses the old name, so it now needs to be updated to use the new name.


briaguya0 (approved):
documented as the correct rename in https://wiki.libsdl.org/SDL3/README-migration#sdl_eventsh

PR #1269, opened 2026-08-31
Make ConsoleVariable thread-safe (port-maintenance)

bassdr:
Brings main's #1253 to port-maintenance. Same 24 lines; port-maintenance has no Component system, so the config lookups stay as they are.

Fixes the audio-thread crash. #1227 becomes an optimisation on top.


bassdr:
Changes were reverted on `main` with #1270. Closing this as #1227 is strictly better (faster and fixes the crash) unless we want a simpler fix.

PR #1288, opened 2026-09-11
Add an OpenGL ES 2.0 / GLSL ES 1.00 backend target

KiritoDv:
Adds a `USE_OPENGLES2` build option that targets OpenGL ES 2.0 and GLSL ES 1.00, alongside the existing ES 3.0 path. It is off by default and implies `USE_OPENGLES`. This is independent of the renderer PR stack and applies straight to `main`.

## Why this is small

Most of the hard part is already upstream. `textureSize()` was replaced with `texture_width`/`texture_height` uniforms in an earlier change, and `main`'s OpenGL shader directory has no `uint`/bitwise hashing and no `textureLod`. What was left was the dialect split and the ES 2.0 API gaps.

## Shader dialect

GLSL ES 1.00 has no `in`/`out` in the fragment stage, so the output declaration is now driven by a new `frag_out` context flag rather than `core_opengl || opengles`, and `gl_FragColor` is the sink.

`gl_FragDepth` does not exist either. A new `frag_depth` context entry resolves to `gl_FragDepthEXT` with `GL_EXT_frag_depth` enabled, and to `gl_FragDepth` everywhere else.

Two constructs in the shared template were already invalid in ES 1.00:

- `mix(higher, lower, cutoff)` with a `bvec3` selector. That overload arrived in GLSL 1.30 / ES 3.00. Now written `vec3(cutoff)`, which is valid in every dialect and numerically identical.
- `gl_Position.z *= 0.3f`. ES 1.00 has no float suffix. Now `0.3`.

Mask sizes previously came from `textureSize()`. They now follow the same pattern as the texture sizes, as `mask_width`/`mask_height` uniforms.

## API gaps

ES 2.0 has no sized colour formats, no `GL_DEPTH_STENCIL_ATTACHMENT`, no read/draw framebuffer split, no multisampled renderbuffers, no vertex array objects and no `glBlitFramebuffer`.

- Framebuffer copies go through `BlitFramebufferAsQuad`, which draws a textured quad and then invalidates the cached GL state it disturbed.
- MSAA is pinned to 1, since `GL_MAX_SAMPLES` does not exist.
- The depth probe's staging blit is skipped. It only ever fed `glReadPixels(GL_DEPTH_STENCIL, ...)`, which is already compiled out on all GLES, so ES 2.0 returns the same zeros ES 3 does today.
- ImGui has no Android auto-detect, so `IMGUI_IMPL_OPENGL_ES2` is set from CMake. Without it ImGui's desktop path compiles in and calls entry points GLES does not export.
- The SDL context is requested as ES 2.0 explicitly, so an ES 3.x context cannot mask ES 2.0 violations at runtime.

## Validation

No behaviour change on any existing target: every edit outside the shader template is behind `USE_OPENGLES2`, and the two shader fixes are valid in all four dialects.

- **Desktop macOS build** of `libultraship` succeeds.
- **Compile check against real Android NDK GLES2 headers** (`--target=aarch64-linux-android24`, NDK sysroot, `USE_OPENGLES2=1`) for `gfx_opengl.cpp`, `gfx_sdl.cpp` and `Fast3dGui.cpp`: 0 errors. The flags come from the project's own compile database, so every define and include path is the real one. The check was confirmed strict by injecting `glBlitFramebuffer`, `glReadBuffer`, `glGenVertexArrays`, `GL_MAX_SAMPLES`, `glRenderbufferStorageMultisample` and `GL_DEPTH_STENCIL_ATTACHMENT`, all six of which it rejected.
- **Generated GLSL validated with `glslangValidator`.** A harness runs the real prism contexts from `gfx_opengl.cpp` over the real template, with features supplied by the engine's own `gfx_cc_get_features()` driven by shader id pairs. 250 shader ids x 2 stages x 4 dialects, 2000 shaders: zero failures in every dialect, including ES 1.00. That sweep is what caught the `mix()` and float-suffix bugs.
- A repo-wide sweep found no other ES3-only GL usage outside `gfx_opengl.cpp`.

Not yet run on real GLES 2.0 hardware. Non-power-of-two wrapping and mipmaps depend on `OES_texture_npot`, and packed depth/stencil on `OES_packed_depth_stencil`.


bassdr:
See #1294 about the failed check

PR #1289, opened 2026-09-13
fast3d: bound the segment table on both read and write

bassdr:
G_MW_SEGMENT indexes `mSegmentPointers[16]` with `offset / 4`, and `offset` arrives as `C0(0, 16)` — a full 16-bit field. `segNumber` reaches 16383: an unbounded 64-bit write up to ~128 KB past the array. The `G_MW_SEGMENT_INTERP` case directly below it already wraps with `offset % 16`.

`SegAddr` has the matching read: `segNum = (uint32_t)(w1 >> 24)`, unmasked, on a `uintptr_t`. A display list word of `0x24C00157EB` indexes entry 9408.

The RSP segment number is a 4-bit field, so wrapping is what the hardware does.

Found chasing a `gfx_dl_handler_common` SIGSEGV that ended in `double free or corruption (out)` — consistent with the stray write corrupting the heap and the segment table, then the next display list resolving a segment against garbage. Both copies (F3DEX2 and F3D) are fixed.

### DIFF
--- a/src/fast/interpreter.cpp
+++ b/src/fast/interpreter.cpp
@@ -2427,7 +2427,10 @@ void Interpreter::GfxSpMovewordF3dex2(uint8_t index, uint16_t offset, uintptr_t 
             mRsp->fog_offset = (int16_t)data;
             break;
         case G_MW_SEGMENT: {
-            int segNumber = offset / 4;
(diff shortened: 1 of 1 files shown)


bassdr:
See #1294 about the failed check

PR #1297, opened 2026-09-17
Title the window with the application name, not its short name

bassdr:
The window reads `soh (Vulkan)` instead of `Ship of Harkinian (Vulkan)`.

`Context`'s constructor documents its two identifiers: `name` is the human-readable application name, `shortName` the short identifier that names the config directory. #1185 wired the title to `GetShortName()`.

The backend suffix is not involved — `gfx_sdl.cpp` and `gfx_dxgi.cpp` both format `"%s (%s)"` — so every backend shows the short name today.

Verified on a local SoH build: the title becomes `Ship of Harkinian (Vulkan)`.

https://claude.ai/code/session_01EXKsVhU17hmokCt5VtPk7U


bassdr:
`tidy-format` fails here on `src/ship/audio/CoreAudioAudioPlayer.cpp`, which this PR does not touch — it is not clang-format-14 clean on `main` itself, so every PR against `main` currently inherits it. #1294 is the one-file fix.

